PR #45, opened 2015-02-09
Add failing test for 'toString' on media queries

markdalgleish:
I seem to have found a bug when calling `toString` on style sheets with media queries.

I'm not sure yet how to fix this, but I figured I'd open a discussion by providing a failing test.

The current failing test case shows how this:

```js
{a: {color: 'red'}, '@media (min-width: 1024px)': {a: {color: 'blue'}}}
```

is erroneously compiled into this:

```css
.jss-0 {
  color: red;
}
.jss-1 {
  a: [object Object];
}
```

### DIFF
--- a/test/StyleSheet.js
+++ b/test/StyleSheet.js
@@ -107,6 +107,15 @@ test('toString named', function () {
     ss.detach()
 })
 
+test('toString named with media query', function () {
+    jss.Rule.uid = 0
+    var ss = new jss.StyleSheet({a: {color: 'red'}, '@media (min-width: 1024px)': {a: {color: 'blue'}}})
+    ss.attach()
+    equal(ss.toString(), '.jss-0 {\n  color: red;\n}\n@media (min-width: 1024px) {\n  .jss-0 {\n    color: blue;\n  }\n}')
+    equal(ss.element.innerHTML, '\n.jss-0 {\n  color: red;\n}\n@media (min-width: 1024px) {\n  .jss-0 {\n    color: blue;\n  }\n}\n')
+    ss.detach()
+})
+
 test('link', function () {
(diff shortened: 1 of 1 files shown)


kof:
nice catch @markdalgleish will fix it asap

kof:
huh you got me debugging for quite some time util I figured out you have used @media query differently, it has been designed to be like

{
   name: {'@media (min-width: 1024px)': {a: {color: 'blue'}}}
}

where name is there to have this rule accessible via it.

kof:
wait, there is another issue.

kof:
I need to rethink at-rule syntax in jss ... I have tested this way

```
    var rule = new jss.Rule('@media print', {
        button: {
            display: 'none'
        }
    })

````
However in named style sheet, selector is generated. At-identifiers shouldn't be treated as selectors.

**Comment by kof** [comment]
The problem is if we go this way

```
{a: {color: 'red'}, '@media (min-width: 1024px)': {a: {color: 'blue'}}}
```
It is possible to have rules with same name ... So rules within @media become unaccessible if you want to do sheet.getRule('a')

**Comment by kof** [comment]
I think ideally every rule should be still accessible via getRule and every rule should be able to be part of some @rule

**Comment by kof** [comment]
This issue is exists only in named style sheet.

**Comment by kof** [comment]
Semantically @media is a "conditional rule group" https://developer.mozilla.org/de/docs/Web/CSS/At-rule

Potentially we could have following syntax for conditionals: 

```
{
    myName: {
        condition: '@media (min-width: 1024px)',
        color: 'blue'
    }
}
```

kof:
I have added your test but modified to unnamed style sheet 8ac23ecfd3dcf5ef1c539e92661299021ae81b07 just to make sure this will work.

As for named style sheet I need to sleep over it. Its actually not a bug, its a missing feature.

PR #50, opened 2015-02-26
Replace NO-BREAK SPACE character with normal space in source code

nkbt:
For some reason there are heaps of NO-BREAK SPACE characters in the source code

http://www.ltg.ed.ac.uk/~richard/utf-8.cgi?input=%C2%A0&mode=char

### DIFF
--- a/bin/cssToJss.js
+++ b/bin/cssToJss.js
@@ -14,7 +14,7 @@ var css = require('css')
  * @return {String}
  * @api public
  */
-module.exports = function (code, options) {
+module.exports = function (code, options) {
     var ast = css.parse(code)
 
     if (ast.stylesheet && ast.stylesheet.rules) {
--- a/dist/jss.js
+++ b/dist/jss.js
@@ -88,7 +88,7 @@ Rule.uid = 0
  * @return {Rule|String|Number}
  * @api public
  */
-Rule.prototype.prop = function (name, value) {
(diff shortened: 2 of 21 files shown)


kof:
thanks, merged ... I would love to know when and why sublime does this.

kof:
published

nkbt:
That could be OSX rather then Sublime. And Sublime is not smart enough to un-fix this. OSX replaces a lot of things as-you-type, so this could be the case. IDEA/Webstorm do not have this issue as far as i know.

nkbt:
http://superuser.com/questions/78245/how-to-disable-the-option-space-key-combination-for-non-breaking-spaces

> As mentioned in the comments, the solution provided by Martin does not work in some applications, most importantly (for me) in my editor, Sublime Text 2.

> So for all of you who are troubled by none breaking spaces in your code, you can add the following custom keybinding into your user keymap ("Sublime Text 2 > Preferences > Key Bindings - User"):
```
{ "keys": ["alt+space"], "command": "insert_snippet", "args": {"contents": " " } }
```
This will insert a regular space instead of a non-breaking one.

kof:
wow thanks a lot!

nkbt:
@kof check this: http://eslint.org/docs/rules/no-irregular-whitespace

kof:
:+1:

PR #52, opened 2015-02-27
allow value of 0 for props that can have value 0. Closes #51

### LINKED ISSUE #51: Falsy value does not permit valid 0
Prop should be able to have a value that is a string or number.  Currently, if the value is falsy then the prop does not receive a value.

The evaluation should permit the value 0.

urbantumbleweed:


kof:
Please change it to value != null, this way we support any non falsy value

kof:
no but then its up to user to pass something usable we just ensure if its non-falsy its a  setter.

kof:
merged

PR #63, opened 2015-05-29
Add jss to global

trusktr:


kof:
oh what you did is actually already done by browserified version

kof:
If you load jss script from dist without require on the page, it will expose the api on window

trusktr:
Oh, cool!! I missed the dist. :+1:

PR #65, opened 2015-05-31
Add package entry point. Fixes #62

### LINKED ISSUE #62: global version of jss.
It'd be super nice if the scripts section of package.json had a script using browserify or webpack to compile a standalone bundle to run in the browser that adds `jss` to the global scope. Some people could benefit from this right away in [certain scenarios](https://forums.meteor.com/t/stylus-css-vs-less-vs-sass/3752/29?u=joe) depending on their stack.

joshgillies:


kof:
package.json and bower.json are generated through xpkg module, you need to change only x-package file and run `xpkg .`

Also I think the right main for npm would be the index.js, not builded one, but the dist one for bower

joshgillies:
Oh! Sorry I didn't even consider the xpkg thing. Nor have I encountered it before. Just taking a look now...

So really, we need to expose `"main": "index.js"` for Node,  which is the default for Node anyway so not necessary, and `"browser": "dist/jss.js"` for bower/the browser?

joshgillies:
Can xpkg manage generating a package.json with `"main": "index.js"` and a bower.json with `"main": "dist/jss.js"`?

joshgillies:
Ah looks like you can via and "overlay". Want me to update https://github.com/jsstyles/jss/blob/master/x-package.json5 with the above?

kof:
exactly, over overlay.

joshgillies:
Ok, let me know if these changes are acceptable.

package.json
```diff
@@ -6,6 +6,7 @@
     "name": "Oleg Slobodskoi",
     "email": "oleg008@gmail.com"
   },
+  "main": "dist/jss.js",
```
kof: you didn't rung xpkg .

joshgillies: Argh! The merge overwrote it. Just re-ran `xpkg`, what do you think?

kof:
merged, thanks!

Issue #69, opened 2015-07-31
Can't run jss in a windowless environment


qimingweng:
When I have a file that is just `require('jss')`, and I run it in node like `node file.js`, it always returns this error: 

```
if (global[namespace] == null) global[namespace] = 0
          ^
TypeError: Cannot read property '__JSS_UID_PREFIX__' of undefined
```

The offending line is `jss/lib/uid.js:5:11`

This can be traced to the way the variable `global` is defined. This file will also not compile in node (I'm using 0.12.7, the newest one as of writing).

```
var global = typeof window === 'undefined' ? global : window;
global[0] = 0;
```

The error:
```
global[0] = 0;
          ^
TypeError: Cannot set property '0' of undefined
```

qimingweng:
I created a pull request #70

kof:
merged, thanks

PR #70, opened 2015-07-31
Rename global reference to globalReference

qimingweng:
Allows jss to be run in node


qimingweng:
This references the issue I described here #69

qimingweng:
I'm not entirely sure about the build process, but I did run `make` before submitting this pull request. Is that enough?

kof:
thanks!

PR #72, opened 2015-08-17
Return the instance from jss.use()

mpal9000:


kof:
do you really want it to look like this:

```
  jss.use(something)
     .use(something)  
     .use(something)
     .use(something)
```

vs.

```
  jss.use(something)
  jss.use(something)
  jss.use(something)
  jss.use(something)
```

I intentionally didn't return

kof:
Also there is not much value in returning  something one have a reference to anyways.

mpal9000:
I think that `use()` is a very common plugin related method and being chainable is something also very common. Probably it is the expected behaviour and also it's better to leave that style choice to the user. But in any way this is a very useful library and thank you!

kof:
allright, lets allow them this fun)

mpal9000:
These are also nice:
```javascript
jss.use(something).use(something).use(something).use(something)
```
```javascript
jss
.use(something)
.use(something)  
.use(something)
.use(something)
```
:smile:

kof:
yeah, love them...

PR #73, opened 2015-08-17
Return added rules from styleSheet.addRules()

mpal9000:
Without this, there is no way to get the `classNames` that are related ONLY to the rules that have been currently added (with `addRules()`). This is different from `styleSheet.rules` , which stores ALL current rules.

A use case is when we need the `classNames` of some added rules, while rendering on the server. Maybe there is a need to add rules of all components in a tree, to the same global StyleSheet instance. That instance will be created on each request. We need `classNames` of each component, before calling `styleSheet.toString()` at the final step, when all rules have been added and the complete tree has been rendered. In that case we have to get the `classNames` from the returned values of `addRule()/addRules()`.


kof:
I think its  a good idea

PR #77, opened 2015-08-27
Remove invalid character: fix #76

### LINKED ISSUE #76: MAJOR: weird character / encoding issue in lib/uid.js
Error
=====
There is an invalid `0xc2` character in `lib/uid.js` as you can see in the hexdump I just retrieved from master:
```
000001f0  72 74 73 2e 72 65 73 65  74 20 3d 20 66 75 6e 63  |rts.reset = func|
00000200  74 69 6f 6e 28 29 c2 a0  7b 0a 20 20 20 20 63 6f  |tion()..{.    co|
00000210  75 6e 74 65 72 20 3d 20  30 0a 7d                 |unter = 0.}|
```

After running it through webpack I get this:
``` javacript
	exports.reset = function()Â {
	    counter = 0
	}
```

Background
==========
I'm trying out jss with webpack and the resulting javascript is invalid. At first I was blaming my webpack config so I removed everything to find the cause. Eventually I ended up with a close to empty webpack config. Before blaming webpack I looked at the orignal `uid.js` and found the error.

I also scanned the rest of the js files and it is only in this file and the non minified dist file.

Workaround
===========
Minifying with UglifyJS removes the invalid character.

0xR:
Remove invalid 0x2C character from lib/uid and create patch release

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,7 @@
 {
   "name": "jsstyles",
   "description": "Composable and reusable style sheets.",
-  "version": "2.1.5",
+  "version": "2.1.6",
   "author": {
     "name": "Oleg Slobodskoi",
     "email": "oleg008@gmail.com"
--- a/lib/uid.js
+++ b/lib/uid.js
@@ -24,6 +24,6 @@ exports.get = function() {
  *
  * @api private
  */
-exports.reset = function() {
+exports.reset = function() {
     counter = 0
 }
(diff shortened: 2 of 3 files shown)


0xR:
Fixes #76

kof:
thanks

ghengeveld:
@kof Please publish to npm, so we can stop using our fork.

kof:
done

PR #104, opened 2015-10-18
Rule logic refactoring

### LINKED ISSUE #81: Named rules not generating when using @media only
When I create a stylesheet and do something like this:

```js
sheet.addRules({
  something: { ... },
  '@media (max-width: 540px)':
    desktopOnly: {
      display: 'none'
    }
  }
})
```

I get `undefined` for `sheet.classes.desktopOnly`. In order to work I need to do this instead:

```js
sheet.addRules({
  something: { ... },
  desktopOnly: {},
  '@media (max-width: 540px)':
    desktopOnly: {
      display: 'none'
    }
  }
})
```

I only bring it up because in CSS I don't need to create an empty selector to use media queries, so if the latter is what needs to be used it should be documented.

kof:
- also fixes #81


kof:
@mpal9000 wanna help with code review?

mpal9000:
yes, I will check tomorrow

kof:
I will merge it, if you find something, let me know here ....

mpal9000:
Hi, some thoughts:

- The dom separation needs one more step in my opinion. To let the user import the parsing functionality without the dom logic. One way could be to move the dom related logic from Stylesheet.js (eg. `attach()`, `detach()`, `deploy()`, `link()`) to dom.js, as separate functions. Then the `Stylesheet` class could be refactored so it would be portable between browser and node, ie. use `Stylesheet` only as a container and "manager" of multiple rules, without any dom related logic. Another way could be to have two different Stylesheet.js files (browser/node) and use the `browser` field in package.json

- I think it's safer to not mutate object arguments like `options` here https://github.com/jsstyles/jss/blob/d9970dd1dc34e1d71ca17f9a37660cd12b6117f7/src/StyleSheet.js#L148,
https://github.com/jsstyles/jss/blob/f8685da8ca85148cb73e2e014813a15a96e7fa76/src/Jss.js#L33,
https://github.com/jsstyles/jss/blob/f8685da8ca85148cb73e2e014813a15a96e7fa76/src/Jss.js#L44

- You can use `Object.create(null)` is some places like here https://github.com/jsstyles/jss/blob/7fe2623f86b76c04358ca90be862cf587ce0ee14/src/ConditionalRule.js#L22
or here https://github.com/jsstyles/jss/blob/7fe2623f86b76c04358ca90be862cf587ce0ee14/src/StyleSheet.js#L29,
because you only set/get properties from those objects and will make `for...in` iterations and lookups safer and maybe a little faster

- You can replace `let` with `const` in some places

I will run the new changes in the weekend

kof:
- dom separation: yep dom renderer similar to the latest react would be the next step, for now it was just separation of dom related functions.
- mutations: agree
- Object.create ... well its cloning via inheritance ... faster - yes, sideeffects? for e.g. those props are not in .hasOwnProperty any more .... hmm

mpal9000:
About `Object.create()`, what do you mean by cloning and side effects? I meant only for the creation of some specific objects like `rules`. The Object.prototype's properties won't be inside the object and that will make `for...in` to work like you really want.

mpal9000:
A native `Map()` could be an alternative, but I'm not sure if the transpiled `Map` will give you the same performance.

kof:
Oh you are actually right, for places like style and rules etc where I really expect plain object, pure object would be nice.

mpal9000:
With `null` prototype is as pure as it gets :-)

kof:
https://github.com/jsstyles/jss/commit/335ad580c77426641ce5261c18691e3627f0b993

PR #168, opened 2015-11-22
fix multiple non-regular rule types in the same stylesheet

alexkuz:
`rule.id` is `undefined` for keyframes or media rules, so it ends up rendering only one of them.

(haven't check with `@font-face` rule though, it might fail as it has `'regular'` type)

### DIFF
--- a/src/StyleSheet.js
+++ b/src/StyleSheet.js
@@ -126,12 +126,15 @@ export default class StyleSheet {
       const rule = rules[name]
       // We have the same rule referenced twice if using named rules.
       // By name and by selector.
-      if (stringified[rule.id]) {
+      if (rule.type === 'regular' && stringified[rule.id]) {
         continue
       }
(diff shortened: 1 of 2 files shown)


kof:
OK on it now.

kof:
I see 2 possibilities:
1. What you did
2. add id to all rules, also to conditional ...

kof:
I think its a more streamlined logic to make all rules have an id.

kof:
Don't worry, I will fix it. Taking your test as a  base and adding ids to all rule types.

kof:
fixed.

kof:
Will be released soon with a bunch of other fixes/enhancements. Please let me know if you want to participate in pre-release testing or want to help in any other way. I could use some help)

alexkuz:
Cool, thanks!

PR #174, opened 2016-01-15
Make the `use` method variadic.

wldcordeiro:
This allows you to use the `use` method like so:
```js
jss.use(extend(), nested(), defaultUnit());
```

### DIFF
--- a/src/Jss.js
+++ b/src/Jss.js
@@ -60,11 +60,11 @@ export default class Jss {
   /**
    * Register plugin. Passed function will be invoked with a rule instance.
    *
-   * @param {Function} fn
+   * @param {Function} plugins
    * @api public
    */
-  use(fn) {
-    this.plugins.use(fn)
+  use(...plugins) {
+    plugins.forEach(plugin => this.plugins.use(plugin))
     return this
   }
 }
--- a/test/Rule.js
+++ b/test/Rule.js
@@ -195,6 +195,16 @@ test('add plugin', () => {
   jss.plugins.registry = []
 })
 
+test('add multiple plugins', () => {
+  const plugin = () => {}
+  const plugin2 = () => {}
+  jss.use(plugin, plugin2)
+  equal(jss.plugins.registry.length, 2)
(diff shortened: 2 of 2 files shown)


kof:
can nodes' connect middleware work like this?

In general its nice to have, however jss setup is done only once and we don't really need to have 2 ways to do the same thing. Or I miss any other concerns?

wldcordeiro:
I'd compare it to how PostCSS registers its plugins, in that you add a block like
```js
    postcss() {
        return [
            autoprefixer({browsers: ['last 2 versions']}),
            postcss_import,
            lost,
            postcss_cssnext,
            postcss_mixins,
            postcss_nested,
            postcss_simple_vars,
            postcss_extend,
        ];
    },
```

to your webpack config and that would register all your plugins at once. I know that it's only a small amount more typing but to me, it seems like I shouldn't need to call a method for each plugin, when dealing with a registry. I should be able to give it a list of plugins and it handles registering each. It seems like a nicer UX to me to have it be variadic.

kof:
what about this)

```javascript
[plugin1(), plugin2()].forEach(jss.use)
```

kof:
Just took a look at postcss, they do it because postcss() returns a stream which can be piped to something else. This allows to chain pipes.

```
    return gulp.src('src/**/*.css')
        .pipe( sourcemaps.init() )
        .pipe( postcss([ require('autoprefixer'), require('precss') ]) )
        .pipe( sourcemaps.write('.') )
        .pipe( gulp.dest('build/') );
```

kof:
I think we shouldn't do it just for the sake of not having multiple ways to do the same thing without a reason. This optional possibilities always lead to confusion and increase api complexity.

wldcordeiro:
@kof I think we approach API design differently. I think making the function variadic adds flexibility to it that allows the user to more expressively write code. Yes you can map, or forEach it but that's just boilerplate that could easily be hidden away in the `use` function at no cost to the user.

kof:
ok, will merge it. I see one point in this change: the order of plugins is more explicit.

PR #179, opened 2016-01-30
Trim empty rules.

wldcordeiro:
This PR introduces empty rule trimming, this isn't too useful to raw JSS (because you won't write empty rules), but with jss-nested you could end up with empty rules with a structure like:
```js
var style = {
  awesomeThing: {
    '& nestedAwesomeThing': {
      color: red;
    }
  }
};
```

### DIFF
--- a/src/PluginsRegistry.js
+++ b/src/PluginsRegistry.js
@@ -23,8 +23,8 @@ export default class PluginsRegistry {
    * @api private
    */
   run(rule) {
-    for (let index = 0; index < this.registry.length; index++) {
-      this.registry[index](rule)
-    }
+    this.registry.forEach(plugin => {
+      plugin(rule)
+    })
   }
 }
--- a/src/PluginsRegistry.js
+++ b/src/PluginsRegistry.js
@@ -23,8 +23,8 @@ export default class PluginsRegistry {
    * @api private
    */
   run(rule) {
-    this.registry.forEach(plugin => {
-      plugin(rule)
-    })
+    for (let index = 0; index < this.registry.length; index++) {
+      this.registry[index](rule)
+    }
   }
 }
--- a/src/StyleSheet.js
+++ b/src/StyleSheet.js
@@ -1,3 +1,4 @@
+import isEmptyObject from './empty-object'
 import createRule from './createRule'
 import findRenderer from './findRenderer'
 
@@ -130,7 +131,7 @@ export default class StyleSheet {
         continue
       }
 
-      if (rule.style && Object.keys(rule.style).length === 0) {
+      if (rule.style && isEmptyObject(rule.style)) {
         continue
       }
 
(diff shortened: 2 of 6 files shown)


wldcordeiro:
I'm not sure whether this is a better solution or having each Rule type's `toString` return a null and check for the existence before joining. This way may only work with normal rules.

src/PluginsRegistry.js
```diff
@@ -23,8 +23,8 @@ export default class PluginsRegistry {
    * @api private
    */
   run(rule) {
-    for (let index = 0; index < this.registry.length; index++) {
-      this.registry[index](rule)
-    }
+    this.registry.forEach(plugin => {
```
kof: it was a micro performance optimization, because this runs on every rule.

src/StyleSheet.js
```diff
@@ -129,8 +129,14 @@ export default class StyleSheet {
       if (stringified[rule.id]) {
         continue
       }
+
+      if (rule.style && Object.keys(rule.style).length === 0) {
```
kof: As not every rule has a style property.

lets create a util isEmptyObject (separate file, same like clone)

http://jsperf.com/object-keys-length-vs-for-in-return
```
function isEmptyObject(obj) {
  for (var key in obj) return false
  return true
}
```

src/PluginsRegistry.js
kof: so please remove your change ... it generates a function on every .rule call.

src/empty-object.js
```diff
@@ -0,0 +1,5 @@
+export default function isEmptyObject(obj) {
```
kof: file name should be is-empty-object

src/empty-object.js
```diff
@@ -0,0 +1,5 @@
...
+  for (let key in obj) return false
```
kof: should be const ...

kof: didn't you get an error from linter?

src/StyleSheet.js
```diff
@@ -129,8 +130,14 @@ export default class StyleSheet {
       if (stringified[rule.id]) {
         continue
       }
+
+      if (rule.style && isEmptyObject(rule.style)) {
```
kof: empty check could be also part of isEmptyObject I think ...

src/empty-object.js
kof: would add here 
`if (!obj) return true`

src/StyleSheet.js
kof: please ignore this comment ...

src/empty-object.js
wldcordeiro: I did, but then it also errored about the const being unused, maybe update the eslint rules to allow `_` as an unused? That way the line can be `for (const _ in obj) return false`

src/empty-object.js
kof: pleae ignore that comment  ...

kof:
merged, thanks!

PR #208, opened 2016-04-14
Add support for array of styles for @font-face rule

### LINKED ISSUE #199: [jss] Can't have multiple @font-face entries in a stylesheet
You can have a stylesheet with a single `@font-face` entry in it, and it works fine, however, I have multiple fonts I want to include on a stylesheet.  From the code, this seems not possible today (let me know if I missed something).

If the style behind the `@font-face` object key were optionally an array instead of an object, then you could have multiple fonts.

    '@font-face': [
      {
        fontFamily: 'Lato-Regular',
        src: 'url("/fonts/Lato-Regular.ttf") format("truetype")',
      },
      {
        fontFamily: 'Lato-Bold',
        src: 'url("/fonts/Lato-Bold.ttf") format("truetype")',
      },
    ],

  The `toString()` function in `FontFaceRule.js` could be modified something like this:

    toString(options) {
      let str = ''
      if (Array.isArray(this.style)) {
        for (let index = 0; index < this.style.length; index++) {
          str += toCSS({...this, style: this.style[index]}, options)
        }
        console.log('@font-face CSS =  ' + str)
        return str
      }
    
      return toCSS(this, options)
    }

Thoughts?  If this sounds good, I can submit a PR.

mindjuice:
Also adds a test case to handle @font-face with multiple styles in an
array

Fixes #199

(corresponding fix to jss-camel-case coming shortly)

### DIFF
--- a/src/rules/FontFaceRule.js
+++ b/src/rules/FontFaceRule.js
@@ -21,6 +21,15 @@ export default class Rule {
    * @api public
    */
   toString(options) {
+    let str = ''
+    if (Array.isArray(this.style)) {
+      for (let index = 0; index < this.style.length; index++) {
+        str += toCSS({...this, style: this.style[index]}, options)
+        str += '\n'
+      }
+      return str
+    }
+
     return toCSS(this, options)
   }
 }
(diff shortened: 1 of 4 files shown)


kof:
looks nice to  me

PR #211, opened 2016-04-16
Add jss.version

watilde:
Fixes https://github.com/jsstyles/jss/issues/207.

Updates
+ Added json-loader to load `package.json` in test code
+ Added `__JSS_VERSION__` as the version of `jss`


webpack.config.js
```diff
@@ -6,7 +6,10 @@ var plugins = [
   new webpack.DefinePlugin({
     'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
     __DEV__: process.env.NODE_ENV === 'development',
-    __TEST__: process.env.NODE_ENV === 'test'
+    __TEST__: process.env.NODE_ENV === 'test',
+    '__JSS_VERISON__': (function () {
```
kof: I think we can simply call it __VERSION__

webpack.config.js
```diff
@@ -6,7 +6,10 @@ var plugins = [
...
+      return JSON.stringify(require('./package.json').version)
```
kof: you don't need the closure and also the JSON.stringify call

kof:
we have 2 builds, lib and dist, dist ist created by webpack, lib is just babel, so right now the babel build is broken because the variable won't be replaced. I think we need something like a plugin for babel to do that. `__VERSION__` is not the same like `process.env.NODE_ENV` because the last one is something we expect to be replaced by the root project build, not the version though, it results in "Uncaught ReferenceError: __JSS_VERISON__ is not defined"

src/index.js
```diff
@@ -6,6 +6,9 @@
  * @license MIT
  */
 
+// Optimize for webpack UglifyJsPlugin
+global.__VERSION__ = global.__VERSION__ || false
```
watilde: If we doesn't add this line, `dist` file will contain all data of the `package.json`.

watilde:
@kof Thanks for your review! I've fixed 3 points you commented.

src/Jss.js
```diff
@@ -15,6 +15,11 @@ export default class Jss {
     this.sheets = new SheetsRegistry()
     this.plugins = new PluginsRegistry()
     this.uid = uid
+    if (__VERSION__) {
```
watilde: This `if` is optimisation for webpack. To not contain all data of the `package.json` in `dist` file, we need to add this line.

kof: why not to simply find some babel plugin for string interpolation and use it for both lib and dist?

kof: for e.g. http://babeljs.io/docs/plugins/transform-inline-environment-variables/

watilde: Yeah I searched babel plugins to do it before creating patch. To keep not adding more build tools, I didn't use it because `package.json` can be ugly:
```
    "build:lib": "V=`npm run v -s` && __VERSION__=$V babel src --out-dir lib",
    "v": "node -e \"console.log(require('./package.json').version)\""
```

Also, `src` files are called from `test` codes so we need to add `require('../package.json').version`, but if we don't add this line, `dist/jss.js` and `dist/jss.min.js` actually contain `package.json` data like this: https://gist.github.com/watilde/7aacb5e924fc7317308dbab367a37d89#file-jss-js-L1602-L1668

kof: I am sure we can do the same but in a non-ugly way, for e.g. by using .babelrc if its possible to do from there, otherwise by creating a script in bin/babel and doing the same by requiring babels api.

kof: Tests are using webpack, so lets use both, webpacks replacement and plugin for babel at the same time ...

kof: and yes, I know, such a simple task but so hard to do right ...

kof: Another option would be to just use a custom script or a lib which just runes over lib and dist files and replaces something. And call it in the "scripts" commands

kof: for e.g. something like this https://github.com/gkz/grasp and then just use it exclusively, no webpack, no babel .... I didn't try it, but worth considering, it might make it much simpler.

watilde: Agreed, this can be more simple implementation. Let me try again!
Thanks for your comments.

watilde:
Close by https://github.com/jsstyles/jss/issues/213#issuecomment-212549709. Also I will recreate a patch to make it simpler!

PR #215, opened 2016-04-20
[wip] functional style

kof:


kof:
much slower ...

kof:
2x slower

mr-mig:
What is that and where can I have a look at it?

src/createPluginsRegistry.js
```diff
@@ -0,0 +1,28 @@
...
+    registry.push(fn)
+  }
+
+  /**
+   * Execute all registered plugins.
+   *
+   * @param {Rule} rule
+   */
+  function run(rule) {
+    for (let index = 0; index < registry.length; index++) {
```
mr-mig: WDYT?
`registry.forEach( fn => fn(rule))`

src/rules/createFontFaceRule.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+export default function createFontFaceRule(selector, style, options) {
+  const rule = createBaseRule('font-face', {selector, style, options})
+
+  /**
+   * Generates a CSS string.
+   *
+   * @see toCSS
+   */
+  rule.toString = (options) => toCSS(rule, options)
```
mr-mig: I think assigning functions like that is a perf killer in general, but I am not sure.

src/rules/createKeyframeRule.js
```diff
@@ -0,0 +1,44 @@
+import createBaseRule from './createBaseRule'
+
+/**
+ * Creates formatted frames where every frame value is a rule instance.
+ */
+function formatFrames(frames, rule) {
+  const newFrames = Object.create(null)
+  for (const name in frames) {
```
mr-mig: This may be the case: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#52-the-object-being-iterated-is-not-a-simple-enumerable

src/rules/createRegularRule.js
```diff
@@ -0,0 +1,145 @@
...
+  Object.defineProperty(rule, 'selector', {
+    get: () => {
+      if (rule.renderable) {
+        return rule.options.Renderer.getSelector(rule.renderable)
+      }
+
+      return selector
+    },
+
+    set: (newSelector = '') => {
```
mr-mig: I believe the whole function will be deoptimized: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#2-unsupported-syntax

mr-mig:
Some useful resources for functional-style related optimizations: 

* https://github.com/petkaantonov/bluebird/wiki/Optimization-killers
* http://www.eecs.berkeley.edu/Pubs/TechRpts/2014/EECS-2014-144.pdf
* http://www.html5rocks.com/en/tutorials/speed/v8/

kof:
benchmarks are in master, I ported only one part of this code rules/createRule to the new code base. It was at least twice slower.

Slowness was actually expected, the question was how much slower. And it turns too much for this library. It has some critical performance parts.

mr-mig:
https://github.com/Berkeley-Correctness-Group/JITProf

mr-mig:
I am trying to speculate that those perf problems can be solved :)

kof:
Well those perf problems exist in master as well, try to run benchmark and see if you can change something.

PR #216, opened 2016-04-20
Add jss.version

watilde:
from https://github.com/jsstyles/jss/pull/211. This part we discussed before become simple and readable code. The benefit is the `jss.version` in `src/jss.js` can work fine without babel's transpile process.

```js
if (!global.__VERSION__) {
  global.__VERSION__ = require('../package.json').version
}
```


kof:
Don't understand how this should work in the lib version generated by babel.

Also why do you need to use global ... you can just do typeof

kof:
Oh I do understand now how this should work within babel, but this way you load the entire config and also use require. We use imports everywhere , require looks like  a hack.

watilde:
The reason why I used `global.__VERSION__` is, webpack will build `dist/jss.js` and `dist/jss.min.js` as a file contains entire `package.json` config unfortunately if we use `typeof __VERSION__`.

It's really correct. Currently [test/Jss.js](https://github.com/jsstyles/jss/blob/master/test/Jss.js#L1) is depending on src files and it the complex point. If the test can use `lib/index.js` for test, we can make it simpler. Just it can be passed the version value from Babel plugin.
> We use imports everywhere , require looks like a hack.

kof:
What if we just use something like this: https://github.com/gkz/grasp
No webpack variable, no babel plugins.
I didn't try it but it seems like we could do it very easily, take a look.

watilde:
I updated this patch to make it simple again. Finally it worked fine with only webpack and babel. The only thing I should do is just use same name of the variable for version...

kof:
Nice, this looks much better!

kof:
merged!!!!

kof:
thanks a lot!

watilde:
Thank you so much for following up hardly! Cheers.

PR #251, opened 2016-06-27
delete deployed property and move check on attached

b2whats:


kof:
why?

b2whats:
we have two identical property - attached and deployed. when 'deployed' === true, 'attached' as well === true and deployed uses in one case https://github.com/jsstyles/jss/blob/master/src/StyleSheet.js#L87

kof:
no, this is wrong and also the last pr was also wrong, I misread it and need to revert.

.deployed means CSS has been injected into style node.
.attached means style node is inserted into the render tree.

StyleSheet should be deployed only once. After that when rules are added, a different api is used internally compared to adding rules before first deployment.

This is a performance optimization.

b2whats:
if i add rule in moment when sheet detached i get error

```js
const sheet = jss.createStyleSheet({}, options)
sheet.addRule('test', {color: red})
sheet.attach()  // componentWillMount
sheet.detach() // conmonentWillUnmount

sheet.addRule('test2', {color: blue})
```
but there will be no error if deployed will be equal true when sheet.detach()

kof:
That's what you had to start with in the first place,lets create  an issue and  a failing test first of all. Closing this for now.

Issue #180, opened 2016-02-01
Media queries and other conditional selectors before the rules they overwrite


wldcordeiro:
Currently if a media query is created before a rule it matches it will generate a different uid class than the rule it should match.

kof:
th

PR #257, opened 2016-07-01
Media queries and other conditional selectors before the rules they overwrite

wizardzloy:
Implements #180

I wanted to add a test for this, but looks like with the current test setup it is hard to verify the scenario, because `jss.generateClassName()` always generates class names with the same suffix: https://github.com/jsstyles/jss/blob/master/tests/utils.js#L44 . Please let me know if you'd like to still have a test for it and advise how to overcome the aforementioned issue.

### DIFF
--- a/src/StyleSheet.js
+++ b/src/StyleSheet.js
@@ -171,7 +171,8 @@ export default class StyleSheet {
       ...options,
       sheet: this,
       jss: this.options.jss,
-      Renderer: this.options.Renderer
+      Renderer: this.options.Renderer,
+      className: this.classes[name]
     }
     // Scope options overwrite instance options.
     if (options.named == null) options.named = this.options.named
--- a/tests/integration/sheet.js
+++ b/tests/integration/sheet.js
@@ -1,5 +1,5 @@
 import expect from 'expect.js'
-import jss, {Rule} from 'jss'
+import jss, { create as createJss, Rule} from 'jss'
 import {reset} from '../utils'
 
 afterEach(reset)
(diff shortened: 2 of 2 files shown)


kof:
Is something wrong with the PR? I don't see how the implementation should work ...

wizardzloy:
What it does is just checks if we already have generated class name for a particular "selector", if so uses that in the consequent places:

```js
{
  '@media print': {
    app: { ... }  // generated className is .app-jss-001
  },

  app: { ... } // use previously generated className here (.app-jss-001)
}
```

kof:
Yeah, but look at the diff here on github ... it's almost empty

wizardzloy:
It is actually a one-line-fix)

kof:
haha nice, didn't realize that's all we need, we need to add tests though

kof:
you can make sheet.toString() and see everything is generated properly

kof:
Here is a good place https://github.com/jsstyles/jss/blob/master/tests/integration/sheet.js#L108 for this test

wizardzloy:
There is an issue with testing that as I mentioned in the PR description, looks like in every test we mock `jss.generateClassName()` method that is responsible for generating the unique class names. In mocked version, it always appends the same suffix to the selector name (e.g. `"a" -> ".a-id"`). Having this, it is impossible to test if the generated class name for root `app` selector is the same that the one for `app` within @media block because they will always be same ('app-id'). 

I should probably just override `jss.generateClassName()` just for a single test to implement "random" version

kof:
You can create an own jss instance, generateClassName won't be mocked then.

kof:
import jss, {Rule,} from 'jss' -> import jss, {create, Rule} from 'jss'

wizardzloy:
ok, I'm gonna add the test

wizardzloy:
@kof, added, plz check

kof:
merged thanks, also use linter next time) npm run lint

PR #288, opened 2016-08-18
Add typescript definitions

cvle:
I'm using `jss` with `typescript`, it would be great to have the `typescript` definitions right in the repository.

I made the definitions from the api docs and briefly compared with the actual implementation for mistakes. Can someone double check this? I am especially unsure about the type of `element` in `StyleSheetOptions`.

I placed this file directly in the root folder where it will be automatically picked up by typescript. Alternatively we can put it in `src` and [add the path](http://www.typescriptlang.org/docs/handbook/typings-for-npm-packages.html) to the `package.json` file. I wasn't sure if I have to tweak the build scripts for this.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,50 @@
+// Type definitions for jss.
+export type RuleType = { [rule: string]: any } | Object;
+export type RulesType = { [name: string]: RuleType } | Object;
+export type Plugin = (rules: RulesType) => void;
+
+export interface StyleSheetOptions {
+  media?: string;
+  meta?: string;
+  named?: boolean;
+  link?: boolean;
+  element?: HTMLStyleElement;
+}
+
+export interface SetupOptions {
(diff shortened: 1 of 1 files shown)


kof:
Hmm, the problem with this is that it needs to be maintained. I am not using typescript, so it is likely to get broken. Any way we can put it into a separate repository?

kof:
And make it somebody elses task.

kof:
Found a perfect place for this https://github.com/DefinitelyTyped/DefinitelyTyped

cvle:
Ok having thought about this, I'll put them in https://github.com/DefinitelyTyped/DefinitelyTyped for now.

zhaojinxiang:
@cvle [The @types/jss package has nothing to do with jss](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/19974)

PR #289, opened 2016-08-18
Fix typo in dom renderer

nathanmarks:
`jss` would definitely benefit from a better test suite, would help avoid little things like this typo 👍 

If we end up using `jss` I can certainly help with that!


kof:
Huh thanks, yeah test coverage can never be good enough ...

Issue #290, opened 2016-08-19
Add sheet rendering priority feature


nathanmarks:
@kof quoting your reply from the stylishly issue:

> @cvle lets move priority discussion to a separate issue in jss project, because this thread should be focused. I think I could add something to jss to allow manual priority definition.

Do you have an implementation in mind already?

I know previously you weren't 100% certain about adding this feature, but given what you said in the `stylishly` thread I wanted to start up a discussion about it here as it is a critical feature for us. Otherwise I will have to create + manage DOM nodes myself to pass to `jss` to use.

cvle:
Same here, I could kick out all the custom renderer code from my [theming prototype](https://github.com/wikiwi/react-jss-theme), when we have this in core.

All I need is to associate a `priority` number to the stylesheets, e.g. as an additional option. The renderer should mount stylesheets with lower priorities before stylesheets with higher priorities. If a new stylesheet has the same priority as an already mounted one, the renderer should append to the last stylesheet with the same priority. The latter is not a hard requirement but we should define the behavior when stylesheets have the same priority.

I think this is related to https://github.com/cssinjs/jss/issues/287.

nathanmarks:
@kof 

Since the `DomRenderer` is designed to have a new instance per `StyleSheet`, I see 2 options here for getting the insertion point in `attach()` for new sheets:

1) Use the sheets in `options.jss.sheets`
2) Query the DOM in every `attach()` to get the style elements (assuming the priority would also be put in a `data-xxx` attribute)

nathanmarks:
@cvle I made a rough priority feature: https://github.com/nathanmarks/jss/blob/priority/tests/integration/priority.js

kof:
@nathanmarks what about this:

1. by default, all sheets have same priority, it is an option and can be passed
1. first style tag is rendered before all other style/link tags in the head
1. all next style tags are rendered after already rendered style tags in the order they are requested
1. if they have priority defined, renderer need to consider priority and move sheets in the dom as they come

Any issues or missing features here?

nathanmarks:
@kof That should do the trick 👍 

Quick question: what will be the allowed input for priority?

kof:
>  Quick question: what will be the allowed input for priority?

You mean which option name would we want to use? e.g. "index", or "priority" or "order"

nathanmarks:
@kof As in, any number above 0?

kof:
I would say it can be even any number, similar to z-index ...

[...]

PR #298, opened 2016-08-27
Add DOM priority feature via index option

nathanmarks:
ref #290

@kof Once you have 👍 the implementation I will document the feature.

Please review and let me know anything you'd like to see changed.

When using the priority/index feature there's some extra array iteration when adding and attaching sheets. One could use a binary search but I don't think it would make a real difference with the length of the registry array in typical usage.

This should give you a good handle on the usage -> output: https://github.com/nathanmarks/jss/blob/priority/tests/integration/priority.js

### DIFF
--- a/src/SheetsRegistry.js
+++ b/src/SheetsRegistry.js
@@ -17,7 +17,7 @@ export default class SheetsRegistry {
   add(sheet) {
     const {registry} = this
 
-    if (!registry.length || typeof sheet.options.index !== 'number') {
+    if (!registry.length) {
       registry.push(sheet)
       return
     }
@@ -30,6 +30,7 @@ export default class SheetsRegistry {
       }
     }
 
+    // Higher priority than any sheets in registry, so push on the end
     registry.push(sheet)
   }
 
--- a/src/SheetsRegistry.js
+++ b/src/SheetsRegistry.js
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
    * @api public
    */
   add(sheet) {
-    this.registry.push(sheet)
+    const {registry} = this
+
+    if (registry.length && sheet.options.index) {
+      for (let i = 0; i < registry.length; i++) {
+        const registration = registry[i]
+        if (
+          !registration.options.index ||
+          registration.options.index < sheet.options.index
+        ) continue
+        registry.splice(i, 0, sheet)
+        return
+      }
+    }
+
+    registry.push(sheet)
   }
 
   /**
--- a/src/SheetsRegistry.js
+++ b/src/SheetsRegistry.js
@@ -17,13 +17,14 @@ export default class SheetsRegistry {
   add(sheet) {
     const {registry} = this
 
-    if (registry.length && sheet.options.index) {
-      for (let i = 0; i < registry.length; i++) {
-        const registration = registry[i]
-        if (
-          !registration.options.index ||
-          registration.options.index < sheet.options.index
-        ) continue
+    if (!registry.length || typeof sheet.options.index !== 'number') {
+      registry.push(sheet)
+      return
+    }
+
+    for (let i = 0; i < registry.length; i++) {
+      const {options} = registry[i]
+      if (options.index > sheet.options.index) {
         registry.splice(i, 0, sheet)
         return
       }
--- a/src/SheetsRegistry.js
+++ b/src/SheetsRegistry.js
@@ -16,22 +16,20 @@ export default class SheetsRegistry {
    */
   add(sheet) {
     const {registry} = this
+    const {index} = sheet.options
 
-    if (!registry.length) {
+    if (!registry.length || index >= registry[registry.length - 1].options.index) {
       registry.push(sheet)
       return
     }
 
     for (let i = 0; i < registry.length; i++) {
       const {options} = registry[i]
-      if (options.index > sheet.options.index) {
+      if (options.index > index) {
         registry.splice(i, 0, sheet)
         return
       }
     }
-
-    // Higher priority than any sheets in registry, so push on the end
-    registry.push(sheet)
   }
 
   /**
--- a/src/backends/DomRenderer.js
+++ b/src/backends/DomRenderer.js
@@ -85,8 +85,8 @@ export default class DomRenderer {
           const sheet = registry[i]
           if (
             !sheet.attached ||
-            !sheet.options.index ||
-            sheet.options.index < index
+            typeof sheet.options.index !== 'number' ||
+            sheet.options.index <= index
           ) continue
           anchorEl = sheet.renderer.element
           break
@@ -96,9 +96,9 @@ export default class DomRenderer {
       // Otherwise insert after the last attached
       if (!anchorEl) {
         for (let i = registry.length - 1; i >= 0; i--) {
-          const registrySheet = registry[i]
-          if (registrySheet.attached) {
-            anchorEl = registrySheet.renderer.element.nextElementSibling
+          const sheet = registry[i]
+          if (sheet.attached) {
+            anchorEl = sheet.renderer.element.nextElementSibling
             break
           }
         }
@@ -107,9 +107,9 @@ export default class DomRenderer {
 
     if (!anchorEl) {
       // Try find a comment placeholder if registry is empty
-      for (let i = 0; i < document.head.childNodes.length; i++) {
-        const el = document.head.childNodes[i]
-        if (el.nodeType === 8 && el.nodeValue === 'jss') {
+      for (let i = 0; i < this.head.childNodes.length; i++) {
+        const el = this.head.childNodes[i]
+        if (el.nodeValue === 'jss') {
           anchorEl = el
           break
         }
--- a/src/backends/DomRenderer.js
+++ b/src/backends/DomRenderer.js
@@ -18,6 +18,7 @@ export default class DomRenderer {
     this.head = document.head || document.getElementsByTagName('head')[0]
     this.element = element || document.createElement('style')
     this.element.type = 'text/css'
+    this.element.setAttribute('data-jss', '')
     if (media) this.element.setAttribute('media', media)
     if (meta) this.element.setAttribute('data-meta', meta)
   }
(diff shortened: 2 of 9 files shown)


nathanmarks:
@cvle you may also be interested in taking a look at this, I know you were after this feature too

src/SheetsRegistry.js
```diff
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
    * @api public
    */
   add(sheet) {
-    this.registry.push(sheet)
+    const {registry} = this
+
+    if (registry.length && sheet.options.index) {
```
kof: won't work if index is 0, use != null in such cases

src/SheetsRegistry.js
```diff
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
...
+      for (let i = 0; i < registry.length; i++) {
+        const registration = registry[i]
```
kof: const sheet = registry[i] ?

kof: oh I see, it conflicts with outer sheet const, maybe call the first sheet "newSheet" ?

kof: actually even better - you can take directly options out

```js
const {options} = registry[i]

File: src/SheetsRegistry.js [hunk 3]
```diff
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
...
+        if (
+          !registration.options.index ||
```
kof: here again, it can be 0 ...

src/SheetsRegistry.js
```diff
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
...
+          registration.options.index < sheet.options.index
+        ) continue
+        registry.splice(i, 0, sheet)
```
kof: What about the case where 2 sheets have the same index, I think the order of insertions should be the same as they come, right now, you would insert sheets as they arrive before the previous one, if I understand correctly from reading ...

tests/integration/priority.js
```diff
@@ -0,0 +1,152 @@
+import expect from 'expect.js'
```
kof: I put all tests which touch the dom functional. Normally I try to separate integration and functional tests.

src/SheetsRegistry.js
```diff
@@ -15,7 +15,21 @@ export default class SheetsRegistry {
...
+        return
+      }
+    }
+
+    registry.push(sheet)
   }
```
kof: Here is an untested sample of the same code, how I would write it, I think it is a bit easier to read because of early returns, less indentation, less conditions:

```js

const {registry} = this

if (!registry.length || sheet.options.index != null) {
  registry.push(sheet)
  return
}

for (let i = 0; i < registry.length; i++) {
  const {options} = registry[i]
  if (options.index >= sheet.options.index) {
    registry.splice(i, 0, sheet)
    return
  } 
}

```

kof: still the case where we have same index and need to preserve the order is not implemented

src/backends/DomRenderer.js
```diff
@@ -18,6 +18,7 @@ export default class DomRenderer {
     this.head = document.head || document.getElementsByTagName('head')[0]
     this.element = element || document.createElement('style')
     this.element.type = 'text/css'
+    this.element.setAttribute('data-jss', '')
```
kof: good idea.

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+    const {registry} = jss.sheets
+
+    if (registry.length > 1) {
+      // Try to insert by index if set
+      if (typeof index === 'number') {
+        for (let i = 0; i < registry.length; i++) {
+          const sheet = registry[i]
+          if (
+            !sheet.attached ||
+            !sheet.options.index ||
```
kof: 0 case

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+            sheet.options.index < index
+          ) continue
+          anchorEl = sheet.renderer.element
+          break
+        }
+      }
+
+      // Otherwise insert after the last attached
```
kof: We could also safe a ref to the last attached sheet in the previous loop, not sure if it is worth the optimization though ...

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+      if (!anchorEl) {
+        for (let i = registry.length - 1; i >= 0; i--) {
+          const registrySheet = registry[i]
```
kof: why not just `const sheet = registry[i]`?

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+          if (registrySheet.attached) {
+            anchorEl = registrySheet.renderer.element.nextElementSibling
```
kof: What happens if there is no nextElementSibling, but there are previous sheets, so insertion should be at the latest position ...? Maybe a test would be good.

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+          }
+        }
+      }
+    }
+
+    if (!anchorEl) {
+      // Try find a comment placeholder if registry is empty
+      for (let i = 0; i < document.head.childNodes.length; i++) {
+        const el = document.head.childNodes[i]
+        if (el.nodeType === 8 && el.nodeValue === 'jss') {
```
kof: if el.nodeValue is 'jss' we basically should not care about nodeType ....

src/backends/DomRenderer.js
```diff
@@ -71,7 +72,51 @@ export default class DomRenderer {
...
+          anchorEl = el
+          break
+        }
+      }
+    }
+
+    this.head.insertBefore(this.element, anchorEl)
   }
 
   /**
```
kof: I would try to split some of this code into separate functions if it doesn't hurt performance too much, should be optimized through inlining by the runtime.

tests/utils.js
```diff
@@ -38,9 +38,14 @@ export function computeStyle(className) {
...
-export function reset() {
-  jss.plugins.registry = []
-  jss.sheets.registry.forEach(jss.removeStyleSheet, jss)
+export function reset(jssInstance = jss) {
+  jssInstance.plugins.registry = []
+
+  const numSheets = jssInstance.sheets.registry.length
+
+  for (let i = 0; i < numSheets; i++) {
+    jssInstance.removeStyleSheet(jssInstance.sheets.registry[0])
```
kof: good catch.

kof:
Ok, I am through, lets go through my comments, when ready I will make another review and go deeper into tests.

kof:
@nathanmarks regarding performance, you can try running `npm run bench` to see if your version is slower than master, if there are no good bench for this specific part, you can add them. I always run same benchmark on different branches (or stash changes) to see if there is perf degradation.

src/backends/DomRenderer.js
nathanmarks: @kof it returns `null` which works in `insertBefore` and will just append at the end.

nathanmarks:
@kof We can certainly add a benchmark test (I did run the existing ones) -- one would expect that adding with an `index` set will be very _slightly_ slower, but should not be a problem.

src/backends/DomRenderer.js
nathanmarks: I don't think it will make any real difference in perf, but will make the control flow more difficult to reason with.

src/SheetsRegistry.js
nathanmarks: @kof Yep, I will fix this -- I had initially set this function up using `1+` input before our prev discusson regarding valid range.

src/SheetsRegistry.js
nathanmarks: If we change the `<` to `<=` it will change that.

tests/integration/priority.js
nathanmarks: @kof No problem, easy to move. I set my tests up a little differently and generally do unit+integration and toss these in with integration. Sorry for the mixup!

src/SheetsRegistry.js
nathanmarks: @kof Preserve order of separate sheets with the same index regardless of the render order? It makes sense to me to just append them to the end of sheets with the same index.

nathanmarks:
I'll add `0` cases to the tests as well.

tests/utils.js
kof: a little bit shorter implementation would be

```js
jss.sheets.registry.slice(0).forEach(jss.removeStyleSheet, jss)
```

nathanmarks:
I've fixed some of the outstanding issues.

Still to do:

- [ ] Benchmarks. Do you have a specific setup in mind? ops/second for attaching/detaching sheets isn't a great measure without a target that represents something meaningful in practical use.
- [ ] Break down `attach()` into smaller functions with a single responsibility

src/SheetsRegistry.js
```diff
@@ -15,7 +15,22 @@ export default class SheetsRegistry {
...
+
+    for (let i = 0; i < registry.length; i++) {
+      const {options} = registry[i]
+      if (options.index > sheet.options.index) {
+        registry.splice(i, 0, sheet)
+        return
+      }
+    }
+
+    registry.push(sheet)
```
kof: you sure you didn't forget to remove this line? If not, which case is it?

kof:
Looks good, besides of one latest comment I could merge it.

kof:
> Benchmarks. Do you have a specific setup in mind? ops/second for attaching/detaching sheets isn't a great measure without a target that represents something meaningful in practical use.

I have no other idea than attaching/detaching, if we create a bench for "finding the anchor element" specifically, we have nothing to compare with, absolute number will be very high. I would try to bench attaching sheets with variable index values. Most probably there will be no difference to the current implementation, then it probably doesn't make sense to have such a test, because attaching of sheets is anyways not a bottleneck I assume.

src/SheetsRegistry.js
nathanmarks: @kof the `for` loop will leave it hanging if it belongs at the end.

 We could add an extra conditional if we're on the last item but it hasn't been inserted (so just after line 30). Or instead check the last sheet in the registry beforehand to avoid the iteration all together.

Another option is to use a binary search to find the insertion index, but I don't think there would be any point.

kof: ah ok, just add a comment, because it is not easy to guess

nathanmarks:
I will run some tests just to make sure there isn't any unexpected slowdown (I'm thinking ~30-50 sheets with some attaching in various positions). I've started using this branch in a project and so far the feature is working flawlessly.

Do you want me to add some documentation for it too?

nathanmarks:
@kof There is one use case I would like to add some tests for so that the functionality is well defined: What if you have sheets without an index, and some sheets with an index? 

My gut tells me that any sheet with an index set should be considered higher priority, but one might argue that `< 0` should come before sheets (therefore treating no index `=== 0`). WDYT?

kof:
> @kof There is one use case I would like to add some tests for so that the functionality is well defined: What if you have sheets without an index, and some sheets with an index?

I would even say there should be always an index defined, default should be 0 as user creates a sheet. Also you don't need typeof checks later.

cvle:
I agree with @kof, I think index should always be defined and default to 0, allowing e.g. reset sheets with an index < 0.

kof:
> Do you want me to add some documentation for it too?

Would be nice)

kof:
you have got some linter errors)

kof:
merged

Issue #285, opened 2016-08-15
[jss-nested] Unexpected order in generated style


cvle:
This stylesheet generates classes in wrong order:

    const stylesheet = {
      button: {
        "&:hover": {
          background: "rgba(255, 255, 255, 0.1)",
        },
        background: "none",
      },
      // This disables hover on mobile devices.
      "@media (hover:none)": {
        button: {
          "&:hover": {
            background: "none",
          },
        },
      },
    };

Output:

    .button-2260332263 {
      background: none;
    }
    @media (hover:none) {
      .button-2260332263:hover {
        background: none;
      }
    }
    .button-2260332263:hover {
      background: rgba(255, 255, 255, 0.1);
    }

Should be:

    .button-2260332263 {
      background: none;
    }
    .button-2260332263:hover {
      background: rgba(255, 255, 255, 0.1);
    }
    @media (hover:none) {
      .button-2260332263:hover {
        background: none;
      }
    }

Workaround using `!important`:

    const stylesheet = {
      button: {
        background: "none",
        "&:hover": {
          background: "rgba(255, 255, 255, 0.1)",
        },
      },
      // This disables hover on mobile devices.
      "@media (hover:none)": {
        button: {
          "&:hover": {
            background: "none !important",
          },
        },
      },
    };

kof:
Yeah, the problem is that plugins are processed in a separate loop since c40c616fc0b5c68eabb1e34f842229a40f674c08

kof:
This has been done to allow referencing rules from nested rules.

kof:
Now the problem is that we need to put nested rules in the old order, which is actually not that easy ... it requires either to sort the rules map on every nested rule or to introduce additional mechanism for saving the order of rules and using this order when generating css.

cvle:
I removed the workaround because it's not working, still trying to find a way to get around this.

kof:
adding an !important would work ...

cvle:
Confirmed. Pretty ugly but works for me, at least for now. Adding a workaround above.

kof:
Released.

PR #306, opened 2016-09-05
Run plugins tests + fix a bug #285

kof:
I mixed those 2 things, because I wanted to how running plugins in the main repository will work out to fix a bug in a plugin repository.

The problem was that to fix a bug in jss-nested I needed to fix JSS first and then to fix jss-nested. I wanted to be able to do that from one test suite and this was possible.

### DIFF
--- a/src/StyleSheet.js
+++ b/src/StyleSheet.js
@@ -1,6 +1,5 @@
-import {isEmptyObject} from './utils'
-import createRule from './createRule'
 import findRenderer from './findRenderer'
+import RulesContainer from './RulesContainer'
 
 /**
  * StyleSheet model.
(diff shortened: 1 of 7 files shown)


kof:
@nathanmarks please review

kof:
- [ ] add documentation for index option

kof:
@nathanmarks to give you an overview: in order to be able to insert rules at specific indexes in the sheet as well as in the ConditionalRule, I needed to extract the rules container into  a separate class called RulesConntainer which is now used in both StyleSheet as well as ConditionalRule. So now nested  rules can be inserted at the right position in the style sheet AND within the @-rule.

nathanmarks:
Sounds good 👍 I'll review after I get home today

src/StyleSheet.js
```diff
@@ -30,22 +29,26 @@ export default class StyleSheet {
     this.rules = Object.create(null)
     // Used to ensure correct rules order.
     this.rulesIndex = []
-    this.classes = Object.create(null)
+    this.classes = this.options.classes = Object.create(null)
```
nathanmarks: I'm wondering if there is a better way to implement this. It doesn't feel completely natural having the classes reference on the options object. I feel like the classes should be retrieved from the rules container rather than passing around the object reference like this. Is there ever a scenario where the classes are provided to the RulesContainer? If not, this is a bit of a stretch for the usage of options.

kof: Does classes mutation from RulesContainer this way sound better `this.options.sheet.classes` ?

kof: Another option: create classes object in RulesContainer constructor and then reference it in StyleSheet:

```js
this.rules = new RulesContainer(this.options)
this.classes = this.rules.classes
```

kof: Oh, can't do the last one, classes is a hash of all classes for the sheet, it needs to be accessed from all RulesContainer instances. By passing a classes object I basically pass a registry for class names to all RulesContainer. Knowing that it is an internal implementation detail, is it actually that bad?

kof: I know it is  a bad thing to pass an object and then modify it in some random places ... however if an object has a very clear interface and is basically a registry, just without methods for performance reasons, not sure this is actually bad.

kof: Another option is to implement a setClasses method in RulesContainer and set the classes from the parent, but this is a bit of an overhead for this purpose.

nathanmarks: As an implementation detail I think it is okay. I understand what you are saying about the additional overhead.

Quick Q, when you say:

> Oh, can't do the last one, classes is a hash of all classes for the sheet, it needs to be accessed from all RulesContainer instances.

What do you mean by **all** instances? I thought that StyleSheet -> RulesContainer was a 1:1 relationship.

nathanmarks: PS: `flow` on weak mode will **not** let you add additional properties on an object after it is initialized (like the options object) unless the object is created empty.

kof: > What do you mean by all instances?

RulesContainer is used from sheet as well as from a ContainerRule

> flow on weak mode will not let you add additional properties

Good to know, lets tackle this step by step?

kof:
merged

PR #309, opened 2016-09-08
Added lint-staged with autofix config for *.js

### LINKED ISSUE #303: Use lint-staged in all repositories
To avoid forgetting to lint.

okonet:
Closes #303 

@kof it says "in all repositories". Can you add a to-do list of where it makes sense to integrate here?

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,10 @@
     "karma-safari-launcher": "^1.0.0",
     "karma-sourcemap-loader": "^0.3.5",
     "karma-webpack": "^1.7.0",
+    "lint-staged": "^3.0.1",
     "lodash.assign": "^4.0.9",
     "mocha": "^2.4.5",
+    "pre-commit": "^1.1.3",
     "webpack": "^1.12.2"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -76,10 +78,18 @@
...
-    "version": "node -e \"console.log(require('./package.json').version)\""
+    "version": "node -e \"console.log(require('./package.json').version)\"",
+    "lint-staged": "lint-staged"
   },
   "dependencies": {
     "is-browser": "^2.0.1",
     "murmurhash-js": "^1.0.0"
-  }
+  },
+  "lint-staged": {
```
kof: Shouldn't this use  https://github.com/cssinjs/jss/blob/master/package.json#L72 ?

kof:
I think for now jss-nested and jss-expand, other plugins need to migrate to mocha first.

package.json
okonet: No, since lint-staged will pass only staged files as an argument to the linter: https://github.com/okonet/lint-staged/blob/master/src/findBin.js#L7. For that reason it _must_ be run without paths as argument.

okonet: Use `.eslintignore` in case you want to ignore some files.

kof: Sure, another question, it seems like a good name in scripts would be:

```
lint -> lint:all
linst-staged -> lint:staged
```
Makes sense?

kof: Or maybe `npm run lint` should stay as is, because we are all used to call it this way ...

kof:
Oh sorry, I messed up things, mocha has nothing to do with linting ... we need this actually in all repos that contain linter ..

package.json
okonet: Not sure what you mean by the last one. 

> `linst-staged` -> `lint:staged`

This is a matter of preference. I'll update it to match your style.

kof:
merged

okonet:
@kof integrating with all repositories is cumbersome. What do you think of integrating https://github.com/FormidableLabs/builder in order to make it more controllable?

kof:
Sounds interesting, need to take a deeper look at it!

kof:
We also have the problem of having too many same dependencies across repositories .... this will result in too much work of updating them ...

okonet:
https://github.com/cssinjs/jss/issues/310

PR #319, opened 2016-09-18
fix addRule insertion order from plugins, fix @media insertion

### LINKED ISSUE #318: [jss-nested] rules order in attached sheet when using .addRule
Currently when a sheet is already deployed and attached and addRule is used, the order is not correct. Nested rules are added before their parents. Index option is not used in StyleSheet.addRule method in this case. It is not possible to insert rules at any position like in an array, because it will throw `DomRenderer.js:189 Uncaught IndexSizeError: Failed to execute 'insertRule' on 'CSSStyleSheet': The index provided (1) is larger than the maximum index (0).`

kof:
Fixes #318

When sheet is attached and deployed and `.addRule`:
- is used from a plugin, order is not preserved
- at-rules are not inserted using `insertRule` api in IE

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "expect.js": "^0.3.1",
     "json-loader": "^0.5.4",
     "jss-nested": "2.x",
+    "jss-extend": "2.x",
     "karma": "^1.1.1",
     "karma-benchmark": "^0.6.0",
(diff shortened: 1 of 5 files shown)


kof:
@nathanmarks check this out

PR #328, opened 2016-10-05
Start Flowing the things

wldcordeiro:
Start of Flow type annotations for the project. Adds Flow, and annotates the following

* PluginsRegistry
* Main project index
* SimpleRule

The annotations will receive changes as the project gets more typed to avoid bare `Object` types and use proper type aliases that are more descriptive.


.flowconfig
```diff
@@ -0,0 +1,13 @@
+[ignore]
+.*/tests/.*
+.*node_modules/.*
+
+[include]
+src/
+
+[libs]
+interfaces/
```
wldcordeiro: We don't have this dir yet, but we can populate it with annotations for 3rd party lib stuff when we get to files that use them.

.flowconfig
```diff
@@ -0,0 +1,13 @@
...
+
+[options]
+suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
+suppress_comment= \\(.\\|\n\\)*\\$FlowIssue
```
wldcordeiro: If there is a Flow bug that causes an error use this one.

.flowconfig
```diff
@@ -0,0 +1,13 @@
...
+.*node_modules/.*
+
+[include]
+src/
+
+[libs]
+interfaces/
+
+[options]
+suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
```
wldcordeiro: Sometimes there is complication where you need to rewrite some of the code to guard against possibly nullable types. From reading the documentation, we don't actually need this one anymore as it's part of the defaults (must have been a change since I last read config docs heh)

package.json
```diff
@@ -79,6 +80,7 @@
     "build:min": "cross-env NODE_ENV=production webpack src/index.js dist/jss.min.js",
     "build:tests": "cross-env NODE_ENV=test webpack tests/index.js tmp/tests.js",
     "clean": "rimraf '{lib,dist,tmp}/*'",
+    "flow": "flow --show-all-errors",
```
kof: I think flow checker should run also on npm run all and  on lint-staged ...

kof:
merged

PR #332, opened 2016-10-20
Warn when using a wrong at-rule

oliviertassinari:
This warning is improving the user feedback when not using the API correctly.

> TypeError: AtRule is not a constructor

I just had the above issue with an incorrect `@raw` rule.


nathanmarks:
Nice 👍 Just a couple of lint errors to fix up

kof:
What did you use that is not supported?

oliviertassinari:
@nathanmarks Sure, should we had a unit test too?
@kof I was migrating from another CSS-in-JS engine (stylishly)

kof:
Yeah, ideally we need a test, linted and a better message: `warning(AtRule, '[JSS]: unknown rule %s', name)`

oliviertassinari:
> Yeah, ideally we need a test

We would probably need to stub `console`. I know @nathanmarks isn't keen on this idea.
That would also throw an error. We will need to add a try / catch block.

kof:
Just realized: it shouldn't be a warning, it should be a thrown Error. Something like this is not just a hint, it won't work and we have no fallback to something that could work. And this should be fixable during development and should not happen in production.

kof:
If we use a warning, we need to implement some fallback, which is probably a regular Rule instance. As it is now, the next line after warning it will throw because we still have no AtRule ...

oliviertassinari:
Right. I'm not sure what's the best approach is.
Do you want to have a fallback?
Or should we throw with a better error message than `TypeError: AtRule is not a constructor`.

kof:
Wether we need a fallback or not depends on: can this error happen in production or will this always be visible during development. If the last one  - then throw, if first one - warning and fallback.

kof:
I think your case is something that can happen only in development, right?

oliviertassinari:
For my use case, a throw would be better. For others, I don't know. What if someone is generating the rule with a variable and he mess something around?

kof:
Well this is not entirely impossible, but I can hardly imagine that in this particular case.

oliviertassinari:
let's throw then, it's gonna make the test implementation simpler 👍 . Thanks.

kof:
On the other hand, CSS never throws into jss so that the program can break. Maybe its a good general rule to stick to, no matter if it makes sense in some particular case or not. Just always render something even if it is wrong. A visual bug is better than a program that stopped working completely.

oliviertassinari:
> A visual bug is better than a program that stopped working completely

Put it that way, I can't agree more.

kof:
Then we need a warning and a fallback with a Rule instance.

oliviertassinari:
I have forgotten this PR. Any update?

kof:
thats basically fine, I will make some changes afterwards.

kof:
merged, thanks!

Issue #324, opened 2016-09-28
[jss-global] Plugin that enables global styles without using an option


kof:
We can have this syntax for global styles:

``` javascript
{
  '@global': {
    html: {globalStyles},
    button: {globalStyles}
  },
  button: {localStyles}
}
```
- It allows us to use global selectors any time, without being forced to use a separate sheet with an option `{named: false}`.
- Also it allows us to remove `{named}` option and related logic from the core. Core is always namespaced.
- It is explicit about globals, which is a good thing.

kof:
To sum up the discussion with @nathanmarks on gitter:

- we should call it @global to avoid conflicts
- we should support `@global button` case as well
- we should think of supporting @global  within other rules, so that its like nested but with less complex

kof:
Just finished removing dependency to the "named" option in all repositories.

kof:
Now will start to implement:

```
'@global': {
  html: {color: 'red'}
}
```
and

```
'@global html': {
  color: 'red'
}
```

kof:
Need  a list of all cases where @global should work, like inside of @media etc.

giuseppeg:
fwiw CSS Modules calls this `:global`

kof:
One naming conflict came up as we discussed this feature with @mxstbr: what do we call global, stuff like: `button  {color: red}`  or also stuff like `.my-button-123345 span {color: red;}`. In the first case we use a selector which applies to the entire document, in the second its a scoped selector. Is the second case still global in terms of terminology?

giuseppeg:
> Is the second case still global in terms of terminology? 

Yes and no, it will affect the entire subtree and in my opinion it shouldn't be valid or allowed otherwise we are going back to the old and not isolated css :)

kof:
I think its fine to keep the '@global' name also when nesting them.  I would see it as global in terms of what is generated after this keyword.

giuseppeg:
To clarify I think that those kind of rules are still `global`.

When you scope a selector like `button` there is a context switch, now global rules defined within this new context are relative to it.

The default context is the document or `:root` element.

mxstbr:
The reason you need this escape hatch is for the following case:

```JSX
<div dangerouslySetInnerHTML={{ __html: someHtmlString }} />
```

How do you now style the contents with JSS if you have no way of attaching class names? You can't without the global.

[...]

PR #348, opened 2016-11-05
Preparation for jss-global #324

kof:
This should wait until all plugins, which rely on it are changed as well. Mostly it is just tests.


kof:
cc @nathanmarks

src/RulesFactory.js
```diff
@@ -0,0 +1,69 @@
...
+
+import RegularRule from './rules/RegularRule'
+import SimpleRule from './rules/SimpleRule'
+import KeyframeRule from './rules/KeyframeRule'
+import ConditionalRule from './rules/ConditionalRule'
+import FontFaceRule from './rules/FontFaceRule'
+
+export default class RulesFactory {
+  constructor() {
+    this.classes = {
```
nathanmarks: If this is going to be set in the constructor, it should be extendable via an options object. No point otherwise.

src/RulesFactory.js
```diff
@@ -0,0 +1,69 @@
...
+   * Options:
+   *   - `className` pass class name if you to define it manually
+   *
+   * @param {Object} [name]
+   * @param {Object} [style] declarations block
+   * @param {Object} [options] rule options
+   * @return {Object} rule
+   * @api public
+   */
+  get(name, style = {}, options = {}) {
```
nathanmarks: Wrong method name

src/RulesFactory.js
```diff
@@ -0,0 +1,69 @@
+import warning from 'warning'
```
nathanmarks: Why not use flow for this new module?

kof: will add flow to everything new!

kof: its a draft

src/RulesFactory.js
kof: what do you mean, bad naming? any suggestions?

src/RulesFactory.js
nathanmarks: Why an option to register new classes but no ability to extend them using the constructor?

src/Jss.js
```diff
@@ -74,22 +75,33 @@ export default class Jss {
...
+    const rule = this.factory.get(name, style, options)
+
+    if (options.applyPlugins !== false) {
+      this.plugins.handleRule(rule)
+    }
+
     return rule
   }
 
+  registerRuleClass(name, cls) {
```
nathanmarks: This seems like a very low level function to expose here. Part of me feels like the plugin API should be able to handle this use case with atRules all using the same hook.

src/Jss.js
```diff
@@ -17,9 +17,10 @@ export default class Jss {
    * @see .setup()
    */
   constructor(options) {
+    this.version = __VERSION__
     this.sheets = new SheetsRegistry()
-    this.plugins = new PluginsRegistry()
-    this.version = process.env.VERSION
+    this.plugins = new PluginsRegistry(this)
+    this.factory = new RulesFactory()
```
nathanmarks: This isn't the right property name:

```js
import {create} from 'jss';

const jss = create();

jss.factory // RuleFactory...? Sounds like "JSS Factory" when you read the code which is actually `create()`.
```

src/PluginsRegistry.js
```diff
@@ -6,46 +6,59 @@
...
  */
 export default class PluginsRegistry {
-  registry: Array<Function>;
+  registry: Array<Object>
+  jss: Object
 
-  constructor(): void {
+  constructor(jss: Object): void {
     this.registry = []
+    this.jss = jss
```
nathanmarks: I think that this should be passed into an instance method instead of storing the reference. The plugin registry doesn't need to retain knowledge of the whole `jss` instance IMO.

src/RulesFactory.js
kof: didn't have a use case for it yet.

src/Jss.js
kof: do you have an api in mind?

src/Jss.js
kof: rulesFactory?

src/PluginsRegistry.js
```diff
@@ -6,46 +6,59 @@
...
-    this.runOne(rules)
+  /**
+   * Call all .onRule hooks for passed rule.
+   *
+   * @api public
+   */
+  handleRule(rule: Object): void {
+    for (let index = 0; index < this.registry.length; index++) {
+      const plugin = this.registry[index]
+      if (plugin.onRule) plugin.onRule(rule)
```
nathanmarks: An alternative to conditionally checking every plugin in the registry for the presence of a specific hook every time, is to register the hooks in `.use`. This can be done by having an object holding an array for each hook, eg:

pseudo code:

```js
hooks = {
  onSheet: [],
  onRule: [],
};

use(plugin) {
  Object.keys(plugin).forEach((key) => {
    if (this.hooks[key]) {
      this.hooks[key].push(plugin[key]);
    }
  });
}

handleSheet(sheet) {
  this.hooks.onSheet.forEach((n) => n(sheet));
}
```

src/RulesFactory.js
```diff
@@ -0,0 +1,69 @@
...
+    this.classes = {
+      '@charset': SimpleRule,
+      '@import': SimpleRule,
+      '@namespace': SimpleRule,
+      '@keyframes': KeyframeRule,
+      '@media': ConditionalRule,
+      '@supports': ConditionalRule,
+      '@font-face': FontFaceRule
+    }
+    this.regExp = this.buildRegExp()
```
nathanmarks: Why use a constructor for these use cases? Stylistic preference? I noticed that we have `stage-0` enabled, and class properties are already `stage-2`:

```js
class RulesFactory {
  classes = {
    '@charset': SimpleRule,
  }

  regExp = this.buildRegExp()
}
```

src/PluginsRegistry.js
kof: Oh, I could actually pass jss as a second argument to the .use method. This would work for now, not sure though this would work forever, next time it is needed in another method its not any more nice. Also kinda not very nice for the .use method to require this second argument additionally to the plugin. 

I agree that having the jss instance all over the place is not very nice, but I don't see a real good another way. Also which problems can having .jss property bring? Harder to unit test?

src/RulesFactory.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+   * Build regexp for matching rules.
+   *
+   * @return {RegExp}
+   * @api private
+   */
+  buildRegExp() {
```
nathanmarks: Instead of having a private method like this on the prototype, why not just have it as a function outside of the class?

```js
function buildClassRegExp(classes: Object): RegExp  {
  return new RegExp(`^${Object.keys(classes).join('|')}`)
}
```

src/RulesFactory.js
nathanmarks: Yeah, we shouldn't use `get` for something which **creates**. `get` is the `read` in `CRUD`. This should be `createRule` or `create`.

src/PluginsRegistry.js
kof: yep, thats true, do you like this approach in terms of readability or performance?

src/StyleSheet.js
```diff
@@ -20,8 +18,7 @@ import RulesContainer from './RulesContainer'
  * @api public
  */
 export default class StyleSheet {
-  constructor(rules, options) {
-    const named = options.named == null ? true : options.named
+  constructor(styles, options) {
```
nathanmarks: Why the change to `styles`?

src/RulesFactory.js
kof: can totally be done!

src/StyleSheet.js
```diff
@@ -128,40 +126,40 @@ export default class StyleSheet {
...
+   * Create styles, will render also after stylesheet was rendered the first time.
    *
-   * @param {Object} rules name:style hash.
+   * @param {Object} styles name:style hash.
    * @param {Object} [options]
    * @return {Array} array of added rules
    * @api public
    */
-  addRules(rules, options) {
+  addRules(styles, options) {
```
nathanmarks: Felt very natural having the first arg to `addRules` named `rules`. What's the change for?

src/StyleSheet.js
```diff
@@ -128,40 +126,40 @@ export default class StyleSheet {
   }
 
   /**
-   * Create rules, will render also after stylesheet was rendered the first time.
+   * Create styles, will render also after stylesheet was rendered the first time.
```
nathanmarks: Isn't `Create rules` more specific? I'm curious about this change.

src/RulesFactory.js
kof: well its actually a classical factory get method, which is supposed to instantiate and return. I know what you mean, I had it this way at first. But then I looked up the factory pattern and they use `get`.

src/rules/RegularRule.js
```diff
@@ -8,7 +8,7 @@ const {parse, stringify} = JSON
  * @api public
  */
 export default class Rule {
```
nathanmarks: should be `RegularRule`

src/StyleSheet.js
kof: For naming consistency. We call user style definitions styles, they become rules when they are wrapped into a rule class instance.

src/StyleSheet.js
kof: same here, addRules accepts styles object and creates rules out of it, they are not rules at the moment user passes them, seemed better to me as we always communicate it as styles to the user land.

src/StyleSheet.js
kof: hmm actually in this case, its create rules, we take styles and create rules))))

src/rules/RegularRule.js
kof: yep

src/Jss.js
nathanmarks: Like, Ideally the plugin API should support the following functionality:

- Plugin can parse the rule name before anything is created
- If it's a global scope selector, the plugin can set the selector as is necessary
- If it is a global scope nested object of rules, the plugin can set the rule type, so `jss` knows it is not a regular CSS rule
- When the nested rules are parsed, the plugin can check if the parent is a globally scoped rule and manipulate the selectors as is necessary

nathanmarks: I think that this method (`registerRuleClass`) and the injectable `generateClassName` function are unicorns that feel a little bolted on and out of place.

kof: yep, doesn't feel to me nice as well, will rethink the plugin api to avoid it.

src/RulesFactory.js
nathanmarks: Those are instance properties not static.

src/RulesFactory.js
kof: I think it implies that an instance is not always created, it might be cached and just returned.

nathanmarks: I've seen create or createxxxx used in loads of factory patterns, including textbook examples etc. But I digress... Up to you.

src/Jss.js
nathanmarks: That works.

src/PluginsRegistry.js
nathanmarks: That's a good question. It depends on a number of variables including the size of the application. I need to test to get specifics but I'm interested in finding out.

kof:
I think it makes sense, because different hooks are called at different times, Its nicer to have an overview like this. Perf difference will be minimal but yeah, to call the specific hook iterator needs to iterate less, but as the number is just a bunch, it won't matter that much.

src/RulesFactory.js
kof: ok changed.

src/plugins.js
```diff
@@ -0,0 +1,27 @@
...
+  '@font-face': FontFaceRule
+}
+
+/**
+ * Generate plugins which will register all rules.
+ */
+const plugins = Object.keys(classes).map((key) => {
+  const re = new RegExp(`^${key}`)
+  const onCreate = (name, style, options) => (
+    re.test(name) ? new classes[key](name, style, options) : null
```
nathanmarks: Why not just, `name.indexOf(key) === 0`? Don't have to create the regexp every time that way

kof:
To be sure I just made some benchmarking

https://jsperf.com/indexof-vs-substr-vs-regex-at-the-beginning-3 <https://jsperf.com/indexof-vs-substr-vs-regex-at-the-beginning-3>

On Chrome: regex and indexof perform the same, substr. is twice faster. I assume my bench is not good enough and chrome tricks it out with some optiomizations

Safari, Firefox, IE11 - regex is the fastest.

What you probably oversaw is that the regex is created just once for each plugin, onCreate doesn't generate the regex each time. Hot part is only onCreate, plugin genearation happens just once for each plugin.

nathanmarks:
@kof Yeah, I totally overlooked that. Thanks for doing the pef.

I did some additional testing myself using a smaller string. In Firefox and Safari, the regex was a lot faster. 

I was surprised. I didn't know that it was faster for small strings. TIL.

kof:
merged

PR #350, opened 2016-11-07
Add sheet meta value to className generation

### LINKED ISSUE #347: Unexpected specificity when 2 equal rules are in 2 different sheets
As discussed with @nathanmarks, it is possible that 2 equal rules exist in 2 different sheets rendered at different times. Class name generation uses rules style object at the moment. It might be a problem when:

1. a: {color: 'red', float: 'left'}
1. b: {float: 'right'}
1. `<span class="a b" />`  - b is supposed to overwrite float prop in "a".
1. in different sheet  a: {color: 'red', float: 'left'} rule co-exist
1. once the second sheet is rendered, it will have a higher specificity based on source order, it will have the same selector as the first rule and so will replace the first rule
1. after that b rule will no longer overwrite float property of the "a" rule.

### LINKED ISSUE #349: Use options.meta when generating class name
cc @nathanmarks

nathanmarks:
Resolves #349. Also helps somewhat resolve #347.

**What?**
This PR introduces a chang that prefixes classNames with the sheet meta value (if provided). This provides better readability of classNames in the DOM for debugging purposes in libraries such as `react-jss` and `jss-theme-reactor` that use the `meta` property for component names. It also goes some way towards solving #347 by providing a mechanism to namespace selectors other than by a hash of their style declaration (which can be duplicated across sheets, along with the rule name).

**How?**
`generateClassName` checks if the sheet has a meta value set. If it does, it prepends it to the className separated by a `_`.

I added a couple of tests and updated the test util `generateClassName` stub.

Check this test out for an example: https://github.com/cssinjs/jss/compare/master...nathanmarks:meta-classname-prefix?expand=1#diff-f5956f248a2c73190de0534eeb61390bR49

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -8,8 +8,16 @@ import createHash from 'murmurhash-js/murmurhash3_gc'
  * @return {String}
  */
 export function generateClassName(str, rule) {
+  const {name, options: {sheet}} = rule
   const hash = createHash(str)
-  return rule.name ? `${rule.name}-${hash}` : hash
+
+  let className = name ? `${name}-${hash}` : hash
+
+  if (sheet && sheet.options.meta) {
+    className = `${sheet.options.meta}_${className}`
+  }
+
+  return className
 }
 
 /**
--- a/tests/integration/sheet.js
+++ b/tests/integration/sheet.js
@@ -40,6 +40,16 @@ describe('Integration: sheet', () => {
       })
       expect(sheet.classes.a).to.be('a-id')
     })
+
+    it('should create rule classNames using the rule name', () => {
+      const sheet = jss.createStyleSheet({bar: {}})
+      expect(sheet.classes.bar).to.be('bar-id')
+    })
+
+    it('should create rule classNames using the rule name and meta value', () => {
+      const sheet = jss.createStyleSheet({bar: {}}, {meta: 'foo'})
+      expect(sheet.classes.bar).to.be('foo_bar-id')
+    })
   })
 
   describe('sheet.getRule()', () => {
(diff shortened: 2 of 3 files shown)


src/utils.js
```diff
@@ -8,8 +8,16 @@ import createHash from 'murmurhash-js/murmurhash3_gc'
...
  */
 export function generateClassName(str, rule) {
+  const {name, options: {sheet}} = rule
   const hash = createHash(str)
-  return rule.name ? `${rule.name}-${hash}` : hash
+
+  let className = name ? `${name}-${hash}` : hash
+
+  if (sheet && sheet.options.meta) {
+    className = `${sheet.options.meta}_${className}`
```
kof: 2 questions I got:
1. Me are not enforcing anywhere the format of meta, user could pass 2 words.
2. I personally dislike undercores))), can we  use always "-"?
3. What if user passes meta as component name (for e.g. react-jss does it), it will look like 'Button-label-123456' which is also not very nice. I think we need to normalize it to lower case.

kof: I think we should use rule.name and sheet.options.meta to generate the hash  additionally to the style object. This way we avoid all class compliance issues.

kof: `createHash(name + options.meta + str)`

kof: Its simple to implement and will solve 99% of the issues. If one wants to solve all 100%, he can make sure meta is completely unique.

kof: In the context of material-ui, 99% should be already enough, no?

nathanmarks: @kof Let's do that.

I'm still going to drop a custom method in for `jss-theme-reactor` so that I can output pretty selectors for debugging with the component/sheet name visible. But in terms of fixing the issue in `jss` core, this is a great solution 👍

Issue #357, opened 2016-11-14
jss.version is broken


kof:
Lately I started to use cross-env, but that broke current way for version env variable replacement.

I was already looking to replace the current solution by babel-plugin-inline-package-json. But this one only supports "require" calls, no imports. 

I am happy to any suggestions or maybe someone can fix that babel-plugin, or we just use a require in this case but its not nice.

umidbekkarimov:
Here you go http://astexplorer.net/#/O0SIDMBBNi

kof:
The problem is to replace the version using package.json so that we don't have to maintain it in 2 places.

kof:
Ah so you suggest to use an own transform and require the package.json manually ... that could work as well.

umidbekkarimov:
Yep, I can create PR, tomorrow.

kof:
Cool, where do you put that transform function?

kof:
fixed, will be released with the next major release.

PR #358, opened 2016-11-15
Implement `babel-plugin-inline-version`

umidbekkarimov:
See #357

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,6 +4,7 @@
     "es2015"
   ],
   "plugins": [
+    "./tools/babel-plugin-inline-version",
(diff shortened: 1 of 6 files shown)


kof:
Was thinking about publishing it separately and checking the npm namespace https://www.npmjs.com/package/babel-plugin-inline-version

umidbekkarimov:
@kof I had same idea but didn't wanted to bother about finding proper `package.json` file from parent directories

kof:
Yes, I see the implementation is almost the same. Its better for this project to not have it directly here, going for thats depedency if thats fine for you.

PR #373, opened 2016-11-30
use internal SheetsRegistry in DOMRenderer

jedwards1211:
(not tested yet) this is an example of how an internal `SheetsRegistry` can be used in DOMRenderer to handle sheet indexes.  With this change we could remove the `SheetsRegistry` from `jss` instances without breaking sheet index behavior.


kof:
Closing this, will be integrated in this branch https://github.com/cssinjs/jss/tree/remove-named

Issue #374, opened 2016-12-04
[jss-compose] Conflict with other plugins


typical000:
Suppose we use order as here https://github.com/cssinjs/jss/blob/master/docs/plugins.md
We write such code:
```````````````````````````````````````````````js
linkActiveNoChildren: {
    composes: ['$linkNoChildren', '$linkActive']
}
```````````````````````````````````````````````
And, as result we receive browser warining:
> vendor.bundle.js:25585 Warning: [JSS] Referenced rule is not defined. 
.linkActiveNoChildren-3174634915 {
    composes: $linkNoChildren,$linkActive;
}

But, instead of warning jss-compose must generate classes like `this.props.sheet.linkActiveNoChildren === 'linkNoChildren-2021460718 link-2652749541 linkActive-1277801469 link-2652749541'`

Possibly jss-compose conflicts with jss-expand but i'm not sure. As a solution we can try to move jss-compose up in order and see what happens

kof:
Can you show the full sheet you are using?

typical000:
```````````````````````````````````````````````js
export default {
  // Links
  link: {
    color: vars.sidebarColor,
    padding: [1.5, 6, 1.5, 2.5],
    textDecoration: 'none',
    fontWeight: 300,
    display: 'block',
    borderTop: vars.border(vars.sidebarBorder),
    borderBottom: vars.border(vars.sidebarShadow),
    backgroundColor: vars.sidebarBg,
    transition: vars.transition(),
  },
  linkActive: {
    composes: '$link',
    backgroundColor: vars.sidebarBgActive,
  },
  // Links without childrens
  linkNoChildren: {
    composes: '$link',
    paddingRight: 2.5
  },
  linkActiveNoChildren: {
    composes: ['$linkNoChildren', '$linkActive']
  },
}
```````````````````````````````````````````````

kof:
Looks like it should work, need to try out.

typical000:
Removed all unused stuff. This code throws in repl the same warning:
```````````````````````````````````````````````````````js
return {
  link: {
    color: 'red',
  },
  linkActive: {
    composes: '$link',
  },
  // Links without childrens
  linkNoChildren: {
    composes: '$link',
  },
  linkActiveNoChildren: {
    composes: ['$linkNoChildren', '$linkActive']
  },
};
```````````````````````````````````````````````````````
In console:

> browser.js:49 Warning: [JSS] Referenced rule is not defined. 
[...]

PR #378, opened 2016-12-06
#374 Fix jss-compose conflict with jss-expand

typical000:
Solves https://github.com/cssinjs/jss/issues/374
And is connected with jss-preset-default PR fix https://github.com/cssinjs/jss-preset-default/pull/4


kof:
merged

PR #383, opened 2016-12-11
Add sheet meta value to className hash

nathanmarks:
Follow up to #350 .

From #374 
>  it is possible that 2 equal rules exist in 2 different sheets rendered at different times. Class name generation uses rules style object at the moment.

Addresses this by including the sheet meta (if provided) in the string that gets hashed.

@kof


kof:
merged

PR #415, opened 2017-02-03
Use files field in package.json instead of .npmignore

sapegin:
That’s what you publish to npm now:

image 2017-02-03 at 7 32 54 pm

Most of this files aren‘t necessary. I’m not sure about flow files though.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,0 @@
-src
-benchmark
-tmp
-tests
-.env
-coverage
--- a/package.json
+++ b/package.json
@@ -107,6 +107,10 @@
     "murmurhash-js": "1.0.0",
     "warning": "3.0.0"
   },
+  "files": [
+    "dist",
+    "lib"
+  ],
   "lint-staged": {
     "./src ./tests ./benchmark ./*.js": [
       "eslint",
--- a/package.json
+++ b/package.json
@@ -109,7 +109,8 @@
   },
   "files": [
     "dist",
-    "lib"
+    "lib",
+    "flow-typed"
   ],
   "lint-staged": {
     "./src ./tests ./benchmark ./*.js": [


kof:
I am not sure what to do with docs.
flow-typed needs to be in the package right now

kof:
also changelog is nice to have in the package

kof:
and license

kof:
and readme)))

sapegin:
`flow-typed` or `.flowconfig` or both? 

I never publish docs but some people do. So it‘s up to you ;-)

Readme, changelog and License should be added automatically: https://docs.npmjs.com/files/package.json#files

kof:
Are there any reasoning about what to include and what not to? I found it useful to be able to read quickly the docs right from node_modules. But maybe there is a general consensus to only include code?

sapegin:
I don’t know about that. But thing like tests, linter configs, etc. shouldn’t bloat `node_modules` ;-)

sapegin:
I’ve heard from someone that they read docs in `node_modules` while they are offline ;-)

kof:
Maybe, sometimes tests are saying more than docs …

kof:
I totally agree regarding everything dev related, not sure about tests though…

kof:
Ok so lets clarify what needs to be included for webpack 2, otherwise readme.md, license, changelog, flow-typed, dist, lib should be included I think.

sapegin:
As I understand you need to use `module` field in package.json, it should point to ES6 version of your code. You can’t use anything ES6+ here (you’ll have to transpile) and `import`s should be there.

sapegin:
http://survivejs.com/webpack/advanced-techniques/authoring-packages/#understanding-package-json-

sapegin:
I’ve added `flow-typed`. Readme.md, license, changelog are defaults. Webpack 2 probably needs some experimentation.

sapegin:
Looks like I was wrong: you need to transpile as usual except ES6 modules: https://github.com/survivejs/webpack-book/blob/dev/manuscript/advanced_techniques/02_authoring_packages.md#what-files-to-publish-to-npm

kof:
Yeah, sounds right. So we are actually done here, right?

kof:
Ah no, module field…

sapegin:
I think `module` field would be out of scope of this PR, because it’d require extra Babel configuration.

kof:
ok

kof:
merged, thanks!

sapegin:
Cool, could you do the same for other JSS projects? Or we can setup mrm for that ;-)

kof:
Yes, exactly my plan, need to try mrm for that and lots of other things. Though maybe I will need a separate repository for that common stuff for "plugins", because they are a bit different than the core.

sapegin:
Most probably. Have you seen [webpack-defaults](https://github.com/webpack-contrib/webpack-defaults)?

kof:
nope, sounds like a good place to learn from, most probably would need to change a lot, but could use it as a starting point, right?

sapegin:
Yeah, sure. It’s not integrated into any project yet. Let me know if you need any help ;-)

Issue #413, opened 2017-02-01
Deterministic StyleSheet id


kof:
Now we got a use case for having a deterministic id for the sheet. Currently its needed by https://github.com/iamstarkov/jss-increase-id-specificity/ . cc @iamstarkov

kof:
I see 2 possibilities for the implementation:
1. Implement it as part of the core
2. Make it an interface and let others pass it over options or/and setter.

In both cases we gonna need a new hook which is called after all onProcessRuleHooks. It could be `onProcessSheet(sheet)`.

iamstarkov:
> 1. Implement it as part of the core

as far as so far only my plugin ever needed it, i dont see that much value in implementing id in a core

kof:
@iamstarkov check it out on https://github.com/cssinjs/jss/tree/process-sheet-hook

iamstarkov:
i decided to go with `sheet.toString()`. i guess its deterministic enough

kof:
Ok as we found a solution without id, it seems to be irrelevant for now. Hook `onProcessSheet` is already in master. Closing this for now.

kof:
onProcessSheet hook is released

PR #416, opened 2017-02-03
onProcessSheet hook #413

kof:


docs/plugins.md
```diff
@@ -60,3 +60,15 @@ You need to register a `plugin` only once per JSS instance. There is a number of
...
   ```
 
+1. Hook `onProcessSheet(sheet)`.
+
+  This hook is invoked on every created `StyleSheet` after all rules are processed, with the `sheet` as an argument.
+
+  ```javascript
+  jss.use({
+    onProcessSheet: (sheet) => {
+      // Do something here.
```
iamstarkov: can you add more explicit example? e.g. shall i mutate sheet argument or rather modify and return a copy?

iamstarkov:
let me test this branch tomorrow in the morning

docs/plugins.md
kof: Well in theory I am fully for immutable style, though in case of high performance cases like rule plugins, we do modify objects for the sake of performance, at some point we may think how to make it immutable but still not loose any perf.

kof: Also sheet is not a plain object, its an instance of StyleSheet, so you can't handle it in immutable fashion easily.

kof: Right now I assume user mutate only things that are publicly allowed/documented. StyleSheet has a number of methods.

kof: For setting an id, I need to define an official interface in the documentation.

iamstarkov:
@kof lets merge it, so we can go further

kof:
merged

iamstarkov:
awsm

PR #417, opened 2017-02-06
Renamed method call from '.hexString()' to the new '.hex()'.

janhartmann:
Just a small modification on the docs. As of https://github.com/Qix-/color/pull/96 the method for converting to a hex string is now just `.hex()`.


kof:
thanks, merged!

Issue #356, opened 2016-11-12
Dynamic Sheets for theming and animations


kof:
The idea is to have value functions and rule functions which accept an object and let JSS render them dynamically. This will allow us __efficient__  dynamic theming. The idea has been around for a while, but this time me and @nathanmarks have got a plan.

## What signature will value function and rule function have?

```javascript
{
  button1: {
    padding: 15,
    // Value function.
    color: props => props.buttonColor
  },
  // Rule function.
  button2: props => ({
    padding: props.padding,
    color: props.color
  })
}
```

## How do we know if a rule or prop needs a rerender?

We don't, we just call all dynamic values and compare new results with the current once.
We need to have a proof that calling all dynamic values on each update is an __actual__ perf problem. Right now this is a premature optimization, but if we need it, here are some options:

1. User defines properties which  are used within the value function / rule function
1. Optionally babel transform does a static analysis and provides those properties.

All suggestions are __NOT__ exclusive.

### Suggestion 1

Caching is ensured by the user. It is an opt-in.

```javascript
{
  button: {
    color: (nextProps, prevProps, prevValue) => {
      // User can prevent rerender by comparing props manually.
      // If user returns previous value, no rerendering happens.
      if (nextProps.primary === prevProps.primary) return prevValue

      return calculateColor(nextProps)
    }
  }
}
```

### Suggestion 2

It is basically suggestion 1 + a helper function which allows nicer caching notation.

```javascript

function cache(valueFn, ...propNames) {
  return (nextProps, prevProps, prevValue) => {
    if (hasChanged(propNames, nextProps, prevProps)) {
      return valueFn(nextProps)
    }
    return prevProps
  } 
}

{
  button: {
    color: cache(props  =>  calculateColor(props), 'primary')
  }
}
```
### Suggestion 3

Its  a caching integrated into JSS itself.

```javascript
{
  button: {
    // Once property "primary" has changed, JSS will call user's value function and get the new value.
    color: [props  =>  calculateColor(props), 'primary']
  }
}
```

## How do we make sure to render minimum CSS?

1. If a rule contains lots of dynamic values, it makes sense to make the entire rule dynamic - user should use function for rule declaration. In this case the entire rule will be rerendered if any property needs to rerender.
1. If a rule contains just just one or few dynamic props - value function should be used. In this case we should create a separate rule which contains just one property. This will save some bandwidth for SSR, because we are not copying all static properties of the rule together with the dynamic one.

## Fallbacks

[...]

PR #420, opened 2017-02-13
implement sheet.update() #356

kof:


docs/js-api.md
```diff
@@ -185,6 +185,18 @@ sheet.addRules({
 })
 ```
 
+## Update function values
+
+`sheet.update(data)`
+
+If you use [function values](), you will want to update them with new data. This method will call all your function values, pass the `data` param and update the CSS Rule if needed.
```
iamstarkov: what the link to _function values_?

kof: yep, will add them

kof:
merged

PR #433, opened 2017-02-27
New id generation algorithm

kof:


tests/integration/sheet.js
```diff
@@ -287,12 +275,12 @@ describe('Integration: sheet', () => {
...
       expect(sheet.toString()).to.be(
         '@media print {\n' +
-        '  .a-3787690649 {\n' +
+        '  .my-special-id {\n' +
         '    float: left;\n' +
         '  }\n' +
         '}\n' +
         '@media screen {\n' +
-        '  .a-3787690649 {\n' +
+        '  .my-special-id {\n' +
```
iamstarkov: i believe this usecase should end up with different ids in real world

kof: Why, this case should have ensured actually the opposite. But it was late and I made a mistake with a fixed id. Now we don't verify that the id is really same.

iamstarkov: why do you want this ids to be the same?

iamstarkov: ok, you are right

iamstarkov: thanks for explanation

tests/integration/sheet.js
```diff
@@ -255,14 +259,14 @@ describe('Integration: sheet', () => {
...
-          `  }\n` +
-          `}\n` +
+          '    float: left;\n' +
+          '  }\n' +
+          '}\n' +
           `.${id} {\n` +
-          `  color: red;\n` +
-          `}`
+          '  color: red;\n' +
+          '}'
```
iamstarkov: you may be interested in [`stripindent`](https://www.npmjs.com/package/common-tags#stripindent) literal tag from ['common-tags`](https://www.npmjs.com/package/common-tags) project

kof: yep, already started using it

kof: need a migration script to rewrite them all

iamstarkov: i dont know how did i manage to comment on outdated code. github just showed me latest changes

src/utils/generateClassName.js
```diff
@@ -1,16 +1,17 @@
...
-export default function generateClassName(str: string, rule: Rule, sheet?: StyleSheet): string {
-  if (sheet && sheet.options.meta) str += sheet.options.meta
-
-  const hash = createHash(str)
-
-  // There is no name if `jss.createRule(style)` was used.
-  return rule.name ? `${rule.name}-${hash}` : hash
-}
+export default (str: string, rule: Rule): string => (
+  // There is no rule name if `jss.createRule(style)` was used.
```
kof: actually need to change this interface as we will have no str param any more

kof:
@cvle @iamstarkov can you guys try this branch without calling rehydrate? I have tried just now, seems to be perfectly fine.

cvle:
@kof I'm not running SSR anywhere atm except on an old repository of mine, let's see if I can revive it.

kof:
In order to move forward with this, I will remove rehydration and release a minor version to receive some feedbacks.

kof:
btw. @oliviertassinari maybe material-ui has some server-side tests we can run to make sure this doesn't break anything?

oliviertassinari:
@kof I wish we add some server-side tests but as far as I know, we don't. That's definitely something we should add!

Issue #444, opened 2017-03-16
Remove JSON.stringify/parse, add onProcessStyle hook and more.


kof:
This will strongly improve performance and fix #441.
Instead of JSON.stringify/parse of the style object in `Rule` we will shallow clone rule.style. According to my tests it will be still much faster than it currently is with stringify/parse.

## Todo
- [x] Add new hook `onProcessStyle(style, rule, sheet)` which will behave like an [].map
    - only for rules which have a style object
    - replaces `rule.style` object with the result from `onProcessStyle()`
- [x] Introduce `jss.cloneStyle(style)` utility fn for high perf cloning
    - Deep clone
    - Iterate over object without hasOwnProperty check, because we expect plain objects
    ```
    // Prototype
    function cloneStyle(style) {
      const newStyle = Object.create(null)
      for (const name in style) {
        newStyle[name] = style[name]
      }
      return newStyle
    }
    ```    
- [x] use cloneStyle in the Rule
- [x] Remove JSON.parse/stringify, check the bench
- [x] Breaking change for `generateClassName(styleStr)` - we have no string any more.
- [x] Remove rule.originalStyle - we can get the original style from the sheet, doesn't need to be RegularRule property.
- [x] Remove default hook mapping
- [x] update all plugins which only change/generate style object to use the new hook (like jss-camel-case or jss-extend)
- [x] update all plugins use new api for originalStyle
- [x] update all plugins with the new generateClassName signature

kof:
cc @nathanmarks @cvle @iamstarkov

kof:
I am about to copy/paste the deepFreezeAndThrowOnMutationInDev function from the react-native package. Does any one know a better alternative? I think adding react-native to dependencies is a huge dependency overhead just for this case.

kof:
cc @cssinjs/core

kof:
Alternatively I could use cloning and diffing in order to warn about mutation, because `deepFreezeAndThrowOnMutationInDev` has those caveats: 

```
 * There are two caveats with this function:
 *   - If the call site is not in strict mode, it will only throw when
 *     mutating existing fields, adding a new one
 *     will unfortunately fail silently :(
 *   - If the object is already frozen or sealed, it will not continue the
 *     deep traversal and will leave leaf nodes unfrozen.
```

kof:
[...]

PR #449, opened 2017-03-22
Remove JSON.stringify/parse, addd onProcessStyle hook and more.

kof:
Full description #444


tests/index.js
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable import/first */
-
 import './utils'
+/*
```
kof: migrating plugins is wip

kof:
ok @cssinjs/core this is ready to be reviewed!

src/types.js
```diff
@@ -2,12 +2,17 @@
...
 export type ToCssOptions = {
   indent?: number
 }
 
-export type generateClassName = (str: string, rule: Rule) => string
+export type generateClassName = (rule: Rule) => string
+
+// TODO
+// Find a way to declare all types: Object|string|Array<Object>
+export type JssStyle = Object
```
iamstarkov: seems like its legit type def https://goo.gl/9RnYVu
screen shot 2017-04-05 at 10 57 07

kof: I narrowed down the problem in [this tweet](https://twitter.com/oleg008/status/844634629718102017)

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoALgTwA4FMwDKWM+AvGAPIBGAVrgMboY74AiDMY5RmJAPgEEATkICGmADw8SAPlRgwqKAFcAdowCWcVWAAWo1QBMS03AAoAzsVwAuQtYCUYAN6oAvqiVrN2vQeO4wmKYZoYcduz0ME6uChpQYGZB4gB0GhbJIWFRDk5CuOjKQjrOHh4KXuroWjr6RiSh4WCR0S7yYPTaVmD56Fx+9YEi4o057fGJvXkFRTq97b2zAwGmo9HuQA

kof: Have to try phantom types.

iamstarkov:
```diff
/* @flow */
type Style = Object
type Decl = Style | Array<Style>
  
-function handleStyle(style: Style) {
+function handleStyle(style: Decl) {
}

function handleArray(decl: Decl) {
  if (Array.isArray(decl)) return {}
}
  
function handle(decl: Decl) {
  const ret = handleArray(decl)
  if (ret) return ret
  return handleStyle(decl)
}
```

kof: Sure, the pobl. is that handleStyle accepts only Style in that example, to be specific to our code base: RegularRule(name, style) where style is always an object. We don't expect arrays there. On the other hand FontFaceRule(selector, style) accepts style as object OR as array in order to support multiple [font faces at once](http://cssinjs.org/json-api?v=v6.5.0#font-face)

So the check `if (Array.isArray)` is local in FontFaceRule and RegularRule, but flow only can detect this if this check would be located in createRule() function which creates all the rules.

iamstarkov: ok

PR #455, opened 2017-04-04
Delete wrong dynamic styles

b2whats:
updated dependency 'karma-webpack' in connection with the error - https://github.com/karma-runner/karma/issues/2452

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "karma-mocha": "1.3.0",
     "karma-mocha-reporter": "2.2.1",
     "karma-sourcemap-loader": "0.3.7",
-    "karma-webpack": "1.8.0",
+    "karma-webpack": "^2.0.3",
     "lint-staged": "3.2.2",
     "mocha": "3.2.0",
     "pre-commit": "1.1.3",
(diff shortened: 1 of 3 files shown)


kof:
will check it out soon, currently in the middle of a big refactoring.

tests/unit/utils.js
```diff
@@ -6,6 +6,11 @@ describe('Unit: jss', () => {
     it('should extract dynamic styles', () => {
       const color = data => data.color
       const styles = {
+        span: {
+          '& > *:last-child': {
```
kof: its just any nested object, right?

src/utils/getDynamicStyles.js
```diff
@@ -6,25 +6,23 @@ export default (styles: Object): Object|null => {
...
       }
       else if (type === 'object' && value !== null && !Array.isArray(value)) {
-        if (!to) to = {}
         const extracted = extract(value)
         if (extracted) to[key] = extracted
       }
     }
 
-    return to
+    return Object.keys(to).length > 0 && to
```
kof: We can avoid that function call, perf is critical here.

package.json
```diff
@@ -75,7 +75,7 @@
     "karma-mocha": "1.3.0",
     "karma-mocha-reporter": "2.2.1",
     "karma-sourcemap-loader": "0.3.7",
-    "karma-webpack": "1.8.0",
+    "karma-webpack": "^2.0.3",
```
kof: unrelated change

kof:
Fixed it, even reduced complexity b60654e6e57cf8473673c5939d738613be182886

PR #463, opened 2017-04-11
Add XING to users.md

rudeayelo:


kof:
Nice, will be visible on the site with the next release.

PR #477, opened 2017-04-25
Fix addRule for functional values, fix #475

### LINKED ISSUE #475: Need to detach sheet for addRule with function values
[Online Example](https://www.webpackbin.com/bins/-KiTXf9BfyDXWpbJ0i_J)

If we add new rule with just function values, `.update()` does not work for this new rule:

```js
const styles = {
  title: {
    fontSize: 20,
    color: ({color}) => color
  }
}

const sheet = jss.createStyleSheet(styles, {link: true}).attach()
sheet.addRule('subtitle', {color: ({color}) => color})
sheet.update({color: 'red'}) // still no color for subtitle, with attach() then too
// but if we call .deploy() - color applies
```

But this works:
```js
const sheet2 = jss.createStyleSheet(styles, {link: true}).attach()
sheet2.detach()
sheet2.addRule('subtitle', {color: ({color}) => color})
sheet2.update({color: 'red'}).attach() // red color for subtitle too
```

lttb:
The problem is that `rule.toString()` returns nothing if we have just function values in rule without calculated values. So it needs to return just an empty rule like `.class {}` to be applied to the stylesheet.

### DIFF
--- a/src/utils/toCss.js
+++ b/src/utils/toCss.js
@@ -47,17 +47,20 @@ export default function toCss(selector: string, style: JssStyle, options: Option
     }
   }
 
+  let hasFunctionValue = false
+
   for (const prop in style) {
     let value = style[prop]
     if (typeof value === 'function') {
       value = style[`$${prop}`]
+      hasFunctionValue = true
     }
     if (value != null && prop !== 'fallbacks') {
       result += `\n${indentStr(`${prop}: ${toCssValue(value)};`, indent)}`
     }
   }
 
-  if (!result) return result
+  if (!result && !hasFunctionValue) return result
 
   indent--
   result = indentStr(`${selector} {${result}\n`, indent) + indentStr('}', indent)
--- a/tests/functional/sheet.js
+++ b/tests/functional/sheet.js
@@ -247,6 +247,34 @@ describe('Functional: sheet', () => {
     })
   })
 
+  describe('.addRule() with just function values and attached sheet', () => {
+    let style
+    let sheet
+
+    beforeEach(() => {
+      sheet = jss.createStyleSheet().attach().link()
+      sheet.addRule('a', {color: ({color}) => color})
+      style = getStyle()
+    })
+
+    afterEach(() => {
+      sheet.detach()
+    })
+
+    it('should render an empty rule', () => {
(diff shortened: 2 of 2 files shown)


kof:
Why do you call toString() before .update()?

lttb:
We call .toString in injectRule, that called in addRulle, because we make .addRule before .update

kof:
Maybe we should change the behavior there?

kof:
I am not sure what will happen if we render a rule without props in different browsers.

lttb:
But we can add rule without data (for lazy apply for example), and it's a normal behavior for user

kof:
Yeah, need to run tests in all browsers.

lttb:
if we will have some issues with browsers, we always can render an empty prop

kof:
what is an "empty" prop?

lttb:
> what is an "empty" prop?

just an empty string, for example `.class {color:''}`
or we can also make it `initial`

kof:
color: '' is an invalid value, I don't know what this might cause.

kof:
"initial" is an actual value which might lead to unexpected results from the user perspective

iamstarkov:
`color`'s `initial` is different in each browser

tests/functional/sheet.js
```diff
@@ -247,6 +247,36 @@ describe('Functional: sheet', () => {
     })
   })
 
+  /*
```
kof: why is this commented out?

src/utils/toCss.js
```diff
@@ -47,17 +47,20 @@ export default function toCss(selector: string, style: JssStyle, options: Option
...
       value = style[`$${prop}`]
+      hasFunctionValue = true
     }
     if (value != null && prop !== 'fallbacks') {
       result += `\n${indentStr(`${prop}: ${toCssValue(value)};`, indent)}`
     }
   }
 
-  if (!result) return result
+  if (!(result || hasFunctionValue)) return result
```
kof: much more readable:

```js
if (!result && !hasFunctionValue) return result
```

tests/functional/sheet.js
lttb: sorry, just forgot to uncomment

kof:
some browsers time out on browserstack, trying to figure out why.

PR #494, opened 2017-05-17
Fixes for the SSR classname counter reset bug.

jcheroske:
There is a bug in `jss` where one of the counters used to generate classnames does not reset to `0` with each SSR request. For the bug to be completely fixed, a commit to `react-jss` is also needed. The following  changes were made:
* `generateClassName` was modified to have a factory function, `createGenerateClassName`, exported as a named export , that creates an independent instance of the classname generator.
* The `default export` from `generateClassName` now simply calls the factory function.
* The SheetsRegistry `constructor` now takes an optional `options` object. There are currently two new options:
  * `resetClassNameCounter: boolean`: if set to true, each new instance of `SheetsRegistry` will have its own instance of the built-in `generateClassName` function.
  * `generateClassName: generateClassName`: if supplied, sets the option to the supplied generation algorithm.

The end result is that, if you do:
```
const registry = new SheetsRegistry()
```
then there should be no changes to existing behavior. But if you do:
```
const registry = new SheetsRegistry({resetClassNameCounter: true})
```
Then the registry will have an independent copy of the generator that begins at 0.

### DIFF
--- a/src/Jss.js
+++ b/src/Jss.js
@@ -45,9 +45,19 @@ export default class Jss {
    * Create a Style Sheet.
    */
   createStyleSheet(styles: Object, options: StyleSheetOptions): StyleSheet {
+    let generateClassName
+    if (options
+      && options.sheetsRegistryContext
+      && this.options.generateClassName === generateClassNameDefault) {
+      generateClassName = options.sheetsRegistryContext.generateClassName
+    }
+    else {
+      generateClassName = this.options.generateClassName
+    }
+
     const sheet = new StyleSheet(styles, {
       jss: (this: Jss),
-      generateClassName: this.options.generateClassName,
+      generateClassName,
       insertionPoint: this.options.insertionPoint,
       ...options
     })
--- a/src/SheetsRegistry.js
+++ b/src/SheetsRegistry.js
@@ -1,5 +1,6 @@
 /* @flow */
-import type {ToCssOptions} from './types'
+import {createGenerateClassName} from './utils/generateClassName'
+import type {ToCssOptions, SheetsRegistryContext} from './types'
 import type StyleSheet from './StyleSheet'
 
 /**
@@ -7,6 +8,13 @@ import type StyleSheet from './StyleSheet'
  */
 export default class SheetsRegistry {
   registry: Array<StyleSheet> = []
+  context: SheetsRegistryContext
+
+  constructor() {
+    this.context = {
+      generateClassName: createGenerateClassName()
+    }
+  }
 
   /**
    * Register a Style Sheet.
@@ -52,5 +60,13 @@ export default class SheetsRegistry {
       .filter(sheet => sheet.attached)
       .map(sheet => sheet.toString(options))
       .join('\n')
+  }
+
+  /**
+   * Get the class name generator, if there
+   * is one.
+   */
+  getContext(): SheetsRegistryContext {
+    return this.context
   }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ lib
 dist
 .env
 coverage
+.idea
--- a/src/types.js
+++ b/src/types.js
@@ -56,6 +56,10 @@ export type Plugin = {
   onChangeValue?: (value: string, prop: string, rule: Rule) => string
 }
 
+export type SheetsRegistryContext = {
+  generateClassName: generateClassName
+}
+
 export type JssOptions = {
   generateClassName?: generateClassName,
   plugins?: Array<Plugin>,
@@ -86,6 +90,7 @@ export type StyleSheetOptions = {
   virtual?: boolean,
   Renderer?: Function,
   generateClassName?: generateClassName,
+  sheetsRegistryContext?: SheetsRegistryContext,
   jss: Jss
 }
 
--- a/src/utils/generateClassName.js
+++ b/src/utils/generateClassName.js
@@ -1,17 +1,21 @@
 /* @flow */
-import type {Rule} from '../types'
+import type {Rule, generateClassName} from '../types'
 
 const globalRef = typeof window === 'undefined' ? global : window
 const namespace = '__JSS_VERSION_COUNTER__'
 if (globalRef[namespace] == null) globalRef[namespace] = 0
 // In case we have more than one JSS version.
 const jssCounter = globalRef[namespace]++
-let ruleCounter = 0
 
-/**
- * Generates unique class names.
- */
-export default (rule: Rule): string => (
-  // There is no rule name if `jss.createRule(style)` was used.
-  `${rule.name || 'jss'}-${jssCounter}-${ruleCounter++}`
-)
+export const createGenerateClassName = () : generateClassName => {
+  let ruleCounter = 0
+  /**
+   * Generates unique class names.
+   */
+  return (rule: Rule): string => (
+    // There is no rule name if `jss.createRule(style)` was used.
+    `${rule.name || 'jss'}-${jssCounter}-${ruleCounter++}`
+  )
+}
+
+export default createGenerateClassName()
--- a/tests/integration/sheetsRegistry.js
+++ b/tests/integration/sheetsRegistry.js
@@ -109,4 +109,21 @@ describe('Integration: sheetsRegistry', () => {
       )
     })
   })
+
+  describe('.getClassNameGenerator', () => {
+    it('should return a generator', () => {
+      const registry = new SheetsRegistry()
+      expect(registry.getContext().generateClassName).to.not.be(undefined)
+    })
+
+    it('should return different generators from two instances', () => {
+      const registry1 = new SheetsRegistry()
+      const registry2 = new SheetsRegistry()
+      const generator1 = registry1.getContext().generateClassName
+      const generator2 = registry2.getContext().generateClassName
+      expect(typeof generator1).to.be('function')
+      expect(typeof generator2).to.be('function')
+      expect(generator1).to.not.be(generator2)
+    })
+  })
 })


src/SheetsRegistry.js
```diff
@@ -1,12 +1,22 @@
...
 /**
  * Sheets registry to access them all at one place.
  */
 export default class SheetsRegistry {
   registry: Array<StyleSheet> = []
+  registryOptions: SheetsRegistryOptions
+
+  constructor(options: SheetsRegistryOptions = {}) {
+    const generateClassName = options.generateClassName ||
+      (options.resetClassNameCounter === true && createGenerateClassName())
```
kof: why do you need an option? Why not to reset the counter every time a new registry is created?

iamstarkov:
@jcheroske ping

src/SheetsRegistry.js
jcheroske: Sorry about the slow reply. Somehow this notification slipped through the cracks. 

Anyways, that was my original plan, and I'm happy to go back to it if that's what folks want. My thinking had to do with backwards compatibility. Since the SheetsRegistry now has a classname generator in it, it's going to get passed into `createStyleSheet` by `react-jss`. When that happens, it's going to override any classname generator that was previously configured at the `jss` level, which would break existing code. I guess the question is: how does the registry's generator relate to generators configured in other places? The way I implemented it, it has the highest priority, but you have to turn it on. That way, folks that have configured a hashing generator in their jss instance don't see their code strangely break.

The more I think about it, I'm not satisfied with the approach I took. I think that perhaps the registry generator should get passed into `createStyleSheet` using a different `options` key, and not under `generateClassName`. Something like `options.registryGenerateClassName`. Maybe even `options.sheetRegistryContext`. Then `jss` will only use it if a generator hasn't already been configured somewhere else.

jcheroske: I'm going to submit another commit with a different approach.

jcheroske:
I just reworked it to not need the options param on `SheetsRegistry`. Now, the registry simply creates a new generator and makes it available via a `getContext()` method. The caller to `Jss.createStyleSheet` then passes the context in as the `sheetsRegistryContext` option. If the `Jss` instance is configured with a different generator algorithm, then that takes precedence and the generator from the registry is ignored. But if the `Jss` instance is using the default generator, then the registry's generator is used for that call.

jcheroske:
I'm wondering if someone could give me some feedback on this?

kof:
Looking into it.

jcheroske:
And sorry about the `.gitignore` change. Just learned that's bad form. Got my global ignore file dialed in now.

kof:
I am not quite happy with the implementation, need to put more thoughts into it.

jcheroske:
I'm not offended. Just curious what rubs you the wrong way?

src/Jss.js
```diff
@@ -45,9 +45,19 @@ export default class Jss {
    * Create a Style Sheet.
    */
   createStyleSheet(styles: Object, options: StyleSheetOptions): StyleSheet {
+    let generateClassName
+    if (options
+      && options.sheetsRegistryContext
+      && this.options.generateClassName === generateClassNameDefault) {
```
kof: not sure if this logic will work, for e.g. when user defined already a different generateClassName

jcheroske: The logic in that block, coupled with the `else` block and the `...options` spread below, is as follows:

Did a `sheetRegistryContext` get passed in, and is the Jss-instance top-level `generateClassName` still set to the original default?
* Yes? Then use the `generateClassName` from the registry context.
* No? Then use the `generateClassName` from the Jss instance and disregard the registry context.

Then, override all of that (with the options spread) if a `generateClassName` option got passed into `createStyleSheet`

kof: Thats the problem, what if user changed the generateClassName algorithm and now he uses the ThemeProvider with registry?

kof:
The idea I was tinkering with is - how can we provide a more straight forward way to reset the counters.

jcheroske:
I guess I'm not seeing the problem correctly. The way I wrote it, if the user changes the algorithm then they are on their own and will have to manage the reset of their algorithm themselves. Since the concept of a counter is not shared among all generators, I only dealt with resetting the counter on the built-in generator. I'll need to see some pseudo-code or something if I'm to understand your concerns better. If you just want to tackle it yourself then that's fine too. It's a low-severity bug since we're not seeing it in production, only in dev.

kof:
> they are on their own

Still they may need to react on rerender, thats why I am thinking about a more explicit way.

jcheroske:
Ah, ok, I get it now. Hmmm, I'll ponder that in a background thread...

kof:
Right now SheetsRegistry instance on SSR is not directly connected with the Jss instance. I think we need to change that in order to get an interface for a `reset`.

kof:
Or provide a reset method from jss:  `jss.reset()`, which calls `classNameGenerator.reset()` and instead of `generateClassName` function we would have:

```js
// ClassNameGenerator.js
class ClassNameGenerator() {
  get() {}
  reset() {}
}
```

jcheroske:
You're reading my mind. If changing the generator interface is something
you're open to, then there are possibilities. Making the generator an
object would allow sending a message to it. Another approach would be to
move from a generator function to a generator function factory, and then
jss could just decide when it wanted a fresh generator function.

kof:
> and then jss could just decide when it wanted a fresh generator function.

how would it know when it needs a fresh generator fn?

kof:
Oh yes, so if we still add a jss.reset() method, it can create a new generator function.

jcheroske:
I will submit something in a couple of hours for you to look at.

kof:
Potential problem when using class names generator function - generateClassName references all over the place will point to the old one. On the server its not a problem as the sheets won't be used any more, but on the client it might be if someone for some reason decides to use it.

jcheroske:
Sorry, I haven't been able to get back on this. I did think about it some more though. The issue is really about scope. On the server, we need a request scope, where everything is setup fresh and then discarded. When `createStyleSheet` is called, right now it has no access to a request scope. If you put a `jss.reset()` method in, then that would reset all requests currently happening, which could be bad. To solve this, we can simply do two things:

1. Pass in the `sheetsRegistry` to createStyleSheet. When that happens, `jss` can then use the registry as a request scope. It can look there for a generator. If it finds one there then it can just use it. But if it doesn't find a generator there (when the registry is passed in for the first time) then it will need to make a new generator and put it in the registry. Which means that:

2. Jss should be configured with either a generator or a generator factory. If it's configured with only a generator, then Jss just uses that single generator and hopes for the best. What more can we do? This isn't bad. Hashing generators are probably fine used like this, and never need to be reset. But if the generator does maintain state that would need to be reset, then a factory should be passed in instead. The built-in generator should then be implemented as a factory.

On the client, Jss just makes a generator from the factory and uses it and all is well. On the server, jss looks inside the sheet registry and checks to see if there's a generator there. If not, it just makes a new one from the factory and puts it in there. This is isomorphic because, even if the client is using the sheets registry, the registry will be long-lived and the generator in there essentially becomes the generator for the life of the app. 

I think, with careful argument checking/management we can add this without disrupting old code. We fix the reset bug and enhance the api with two new, but easily comprehensible parts: Configuration with a generator factory and passing in the registry to `createStyleSheet`.

The only new type would be the generator factory.

kof:
@jcheroske I am on it now.

jcheroske:
Ok. Good luck. Curious what you come up with.

kof:
> The issue is really about scope. 

Yep

>  On the server, we need a request scope, where everything is setup fresh and then discarded.

Ideally not everything, for e.g. jss instance according to the current plan should remain the same. There are 2 things that need to be reseted, sheets registry and now also counter for classes generator.

> If you put a jss.reset() method in, then that would reset all requests currently happening, 

True, requests are never parallel, but once they come in, something  will happen async, so we never know when a req will be ended.

> Pass in the sheetsRegistry to createStyleSheet. When that happens, jss can then use the registry as a request scope. It can look there for a generator.

I am still not quite happy with this abstraction, because essentially SheetsRegistry should have nothing to do with  classes generator, they are completely unrelated entities.

I think we need some new entity like "context" which can contain everything that is related to the current lifecycle. Not sure whats the best name for it so that it makes sense equally for both server and client. Possible names: context, scope, lifecycle, anything else?

>  But if it doesn't find a generator there (when the registry is passed in for the first time) then it will need to make a new generator and put it in the registry.

I think JSS should always find a ready/reseted generator function in that context object.

> Jss should be configured with either a generator or a generator factory. 

I think jss style sheet will only use generator, but there will be also generator factory which will be used when created a context.

kof:
Regarding naming, I tend to call it "lifecycle", wdyt?

jcheroske:
"lifecycle" as a name for the request scope? Hmmm... "RequestLifecycle"?
"RequestContext"?

kof:
It will be used on both client and server, so it needs a universal name.

jcheroske:
Can you describe the lifecycle more fully? On the server it's pretty clear.
What is the lifecycle on the client?

I understand you don't like putting the generator in the registry, but,
it's really just an implementation detail. It wouldn't be exposed in the
public API. Keeping the surface area of the API succinct would be nice. I'm
already creating a registry. Why not just pass it to createStyleSheet? And
react-jss can do that automatically, so nothing new to do there. Under the
hood, you can use it as request scope. No one needs to know.

To me, the registry is a server-side thing anyway. I don't really see why
the client needs to have a new concept introduced to fix what is a
server-side bug.

But that said, I'm not going to disparage your judgement! You are probably
seeing this issue much more deeply then I am.

kof:
I want basically a container for such things to avoid having too much logic for different environments. Lifecycle object on the client would live for the entire client lifetime. There will be probably no need for a reset, but the model can still be the same.

kof:
Will write some code now.

kof:
In terms of react-jss, I think we need to make a breaking change then, because instead of  SheetsRegistry provider it will be JssProvider or somemething that will contain the lifecycle object

kof:
Ok I ended up with this: 772073be60493ec8e5263266ce0af8944cbe0b7b react-jss commit will follow soon. Because this is a breaking change I will also add some other breaking changes for the next major release. So everything is happening on the next branch right now.

Issue #495, opened 2017-05-17
Problem with createStyleSheet again


bigslycat:
Error is equal this ussue #484. Reproduced on project with Flow `0.46.0` version.

PR #496, opened 2017-05-17
Fix createStyleSheet type definitions error

bigslycat:
#495 If `options` is optional, it must have the default value. In our case it is empty object. But the empty object is incompatible with `StyleSheetOptions` since `StyleSheetOptions` has a required parameter `jss`. But since the `StyleSheet` constructor still comes with an object with this parameter, we do not need to make it required in the `createStyleSheet` options. To solve the problem, I created `StyleSheetFactoryOptions` type.

### DIFF
--- a/src/Jss.js
+++ b/src/Jss.js
@@ -9,7 +9,7 @@ import findRenderer from './utils/findRenderer'
 import type {
   Rule,
   RuleOptions,
-  StyleSheetOptions,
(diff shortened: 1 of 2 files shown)


kof:
thanks, lets continue in #496

kof:
sorry wrong, wrong one

kof:
merged, thanks!

Issue #498, opened 2017-05-18
Flow - jss.createStyleSheet(styles) throws an error


Ricardo-Marques:
`Undefined (too few arguments, expected default/rest parameters). This type is incompatible with StyleSheetOptions`

Editing https://github.com/cssinjs/jss/blob/master/src/Jss.js#L47 to
`createStyleSheet(styles: Object, options?: StyleSheetOptions): StyleSheet {`
fixes the issue.

bigslycat:
Duplicates #495

PR #499, opened 2017-05-18
fixes flow error on createStyleSheet when options is not passed

Ricardo-Marques:
#498


Ricardo-Marques:
Just realized there's another PR already submitted to fix this, feel free to close this.

bigslycat:
Sorry, but this is not enough. If you pass the function object `options` without the property `jss`, then again get an error. Therefore, in my PR, I was forced to create a new type, in which this property is not required.

PR #502, opened 2017-05-22
Support insertion points within body tag

kohlmannj:
I need more insertion point control than just the current "anywhere in the `head`" functionality of JSS offered. Hence, this PR:

- Adds support for insertion point comments (i.e. `<!-- jss -->`) as immediate descendants in the `body` of the document
- Adds a test to verify stylesheet insertion order when using this capability
- Minorly expands the "Specify DOM insertion point" documentation to describe the new capability

There are no test regressions, and the one added test currently passes.

### DIFF
--- a/src/backends/DomRenderer.js
+++ b/src/backends/DomRenderer.js
@@ -67,6 +67,17 @@ const getHead = (() => {
 })()
 
 /**
+ * Gets the `body` element upon the first call and caches it.
+ */
+const getBody = (() => {
+  let body
+  return (): HTMLElement => {
+    if (!body) body = document.body || document.getElementsByTagName('body')[0]
+    return body
+  }
+})()
+
+/**
  * Find attached sheet with an index higher than the passed one.
  */
 function findHigherSheet(registry: Array<StyleSheet>, options: PriorityOptions): StyleSheet|null {
@@ -97,39 +108,58 @@ function findHighestSheet(registry: Array<StyleSheet>, options: PriorityOptions)
 }
 
 /**
- * Find a comment with "jss" inside.
+ * Look within the document head and body for a comment with "jss" inside.
  */
 function findCommentNode(text: string): Comment|null {
-  const head = getHead()
-  for (let i = 0; i < head.childNodes.length; i++) {
-    const node = head.childNodes[i]
-    if (node.nodeType === 8 && node.nodeValue.trim() === text) {
-      return node
+  let commentNode = null
+
+  const parentNodes: Array<HTMLElement> = [getHead(), getBody()]
+
+  parentNodes.forEach((parentNode: HTMLElement) => {
+    // Return early if the node is null or a non-object
+    if (typeof parentNode !== 'object' || parentNode === null) {
+      return undefined
     }
-  }
-  return null
+
+    for (let i = 0; i < parentNode.childNodes.length; i++) {
+      const node = parentNode.childNodes[i]
+      if (node.nodeType === 8 && node.nodeValue.trim() === text) {
+        commentNode = node
+        return undefined
+      }
+    }
+
+    return undefined
+  })
+
+  return commentNode
 }
 
 /**
  * Find a node before which we can insert the sheet.
  */
-function findPrevNode(options: PriorityOptions): ?Node|null {
+function findPrevNodeAndParent(options: PriorityOptions): [?Node|null, ?Node|null] {
   const {registry} = sheets
 
   if (registry.length > 0) {
     // Try to insert before the next higher sheet.
     let sheet = findHigherSheet(registry, options)
-    if (sheet) return sheet.renderer.element
+    if (sheet) return [sheet.renderer.element, sheet.renderer.element.parentNode]
 
     // Otherwise insert after the last attached.
     sheet = findHighestSheet(registry, options)
-    if (sheet) return sheet.renderer.element.nextElementSibling
+    if (sheet) {
+      return [
+        sheet.renderer.element.nextElementSibling,
+        sheet.renderer.element.parentNode
+      ]
+    }
   }
 
   // Try to find a comment placeholder if registry is empty.
   const comment = findCommentNode(options.insertionPoint)
-  if (comment) return comment.nextSibling
-  return null
+  if (comment) return [comment.nextSibling, comment.parentNode]
+  return [null, null]
 }
 
 export default class DomRenderer {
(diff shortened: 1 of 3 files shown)


kohlmannj:
Tiny nudge on this: I have some work an an internal component library that is soft-blocked on this. (i.e. it'd work without this, but certain style orderings would be incorrect.) Please let me know if I can help clarify or otherwise assist in reviewing this PR.

src/backends/DomRenderer.js
```diff
@@ -67,6 +67,17 @@ const getHead = (() => {
 })()
 
 /**
+ * Gets the `body` element upon the first call and caches it.
+ */
+const getBody = (() => {
+  let body
+  return (): HTMLElement => {
+    if (!body) body = document.body || document.getElementsByTagName('body')[0]
```
kof: do you know which browsers still need a fallback to `getElementsByTagName`, if we target ie9+ maybe this can be omited already also for the head mb? It would mean we don't need to cache those.

kof:
What is an actual use case for this change?

kohlmannj:
@kof It's a really boring one: stick with me for a second:

Our internal tools focus on building a React app within an existing HTML shell as part of our development server. There are enough different resources being added in this shell that I need to carefully control style insertion order across various technologies, including aphrodite-jss and JSS proper.

As it turns out, appending a `<!-- jss -->` comment into the `<head>` of our HTML shell has a pretty awful complication: some other element in the `<head>` (which we have zero control over) gets shifted into the `<body>`. There aren't really great locations to insert the comment between the very end of the `<head>` and somewhere above the offending element because, again, we as a group do not control the contents of the `<head>`.

Aside: I also need need the `<head>` to handle ordering of vendor stylesheets inserted by style-loader (we insert vendor SCSS/CSS at the top of the head, and our internal component library's SCSS/CSS at the bottom of the head. style-loader only has "container" and "top/bottom" insertion order granularity.

Anyway, I _can_, however, dynamically insert a `<!-- jss -->` comment at the very start of the `<body>` with high reliability. Hence, this became the most viable strategy I could conceive of to ensure that our dev server style ordering is similar to the ordering within a static HTML page build.

src/backends/DomRenderer.js
kohlmannj: I'll check and get back to you on this later this evening!

kof:
1. I don't really understand how things may accidentally shift from head to the body
2. Did you know you can specify a different namespace for the insertion point? for e.g.  This is how I control different kinds of jss sheets in my application.

kohlmannj:
@kof Re (1): it's actually kind of depressing; our `<head>` contains something like  `<noscript></noscript>`, which Chrome and Safari move into the `<body>` after parsing. (It's a weak SO reference, but here's one: https://stackoverflow.com/a/38115205)

Re (2): Ooh! I am woefully unfamiliar with this; I would greatly appreciate a usage example, please and thank you.

kof:
The img "hoisting" s known, but why should sheets get "hoisted"?

kof:
I don't have a good example: https://github.com/cssinjs/jss/blob/master/docs/js-api.md#setup-jss-instance insertionPoint is an option to a jss instance, which is also passed to the sheets, the value of this option defaults to "jss".

kof:
Here is a test https://github.com/cssinjs/jss/blob/master/tests/functional/priority.js#L138

kohlmannj:
@kof Oh, okay: I've looked at that test and have definitely examined that part of the docs. I wasn't sure what you were referring so, I thought there might be some other capability, like [inserting into a different element or a ShadowRoot](https://github.com/webpack-contrib/style-loader#insertinto) that I hadn't seen.

I'm definitely using `insertionPoint`, but because of our shell, I need more control over where I can place those comment nodes, and where JSS looks for them.

----

Anyway, I have some demo pages, which are representative of our development server environment.

## [Incorrect Style Ordering (using JSS 7.1.2)](http://adv.preview.nytimes.com/prototypes/header-gallery/broken/)
This first demo page shows what happens when I append a `<!-- jss-simple -->` comment to the `<head>` (using Cheerio within a Node environment, after fetching a remote HTML file, but before sending that page to the browser).

(Please view source on these HTML files to verify the document order of the `<!-- jss-simple -->` comment before the HTML is parsed by a browser.)

incorrect style ordering

----

## [Correct Style Ordering (using kohlmannj:master)](http://adv.preview.nytimes.com/prototypes/header-gallery/fixed/)
This second demo is built using this PR. Ahead of time, the `<!-- jss-simple -->` comment is prepended to the `<body>`, and since JSS has been updated to support insertion points that are immediate children of `<body>`, JSS finds the comment and render the `<style>` tags in the expected order on the page.

correct style ordering

kof:
Looks like this hoisting is applied to all elements after the first element that needs to be hoisted. Is that right? If yes, is it possible to move those  "hoisted" elements to the bottom of the head?

kohlmannj:
Closing this PR. I was going to provide additional discussion of why this support is important to my team and the build tools I maintain, but our time is limited. Suffice to say, this particular implementation choice did not come quickly or lightly. Furthermore, I'm disappointed that mere demonstration of a use case for what is ultimately a minor feature enhancement has turned into such an arduous discussion.

kof:
Wow, thats harsh. I am trying to do my best to make sure we don't add things nobody needs.  Its my job to question everything until I see that this is absolutely needed. Also I have other ideas how to support this. Instead of spending time into implementation you should have talked to me first. There is no need to be disappointed just because I am trying to understand exactly why you are trying to do that.

kof:
So the reason I am hesitating to add this is it may lead to ambiguity once someone uses comment nodes in the head __and__ in the body.

kof:
An option I was thinking about is to let user specify a dom node as insertionPoint option instead of a string.

kohlmannj:
@kof Dear Oleg,

I've thought things over, especially my own reaction, and I want to sincerely apologize for failing to give you the benefit of assumed-positive intentions in my moment of frustration. _Of course_ you have raised some valid points, and certainly you should maintain the privilege of healthy skepticism as the project maintainer. Thank you for your patience; I hope you can forgive me.

– Joe

----

It's late here, but I have a few ideas on how to proceed if you're interested:

- Tomorrow I will try to concisely describe our second, more important use case for this feature (turns out it's more important than working around the `head` hoisting issue with our shell; that was the most recent thing I was resolving, so I forgot about the original use case)
- I agree with [your concern about ambiguity](https://github.com/cssinjs/jss/pull/502#issuecomment-304055617), especially since this the current insertion point strategy is a simple, unifying approach within a powerful and complex API
- An insertionPoint DOM node option would very likely satisfy our use case
- Hinting at our more important use case, I can alternately create an issue / PR in [style-loader](https://github.com/webpack-contrib/style-loader) for a more granular insertion strategy, e.g. an `insertBefore` option

kof:
>  Thank you for your patience; I hope you can forgive me.

No worries, we have all our frustration moments.

kof:
> more granular insertion strategy

I am not sure how this issue with jss is related to style-loader

kof:
Ok I am on it now.

kof:
It is in master now, please check it out and let me know if it works for you. Closing this one.

kof:
Just  realized I have added a bug, fixing…

kof:
Its now on the "next" branch

kof:
Ok I will have to "blindly" release the new version now without your approve. Please test JSS 8. It will be published in 5 min.

PR #506, opened 2017-05-26
Implement getSeparatedStyles

lttb:
I think this helper would be useful (in `styled-jss` for example)

cc @kof


lttb:
We decided to implement it in just `styled-jss`

PR #509, opened 2017-05-29
Fixed data param

wellguimaraes:


kof:
Thanks, weird nobody noticed, will add a test.

kof:
I was too fast, if you read exactly there is arguments overload.I wonder why you tried to fix it if nothing is broken.

wellguimaraes:
Hey @kof , thank you for the quick reply. Well, sorry for that, but I'm facing this issue:

- When there's some conditional rule (ex.: media query), the properties defined with functions are called sometimes with undefined param.

I'll investigate it better then.

kof:
please create a failing test or at least a proper example of whats failing

wellguimaraes:
Sure, I'll do.
;)

wellguimaraes:
I had this problem while using styled-jss and I've just found the issue in that project. Sorry for the mess @kof ! ;)

Issue #511, opened 2017-05-31
Dynamic props for media queries do not update with attach before


lttb:
[Online example](https://www.webpackbin.com/bins/-KlTISLUF0D1F82iVFqX)

Does not work:
```js
const styles = {
  container: {
    padding: props => props.active ? 10 : 20,
    '@media screen and (min-width: 100px)': {
      color: 'red',
      background: props => (props.active ? 'white' : 'gray')
    }
  }
}

const sheet = jss.createStyleSheet(styles, {link: true}).attach()
sheet.update({active: false})
```

Works:
```js
const styles = {
  container: {
    padding: props => props.active ? 10 : 20,
    '@media screen and (min-width: 100px)': {
      color: 'red',
      background: props => (props.active ? 'white' : 'gray')
    }
  }
}

const sheet = jss.createStyleSheet(styles, {link: true})
sheet.update({active: false}).attach()
```

Issue #346, opened 2016-11-05
Generated keyframe id 


kof:
Currently keyframe id's we use in jss are global. The questions are:
1. Do we want to generate a keyframe id to avoid conflicts?
2. Do we want to allow user to define the keyframe id optionally and generate it by default?
3. How would user access generated keyframe id?

Currently user can generate keyframe id like this, but I assume it is not very user friendly:

```js
const identifier = Math.random()
export default {
  [`@keyframes ${identifier}`]: {
    from: {opacity: 0},
    to {opacity: 1}
  }
}
```

## Todo

```javascript
const sheet = jss.createStyleSheet({
  '@keyframes fadeIn': {
     from: {opacity: 0},
     to {opacity: 1}
  },
  button: {
    animationName: '$fadeIn',
  }
})

sheet.keyframes.fadeIn // @keyframes fadeIn-12378676134
```

- [x] generate id
- [x] register it in keyframes hash
- [x] docs
- [x] changelog
- [x] allow unscoped keyframes via global plugin
- [x] allow `animation` property?
- [x] if we could run ref replacer after jss-expand, we could also support `animation: {name}` syntax and we wouldn't have to support `animationName` if we could run it after camel case plugin
- [x] document inside of @global rule and global prefix
- [x] rename generateClassName

giuseppeg:
> Do we want to generate a keyframe id to avoid conflicts?

I think so, users can make them global if they want #324

> Do we want to allow user to define the keyframe id optionally and generate it by default?

This would be a neat feature because often times animations are used only once. In fact I twitted [this same idea](https://twitter.com/giuseppegurgone/status/777862367183011840) once and made a [prototype](https://twitter.com/giuseppegurgone/status/777883190094618624)

kof:
Good catch regarding #324, haven't thought about this particular case yet. Needs to be considiered!

kof:
The question is though how would you access the generated keyframe id in other style sheets if you want to reuse the keyframes in different parts of the application.

kof:
Also another question how to access it locally in the same sheet:

```js
const sheet = jss.createStyleSheet({
  '@keyframes fadeIn': {
    from: {opacity: 0},
    to {opacity: 1}
  },
  animationName: '$fadeIn',
  // or
  animation: '$fadeIn steps(4, end) 1s infinite'
})
```

iamstarkov:
nice

iamstarkov:
thanks

kof:
[...]

PR #512, opened 2017-06-01
Fix dynamic values for conditional and keyframe rules

lttb:
Fix https://github.com/cssinjs/jss/issues/511

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "babel-plugin-transform-es3-property-literals": "6.8.0",
     "babel-plugin-transform-flow-strip-types": "^6.18.0",
     "babel-plugin-transform-inline-environment-variables": "6.8.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-es2015": "6.18.0",
     "babel-preset-stage-0": "6.16.0",
     "codecov": "1.0.1",
--- a/src/RulesContainer.js
+++ b/src/RulesContainer.js
@@ -1,6 +1,7 @@
 /* @flow */
 import createRule from './utils/createRule'
 import updateRule from './utils/updateRule'
+import linkRule from './utils/linkRule'
 import type {
   RulesContainerOptions,
   ToCssOptions,
@@ -127,6 +128,17 @@ export default class RulesContainer {
 
     for (let index = 0; index < this.index.length; index++) {
       updateRule(this.index[index], name, RulesContainer)
+    }
+  }
+
+  /**
+   * Link renderable rules with CSSRuleList.
+   */
+  link(cssRules: CSSRuleList): void {
+    for (let i = 0; i < cssRules.length; i++) {
+      const cssRule = cssRules[i]
+      const rule = this.get(this.options.sheet.renderer.getSelector(cssRule))
+      if (rule) linkRule(rule, cssRule)
     }
   }
 
--- a/src/backends/DomRenderer.js
+++ b/src/backends/DomRenderer.js
@@ -36,12 +36,35 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string): 
   return true
 }
 
+function extractSelector(cssText: string, from: number = 0) {
+  return cssText.substr(from, cssText.indexOf('{') - 1)
+}
+
+const CSSRuleTypes = {
+  STYLE_RULE: 1,
+  KEYFRAMES_RULE: 7
+}
 
 /**
  * Get the selector.
  */
-function getSelector(rule: CSSStyleRule): string {
-  return rule.selectorText
+function getSelector(rule: CSSOMRule): string {
+  if (rule.type === CSSRuleTypes.STYLE_RULE) return rule.selectorText
+  if (rule.type === CSSRuleTypes.KEYFRAMES_RULE) {
+    const {name} = rule
+    if (name) return `@keyframes ${name}`
+
+    /**
+     * there is no CSSKeyframesRules.name for browsers like
+     * - IE 9
+     * - Safari 7.1.8
+     * - Mobile Safari 9.0.0
+     */
+    const {cssText} = rule
+    return `@${extractSelector(cssText, cssText.indexOf('keyframes'))}`
+  }
+
+  return extractSelector(rule.cssText)
 }
 
 /**
--- a/src/backends/DomRenderer.js
+++ b/src/backends/DomRenderer.js
@@ -54,12 +54,10 @@ function getSelector(rule: CSSOMRule): string {
     const {name} = rule
     if (name) return `@keyframes ${name}`
 
-    /**
-     * there is no CSSKeyframesRules.name for browsers like
-     * - IE 9
-     * - Safari 7.1.8
-     * - Mobile Safari 9.0.0
-     */
+    // There is no rule.name in the following browsers:
+    // - IE 9
+    // - Safari 7.1.8
+    // - Mobile Safari 9.0.0
     const {cssText} = rule
     return `@${extractSelector(cssText, cssText.indexOf('keyframes'))}`
   }
--- a/tests/functional/sheet.js
+++ b/tests/functional/sheet.js
@@ -9,8 +9,10 @@ import {
   computeStyle,
   getStyle,
   getCss,
+  getCssFromSheet,
   getRules,
-  removeWhitespace
+  removeWhitespace,
+  removeVendorPrefixes
 } from '../utils'
 
 const settings = {generateClassName}
@@ -226,7 +228,7 @@ describe('Functional: sheet', () => {
     let sheet
 
     beforeEach(() => {
-      sheet = jss.createStyleSheet().attach()
+      sheet = jss.createStyleSheet({}, {link: true}).attach()
       sheet.addRule('a', {color: 'red'})
       // It is important to use exactly this query, because
       // IE will add "all" always when `cssRules.insertRule` is used,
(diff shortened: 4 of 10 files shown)


src/RulesContainer.js
```diff
@@ -103,6 +104,7 @@ export default class RulesContainer {
     if (rule.name) this.map[rule.name] = rule
     if (rule.className && rule.name) this.classes[rule.name] = rule.className
     if (rule.selector) this.map[rule.selector] = rule
+    if (rule.selectorId) this.map[rule.selectorId] = rule
```
lttb: maybe we can use something one of this (store just rule.selectorId if exists without selector)

src/backends/DomRenderer.js
```diff
@@ -41,7 +41,8 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string):
  * Get the selector.
  */
 function getSelector(rule: CSSStyleRule): string {
-  return rule.selectorText
+  if (rule.selectorText) return rule.selectorText
```
kof: its a getter, don't call it twice

src/backends/DomRenderer.js
```diff
@@ -41,7 +41,8 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string):
...
+  return rule.cssText.substr(0, rule.cssText.indexOf('{') - 1).replace(' ', '')
```
kof: .replace(' ', '') this won't work when there are 3 spaces

kof: Also we need a comment on why this hack is needed

src/plugins/ConditionalRule.js
```diff
@@ -18,6 +19,7 @@ export default class ConditionalRule {
 
   constructor(selector: string, styles: Object, options: RuleOptions) {
     this.selector = selector
+    this.selectorId = selector.replace(' ', '')
```
kof: still don't like the naming and the replace here

kof: will think of something better

kof: alsowe need to move from this.selector to this.condition

tests/utils.js
```diff
@@ -21,6 +21,12 @@ export function getCss(style) {
       .join('')
 }
 
+export function getCssFromSheet(sheet) {
+  return [...sheet.renderer.getRules()]
```
lttb: there is no `Array.from` in old browser, so some tests fall.
should we use polyfills for tests/`test env` or rewrite it with for?

src/RulesContainer.js
kof: I would just use the rule.condition

tests/utils.js
kof: Is babel using Array.from? I would simply list the required polyfils in the readme

src/backends/DomRenderer.js
lttb: > .replace(' ', '') this won't work when there are 3 spaces
  
ah, lol, my bad

src/RulesContainer.js
```diff
@@ -131,6 +133,17 @@ export default class RulesContainer {
   }
 
   /**
+   * Link renderable rules with CSSRuleList.
+   */
+  link(cssRules: CSSRuleList): void {
+    for (let i = 0; i < cssRules.length; i++) {
+      const CSSStyleRule = cssRules[i]
```
kof: I am a bit wondering why we call a variable like this) I think I started to do that but its not really nice convention-wise, maybe simple `cssRule` is better. Also the idea to call them `renderable` was in order to abstract away from cssom/dom terminology because everything dom related should be in DomRenderer

tests/utils.js
lttb: yeah it fallbacks to Array.from

https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=es2015%2Creact%2Cstage-2&targets=&browsers=&builtIns=false&debug=false&code=const%20a%20%3D%20%5B...b%5D

src/backends/DomRenderer.js
kof: also cssText is a getter, don't call it twice

src/plugins/ConditionalRule.js
lttb: but `condition` isn't feet enough for rules like `keyframe`

kof: keyframeRule is a separate class

lttb: but we need id without spaces for keyframes too, because CSSKeyframesRule does not have `selectorText` too

lttb: here is the same trick https://github.com/cssinjs/jss/pull/512/files#diff-aeadf25904d143f46b87eef264b77f87R22

kof: CSSKeyframesRule should have a name https://developer.mozilla.org/de/docs/Web/API/CSSKeyframesRule

lttb: but we store it like `@keyframe name`, and `CSSKeyframesRule.name` would be just `name`

you suggest to build `selector` by hands like 
```
`@keyframe ${CSSKeyframesRule.name}`?
```

kof: https://github.com/cssinjs/jss/issues/346

src/utils/getSelectorId.js
```diff
@@ -0,0 +1,5 @@
+const spacesRe = /\s/g
+
+export default function getSelectorId(selector) {
```
lttb: we need to think about naming so

tests/utils.js
lttb: added `transform-runtime` here https://github.com/cssinjs/jss/pull/512/commits/4a65da7a9b6ad199d4aee5351e768386a3134c02

src/backends/DomRenderer.js
```diff
@@ -36,12 +36,35 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string):
...
+  KEYFRAMES_RULE: 7
+}
 
 /**
  * Get the selector.
  */
-function getSelector(rule: CSSStyleRule): string {
-  return rule.selectorText
+function getSelector(rule: CSSOMRule): string {
+  if (rule.type === CSSRuleTypes.STYLE_RULE) return rule.selectorText
```
kof: have you checked if rule.type is a getter? If it is - lets not call it more than once.

src/backends/DomRenderer.js
```diff
@@ -36,12 +36,35 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string):
...
+  if (rule.type === CSSRuleTypes.KEYFRAMES_RULE) {
+    const {name} = rule
+    if (name) return `@keyframes ${name}`
+
+    /**
+     * there is no CSSKeyframesRules.name for browsers like
```
kof: Better wording:

"There is no `rule.name` in the following browsers:"

kof: please use single line comments inside of functions/methods

src/backends/DomRenderer.js
```diff
@@ -36,12 +36,35 @@ function setStyle(rule: HTMLElement|CSSStyleRule, prop: string, value: string):
...
+     * - IE 9
+     * - Safari 7.1.8
+     * - Mobile Safari 9.0.0
+     */
+    const {cssText} = rule
+    return `@${extractSelector(cssText, cssText.indexOf('keyframes'))}`
```
kof: you are not extracting a selector here, you are extracting the keyframe name. extract Selector function doesn't make sense here semantically.

kof: lets make a bench to see what is faster, substr + indexOf or a regex

package.json
```diff
@@ -37,6 +37,7 @@
     "babel-plugin-transform-es3-property-literals": "6.8.0",
     "babel-plugin-transform-flow-strip-types": "^6.18.0",
     "babel-plugin-transform-inline-environment-variables": "6.8.0",
+    "babel-plugin-transform-runtime": "^6.23.0",
```
kof: we have removed it some time ago, in cases like other libraries using this lib as a dependency it increases the build too much.

src/backends/DomRenderer.js
lttb: it should be static attribute, but let's store it in variable anyway

tests/functional/sheet.js
```diff
@@ -590,6 +605,12 @@ describe('Functional: sheet', () => {
       expect(getCss(getStyle())).to.be(removeWhitespace(sheet.toString()))
     })
 
+    it('should render sheet with updated props after attach', () => {
+      sheet.attach().update({color: 'green'})
+      const style = removeVendorPrefixes(getCssFromSheet(sheet))
```
kof: why do we need to remove prefixes?

src/backends/DomRenderer.js
lttb: let's use substr/substring, because it's faster https://esbench.com/bench/5930564299634800a034843d

kof:
ping

package.json
lttb: we use this package just for tests here https://github.com/cssinjs/jss/pull/512/files#diff-004e32ee0ee65a48e2e1924046cec030

tests/functional/sheet.js
lttb: because some browsers automatically add their vendors to the keyframes rule, lol

tests/utils.js
```diff
@@ -21,6 +21,16 @@ export function getCss(style) {
...
+    .map(rule => removeWhitespace(rule.cssText))
+    .join('')
+}
+
+export function removeVendorPrefixes(str) {
```
kof: lets add a comment why and where we have to do that

kof:
published in 7.1.6

PR #519, opened 2017-06-14
Fix CSSOM types

lttb:
There are some hacks:
- we need to use generic type, because there is no override for extended interfaces
- we need to use hacks like `1|1` because it looks like flow casts `$PropertyType<T, prop>` for constant numbers into type `number` (invokes `typeof`?)


kof:
merged

PR #526, opened 2017-06-29
Support simplified template literals

kof:
The goal here is not to provide a full CSS parser. It supports only a StyleRule and only the style declaration as a string. It does basic tolerance support like omitting semicolons or newlines and trimming spaces though.

There are 2 motivations behind it:
1. Give user lower case, dashes and values without quotes syntax, they know from styled-components
2. Provide a handy tool for copy paste from dev tools.

### DIFF
--- a/docs/json-api.md
+++ b/docs/json-api.md
@@ -16,7 +16,7 @@ const styles = {
 Compiles to:
 
 ```css
-.button-jss-0 {
+.button-0-0 {
   color: red;
   font-size: 12px;
 }
@@ -69,11 +69,11 @@ const styles = {
 Compiles to:
 
 ```css
-.button-jss-0 {
+.button-0-0 {
   width: 100px;
 }
 @media (min-width: 1024px): {
-  .button-jss-0 {
+  .button-0-0 {
     width: 200px;
   }
 }
@@ -141,7 +141,7 @@ const styles = {
 Compiles to:
 
 ```css
-.container--jss-0-0 {
+.container-0-0 {
   background: red;
   background: linear-gradient(to right, red 0%, green 100%);
 }
@@ -245,7 +245,7 @@ const styles = {
 Compiles to:
 
 ```css
-.button-12345 {
+.button-0-1 {
   border: 1px solid red, 1px solid blue;
 }
 ```
@@ -266,7 +266,7 @@ const styles = {
 Compiles to:
 
 ```css
-.button-12345 {
+.button-0-1 {
   border: 1px solid red, 1px solid blue;
 }
 ```
@@ -283,7 +283,7 @@ const styles = {
 Compiles to:
 
 ```css
-.button-12345 {
+.button-0-1 {
   margin: 5px 10px;
 }
 ```
@@ -313,7 +313,7 @@ const styles = {
 Compiles to:
 
 ```css
-.button-jss-0-1:after {
+.button-0-1:after {
   content: "JSS"
 }
 ```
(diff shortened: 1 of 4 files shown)


kof:
@cssinjs/core wdyt?

kof:
@thysultan do you think I can do better from the performance perspective?

src/plugins/stringStyle.js
```diff
@@ -0,0 +1,32 @@
...
+// This is a very optimistic and simplified implementation.
+// It is not intended to implement a full CSS parser here.
+const toStyle = (cssText) => {
+  const style = {}
+  const split = cssText.trim().split(nlOrSemi)
+  for (let i = 0; i < split.length; i++) {
+    const item = split[i]
+    if (!item) continue
+    const colonIndex = item.indexOf(':')
+    // Missing colon, must by a syntax error.
```
thysultan: You could also do another `indexOf(' ')` to auto insert the missing semicolon in effect monkey patching the syntax error.

kof: That would be done at runtime, we can avoid it by simply warning the user. But yeah also a possibility.

src/plugins/stringStyle.js
```diff
@@ -0,0 +1,32 @@
+/* @flow */
+import createRule from '../utils/createRule'
+import type {JssStyle, RuleOptions, Rule} from '../types'
+
+const nlOrSemi = /\n|;/
```
thysultan: Depending on how far you want to go to make sure no semicolons work, this will fail with valid input like.

```css
box-shadow: 0, 0, 0 #fff,
                       0, 0, 0 #ccc   ,    
                       0, 0, 0 #000
```

Though this might help there.

```js
/[^,]\s*\n|;/
```

kof: Right!

kof: Maybe I simply should warn when no semicolon found before a newline.

kof: To avoid all the edge cases all together.

thysultan: Also i think this will fail with base64 urls that might have the token `;` in them or any property with a string or function `()` with the token inside of it.

You can do a lookahead to make sure there are no string or function tokens `", ', (, )` but since you can chain css rules in one line that might produce false positives.

i.e

```css
color: red; background: url(http://google.com?s=;); color: rgba(0,0,0);
```

kof: Hmm maybe for such a simplistic parse its better to enforce newlines and semicolons? Meaning I would split by semicolon + newline and not support one liners and no semicolons.

thysultan: That would work, thankfully css one liners are not a common pattern.

geelen:
This is great. CSS SYNTAX FOR THE CSS SYNTAX GOD 🔥

kof:
Also I think this still should go into a separate plugin but can be part of the default preset.

kof:
Also cc @tuchk4

docs/json-api.md
```diff
@@ -335,11 +335,30 @@ const styles = {
...
+.button-0-1 {
   color: '#0000B3'
 }
 ```
 
+## Alternative syntax using string templates
+
+You can use a template string literal or a regular string to declare the props and values.
+This notation is only supported for style declarations, nested rules, media queries, keyframes and others still have to be objects.
+
```
lttb: I think we need to add that function values are not supported this way

kof: Yep. I am still thinking maybe its not too hard to add support for them… Kinda missing the opportunity. My problem is that this would require a public function to export and I wanted to contain this within the plugin.

lttb: I see the only one way to support them - with tags, otherwise function value will cast to string value

lttb: so with tag function like `css` it can be something like

```js
button: css`
  padding: 10px;
  color: ${theme => theme.primary ? 'red' : 'black'}
`
```

this is also an approach for static optimizations via babel

kof: Yes, thats why I say that we would need to export a function like `css` for e.g. from jss, while In wanted it to be a pure syntactic sugar implemented via plugin.

kof: Btw. its not a problem for styled-jss, because you got a function anyways. Its a problem for regular styles map.

kof: Of course I could export that function from the plugin for e.g. `import {css} from 'jss-string-style'` if its optional, maybe its ok?

kof: Actually, its ok for jss core, react-jss and styled-jss can have it built in.

kof:
released https://github.com/cssinjs/jss-template

PR #529, opened 2017-07-02
Add size-limit

ai:
[size-limit](https://github.com/ai/size-limit) is a tool to show how many kilobytes your JS library increases the user bundle. You can set size budget and size-limit will prevent JavaScript libraries bloat on Travis CI.

I already use it in Logux and PostCSS and it helps me a lot. For instance, I reduce Logux Client size 2x because I found some unexpected huge dependencies.

@kof BTW, I was excited that JSS size is just 5.5 KB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "mocha": "^3.2.0",
     "pre-commit": "^1.1.3",
     "rimraf": "^2.5.4",
+    "size-limit": "^0.2.0",
     "webpack": "^2.3.3"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


kof:
I like it, thanks for the pr, will merge as soon as I find out whats wrong with the CI. At the first glance  we are missing size-limit in package.json and package-lock.json

ai:
Oops, I added it to `package.json`, but I have no `npm@5` yet to update `package-lock.json`. Can you help me here?

kof:
yep

PR #532, opened 2017-07-02
Consistently use JSS instead of three different styles

sapegin:


kof:
got conflicts

sapegin:
Fixed ;-)

PR #537, opened 2017-07-10
[experiment]Wip rollup

kof:
umd build max 65kb => 45kb
umd build min 23kb => 17kb
umd build min gzip 6kb => 5kb

## Todo
- [ ] karma config
- [ ] lib build
- [ ] size-limit


iamstarkov:
nay, nay, why?

iamstarkov:
you should just ship esm

Issue #547, opened 2017-07-27
insertionPoint option is not used for createStyleSheet function


darkowic:
Checked the code
```
  createStyleSheet(styles: Object, options: StyleSheetFactoryOptions = {}): StyleSheet {
    let index = options.index
    if (typeof index !== 'number') {
      index = sheets.index === 0 ? 0 : sheets.index + 1
    }
    const sheet = new StyleSheet(styles, {
      ...options,
      jss: (this: Jss),
      generateClassName: options.generateClassName || this.generateClassName,
      insertionPoint: this.options.insertionPoint,
      Renderer: this.options.Renderer,
      index
    })
    this.plugins.onProcessSheet(sheet)
    return sheet
  }
```

Even though you pass your `insertionPoint` in options it will be overwritten by default setting...

Found it while using new material ui beta release.

kof:
Do you have a use case for an override of the insertionPoint option over createStyleSheet?

darkowic:
Material ui beta version
```
          const sheet = this.jss.createStyleSheet(styles, {
            meta: name,
            name,
            insertionPoint: CHANNEL,
            link: false,
            ...this.sheetOptions,
            ...currentStyleSheet.options,
          });
```

I want to use `insertionPoint` to place styles in correct place

darkowic:
I'm changing it right now. Will do pull request soon

kof:
You are changing what?

darkowic:
@kof here in material ui beta 
https://github.com/callemall/material-ui/blob/v1-beta/src/styles/withStyles.js#L140
`createStyleSheet` is used to create style sheet for specific component


In material ui code they expect (incorrectly btw) that insertion point
```
insertionPoint: CHANNEL,
```
 will be overridden with `CHANNEL` ('material-ui'). I simply want to insert style sheets in different place in my application to be able to override styles without '!important' all the time (I use styled components).

Maybe @oliviertassinari can confirm that this is behaviour he expects

oliviertassinari:
> insertionPoint option is supposed to be a Jss option

Oops, I have missed that point. I thought it was supported after reading the [jss-theme-reactor](https://github.com/nathanmarks/jss-theme-reactor/blob/master/src/styleSheet.js#L9) source code.

@kof What do you want to do here? I can update the styling solution of Material-UI accordingly.

kof:
[...]

PR #548, opened 2017-07-27
Allow to override default settings in createStyleSheet function

darkowic:
This will fix issue https://github.com/cssinjs/jss/issues/547

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ lib
 dist
 .env
 coverage
+.idea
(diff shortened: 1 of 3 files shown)


kof:
Its not clear if this is an issue, insertionPoint option is supposed to be a Jss option. I need to know first why it would be required to be able to override them over createStyleSheet.

darkowic:
Continue discussion in the issue.

PR #550, opened 2017-07-28
[experiment] support template literals

kof:
This PR is here just in order to not loose this experiment, see [original discussion](https://github.com/cssinjs/jss/pull/526)


kof:
released as a standalone plugin https://github.com/cssinjs/jss-template

PR #551, opened 2017-07-28
[WIP] Next 

kof:
This will be JSS 9 release.


src/utils/createGenerateClassName.js
```diff
@@ -15,5 +21,23 @@ const jssCounter = globalRef[namespace]++
...
+    if (ruleCounter > maxRules) {
+      warning(
+        false,
+        'You might have a memory leak. Rule counter is at %s.',
+        ruleCounter
+      )
+    }
+
+    if (env === 'production') {
+      return `c${jssCounter}${ruleCounter}`
```
oliviertassinari: @thangngoc89 Here is the `createGenerateClassName` version Material-UI will most likely use in the future.
I'm closing https://github.com/callemall/material-ui/pull/7650. Please defend you use case here :).

thangngoc89: Please make this configureable. For some legacy reasons, I have some global styles that was created in this same manner. I'm using a fork of Material UI to avoid this right now and would love to have an option that I can easily configure instead of a fork

kof: it is configurable, you can pass your own  class name generator function http://cssinjs.org/js-api?v=v9.0.0-pre#generate-your-own-class-names

oliviertassinari: @kof Is right, we have examples in the documentation of Material-UI. No need to fork. Also given the edge case you are facing I would vote for people like you creating their own `createGenerateClassName` function, at their own risk.

oliviertassinari: The other use case I'm aware of for a custom `createGenerateClassName` is for people wanting predictible classNames, so they can overrides the style with global CSS selector.

kof: Yeah, its a bad technic anyways. Also they can use jss-compose for that.

thangngoc89: @oliviertassinari I don't think I can override current `createGenerateClassName ` in Material UI (1 beta 3) easily. You used it in several places and changing it in jss object doesn't do anything ? 

I'm I implementing it the wrong way?

oliviertassinari: https://github.com/callemall/material-ui/blob/v1-beta/docs/src/components/App.js#L50

kof:
I am about to release it :)

kof:
MERGED

PR #553, opened 2017-07-31
Update setup.md

oliviertassinari:
That doesn't work given [the following priorities](https://github.com/cssinjs/jss/blob/f424bb04e614234a91f51c7171fd418f2b97e254/src/Jss.js#L65).


kof:
thanks

PR #555, opened 2017-08-03
Fixed warning in sheets manager

HenriBeck:
The warn function had a missing argument and didn't output the passed in the message but instead threw an error because it was missing an argument

It would throw [this](https://github.com/BerkeleyTrue/warning/blob/master/warning.js#L31)


kof:
thanks

PR #568, opened 2017-08-21
Fix createGenerateClassName docs

sapegin:
I may miss something but that’s the only way it works for me.

### DIFF
--- a/docs/js-api.md
+++ b/docs/js-api.md
@@ -312,7 +312,7 @@ console.log(sheet.toString())
 
 ## Generate your own class names
 
-`createGenerateClassName()`
+`createGenerateClassName`
 
 Option `createGenerateClassName` allows you to specify a function which returns a class name generator function `generateClassName(rule, sheet)`. This pattern is used to allow JSS reset the counter upon factory invocation, when needed. For e.g. it is used in [react-jss](https://github.com/cssinjs/react-jss) to reset the counter on each request for SSR.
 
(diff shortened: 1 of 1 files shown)


kof:
why? both is valid

kof:
But yeah maybe it is more clear to use setup?

sapegin:
For some reason `create` didn’t work. Maybe I did something wrong. But the first commit should be right anyway ;-)

kof:
did you do a setup after create? if so it overrides … mb I need to allow multiple .setup calls to extend each other.

sapegin:
Yeah, I did. And it’s surprising behavior ;-)

kof:
I was thinking create(options) is same like create().setup(options) ... so one would use one of them, but yeah ...

sapegin:
Is there any technical requirement to overwrite previous options?

kof:
No, but thats what you expected, right?

sapegin:
I’m not 100% sure, worth a poll ;-) Maybe docs should be clearer that setup and create is almost the same thing.

kof:
Well create creates an instance, setup sets up an existing instance. They are not the same, but create is using setup

sapegin:
Also `use` doesn’t overwrite settings and looks like it’s not documented on this page ;-)

kof:
Maybe we just need an info that create is calling setup and that setup redefines a setup.

sapegin:
Yeah, would be nice. I’ll try to add.

kof:
use is always adding a plugin, yes it is basically extending, but use is also designed for subsequent calls ...

sapegin:
Also I don’t understand this part:

> Note: in production it uses just the "c" + rules counter.

How to enable production mode?

kof:
NODE_ENV=production ... yeah lets add this too.

sapegin:
Seems not to work for me. (Even before I’ve added custom names.)

sapegin:
Or is it 9.x specific?

kof:
yes it is a new feature, exists only in next releases

sapegin:
Will it affect custom names?

kof:
nope, if user has own class name generation function, they will have to do the same in order to support it.

PR #585, opened 2017-09-20
chore: add `es` build.

umidbekkarimov:
screen shot 2017-09-20 at 16 52 54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ node_modules
 npm-debug.log
 tmp
 *~
+es
 lib
 dist
 .env
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "jss",
   "version": "8.1.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "abbrev": {
       "version": "1.0.9",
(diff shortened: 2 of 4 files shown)


kof:
Thanks for the PR!

Why do you need rollup for that?

umidbekkarimov:
Webpack adds too much extra code.

That can be avoided with [ModuleConcatenationPlugin](https://webpack.js.org/plugins/module-concatenation-plugin/) but it works only with webpack@3 (and I'm not sure output will be smaller than rollup output).

Currently rollup is a winner in case of libs.

See also [good post](https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c) from rollup maintainer.

kof:
I was actually experimenting with both, while rollup gives the immediate advantage, webpack is a better overall project to choose. Lets compare what would be the difference and if we can find a webpack plugin to do the best possible output.

umidbekkarimov:
es: webpack@3 + 'ModuleConcatenationPlugin'
old/es: rollup

screen shot 2017-09-20 at 17 39 49

dist: webpack@3 
old/dist: webpack@2

screen shot 2017-09-20 at 17 39 55

umidbekkarimov:
Webpack has it's own internal functionality and module system, thats what makes it great for apps, but not so good for libraries.

Issue #607, opened 2017-10-31
Observable rules


appsforartists:
This should work:

```typescript
const buttonStyle$ = Observable.of({
  backgroundColor: 'blue',
  fontWeight: 'bold',
});

jss.createStyleSheet(
  {
    button: buttonStyle$,
  },
  {  
    link: true,
  },
);
```

appsforartists:
Looking at [the docs](http://cssinjs.org/plugins?v=v9.0.0), this seems like it would be a pretty simple plug-in: 

1. Listen for new rules in `onCreateRule`.
2. If `decl` is observable,
  1. Return a new rule
  2. ```typescript
      decl.subscribe(
        (currentProps) => // sync resulting rule's props to currentProps
      );
      ```
3. If not, return an empty value

kof:
Yep, pretty sure we can do this.

kof:
Regarding deletion - we normally don't delete props, we set values, there is always a default value you can set to achieve what you want. For e.g. instead of removing property opacity use 0.

kof:
CSSOM API does support though removing https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration

appsforartists:
Interesting.  I don't have a concrete use case for deletion either - just seemed like a good idea to clean up after ourselves.  FWIW, deletion and defaults would behave differently for inherited properties like color, but I've yet to build an interaction where I needed to delete a k:v pair.

Issue #514, opened 2017-06-09
Function values don't remove previous value when returning "null" or "undefined"


jaminthorns:
I think this might be due to the nature of how JSS converts JavaScript values to CSS, but I've noticed that if I have a function-valued rule, and that function returns a valid CSS value, then later returns `null` or `undefined`, the CSS value never gets removed from the stylesheet.

Intuitively, this should remove the CSS value, as invalid JavaScript values should not translate over to CSS values. Of course, I can also see why this would happen, since I'm guessing the JS -> CSS value translation code (which I haven't actually dug into), simply short circuits its calculation of the value when it sees that it's invalid.

If this isn't considered an issue, then feel free to close it. Here's an example of the issue: [Link](https://www.webpackbin.com/bins/-KmDhsKtqLZ3BCqqK0ae)

kof:
I don't think its a bug but it might be a useful enhancement. Need to think a bit about its implications.

kof:
One negative aspect of this might be the following: if the function that returns the value is not trivial or calls some other function and its result is unexpectedly falsy, you might end up with a removed property in an unexpected way and then wonder whats going on.
So this API proposal has n implicitness in it.

lttb:
maybe we can remove property only if value is `null`? because `undefined`/`false` may be really implicitness

jaminthorns:
@kof If the very 1st call of a function used as a JSS function value returns some invalid CSS value and doesn't add the CSS rule to the stylesheet (which is the current behavior), then why would a later call with an invalid return value removing the rule be unexpected? Is there another scenario I'm not thinking of?

kof:
>  If the very 1st call of a function used as a JSS function value returns some invalid CSS value and doesn't add the CSS rule to the stylesheet

Thats simply because we don't verify if the value was valid and successfully valid)

kof:
check this out https://github.com/cssinjs/jss/pull/678

kof:
released

PR #609, opened 2017-10-31
WIP add support for Observable rules

appsforartists:
I've started writing a plugin that adds support for Observable rules (#607).  Right now, I'm seeing this warning:

```
Rule is not linked. Missing sheet option "link: true".
```

Even though `{ link: true }` is clearly specified in the tests.

### DIFF
--- a/src/plugins/observables.js
+++ b/src/plugins/observables.js
@@ -8,11 +8,11 @@ export default {
   onCreateRule(name: string, decl: JssStyle, options: RuleOptions): Rule | null {
     if (isObservable(decl)) {
       // Cast `decl` to `Observable`, since it passed the type guard
-      const props$ = (decl: Observable<{[string]: string}>);
+      const props$ = (decl: Observable<{[string]: string}>)
 
       // We know `rule` is a `StyleRule`, and the other types don't have a
       // `prop` method, so we must explicitly cast to `StyleRule`
-      const rule = ((createRule(name, {}, options): any): StyleRule);
+      const rule = ((createRule(name, {}, options): any): StyleRule)
 
       // `stream.subscribe()` returns a subscription, which should be explicitly
       // unsubscribed from when we know this sheet is no longer needed, but I
@@ -25,13 +25,13 @@ export default {
             // The typing here is gross because `Object.entries` is typed as
             // `[string, mixed]`, even though `props` is `{[string]: string}`
             ([key, value]) => rule.prop(key, ((value: any): string))
-          );
+          )
         }
-      );
+      )
 
-      return rule;
+      return rule
     } else {
-      return null;
+      return null
     }
   },
 
--- a/src/plugins/observables.js
+++ b/src/plugins/observables.js
@@ -1,9 +1,26 @@
 /* @flow */
 import isObservable from 'is-observable'
 import StyleRule from '../rules/StyleRule'
-import type {Rule} from '../types'
+import createRule from '../utils/createRule'
+import type {Observable, Rule, RuleOptions, JssStyle} from '../types'
 
 export default {
+  onCreateRule(name: string, decl: JssStyle | Observable<{[string]: string}>, options: RuleOptions) {
+    if (isObservable(decl)) {
+      const rule = createRule(name, {}, options);
+
+      decl.subscribe(
+        (currentProps) => {
+          Object.entries(currentProps).forEach(
+            ([key, value]) => rule.prop(key, value)
+          );
+        }
+      );
+
+      return rule;
+    }
+  },
+
   onProcessRule(rule: Rule) {
     if (!(rule instanceof StyleRule)) return
     const {style} = rule
--- a/src/plugins/observables.js
+++ b/src/plugins/observables.js
@@ -5,19 +5,33 @@ import createRule from '../utils/createRule'
 import type {Observable, Rule, RuleOptions, JssStyle} from '../types'
 
 export default {
-  onCreateRule(name: string, decl: JssStyle | Observable<{[string]: string}>, options: RuleOptions) {
+  onCreateRule(name: string, decl: JssStyle, options: RuleOptions): Rule | null {
     if (isObservable(decl)) {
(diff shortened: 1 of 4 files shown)


src/plugins/observables.js
```diff
@@ -1,9 +1,40 @@
...
 import StyleRule from '../rules/StyleRule'
-import type {Rule} from '../types'
+import createRule from '../utils/createRule'
+import type {Observable, Rule, RuleOptions, JssStyle} from '../types'
 
 export default {
+  onCreateRule(name: string, decl: JssStyle, options: RuleOptions): Rule | null {
+    if (isObservable(decl)) {
+      // Cast `decl` to `Observable`, since it passed the type guard
+      const props$ = (decl: Observable<{[string]: string}>);
```
kof: linter will fail, because I am using no-semicolon style)

kof: why renaming decl to props?

kof: I would call it decl$

src/plugins/observables.js
```diff
@@ -1,9 +1,40 @@
...
+
+      // We know `rule` is a `StyleRule`, and the other types don't have a
+      // `prop` method, so we must explicitly cast to `StyleRule`
+      const rule = ((createRule(name, {}, options): any): StyleRule);
```
kof: No idea how this works, you managed to cast to any and then to StyleRule haha

src/plugins/observables.js
```diff
@@ -1,9 +1,40 @@
...
+            // The typing here is gross because `Object.entries` is typed as
+            // `[string, mixed]`, even though `props` is `{[string]: string}`
+            ([key, value]) => rule.prop(key, ((value: any): string))
+          );
+        }
+      );
+
+      return rule;
+    } else {
+      return null;
```
kof: stylistic thing: I use early exits always in order to reduce indentation

src/plugins/observables.js
appsforartists: I was taking the name from `rule.prop`.  I can change to `decl` if you're prefer, but then you'd end up with shadowing:

```typescript
onCreateRule(name: string, decl: JssStyle, options: RuleOptions): Rule | null {
  // ...
  decl$.subscribe(
    // This is the emission from `decl$`, but might be confused with the argument
    (decl) => {
```

src/plugins/observables.js
```diff
@@ -1,9 +1,40 @@
...
+      const rule = ((createRule(name, {}, options): any): StyleRule);
+
+      // `stream.subscribe()` returns a subscription, which should be explicitly
+      // unsubscribed from when we know this sheet is no longer needed, but I
+      // don't see any hooks to do that in the plugin API.  The Observable props
+      // implementation doesn't store its subscription either, so this is a TODO
+      // for both.
+      props$.subscribe(
+        (props) => {
+          Object.entries(props).forEach(
```
kof: This could be a performance sensitive place, I normally try to avoid function calls and function generations as far as I can. I would simply use a for loop, because we expect only plain objects here

src/plugins/observables.js
appsforartists: Common pattern when you know more about the state of the system than the type-checker: cast to `any`, then cast to the correct type.  It's dangerous if you get it wrong though (you're explicitly telling Flow to ignore what looks like a type error) - it's effectively `!important` for types.

I believe Microsoft actually suggests this pattern in [the TypeScript docs](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions).

src/plugins/observables.js
kof: rule.prop() is a getter/setter for a property/value, but that object is a declaration - set of properties and values)

src/plugins/observables.js
kof: learned something new! thanks

kof:
checking the tests now

appsforartists:
I think the biggest problem is the linking.  If I call `rule.prop(key)` on either side of the set inside `subscribe`, everything is working correctly, but I don't think the sheet knows it's linked, so it's not updating.

kof:
Yes, the reason is it is not deploying an empty rule. It is empty because we set actual props/values later.

kof:
And so it has nothing to link to.

kof:
It basically hits an optimization which is about not rendering empty rules to the CSSOM if user created an empty one by because of some conditions, as rules might be generated by complex functions.

appsforartists:
Good catch.  Passing a useless value (`position: static`) fixes it.

Teaching the linker about empty dicts seems like a better solution than adding an arbitrary (hopefully) no-op style.  What do you think?

src/plugins/observables.js
```diff
@@ -13,7 +13,12 @@ export default {
 
     // We know `rule` is a `StyleRule`, and the other types don't have a
     // `prop` method, so we must explicitly cast to `StyleRule`
-    const rule = ((createRule(name, {}, options): any): StyleRule)
+    const rule = ((
+      // `link: true` will fail if `decl` is an empty dict, so we initialize it
+      // with `position: static`.  Since that's the default anyway, this
+      // shouldn't affect actual styling, but gives JSS something to bite into.
+      createRule(name, {position: 'static'}, options)
```
kof: nono please revert) I have a better idea, will add a real quick fix, give me a sec

kof: it fixes it but we can avoid it completely

appsforartists: Great.  I didn't like this solution anyway.  😄

appsforartists:
Just rolled back to `HEAD^` - awaiting your linker fix. 😃

kof:
Found one use case for removing props https://github.com/cssinjs/jss/issues/514
It would make sense to remove the prop if user returns a null or undefined

aslafy-z:
Have this been merged already ?

kof:
yes, if you got an issue, please file a bug

PR #610, opened 2017-10-31
Observable rules

kof:


src/plugins/observables.js
```diff
@@ -1,9 +1,34 @@
...
-import type {Rule} from '../types'
+import createRule from '../utils/createRule'
+import type {Observable, Rule, RuleOptions, JssStyle} from '../types'
 
 export default {
+  onCreateRule(name: string, decl: JssStyle, options: RuleOptions): Rule | null {
+    if (!isObservable(decl)) return null
+
+    // Cast `decl` to `Observable`, since it passed the type guard.
+    const style$ = (decl: Observable<{[string]: string}>)
```
kof: actually the value can be not just a string

kof: In case someone uses observable and relies on the full jss syntax

kof: cc @appsforartists

kof:
Ok, I think we are good now @appsforartists

kof:
@appsforartists let me know if you want to add anything, I will merge then

src/utils/createRule.js
```diff
@@ -9,7 +9,9 @@ import cloneStyle from '../utils/cloneStyle'
  */
 export default function createRule(name: string = 'unnamed', decl: JssStyle, options: RuleOptions): Rule {
   const {jss} = options
-  const declCopy = cloneStyle(decl)
+  // It has `isDynamic` property if an Observable rule was used, so no need to
+  // clone it.
+  const declCopy = decl && decl.isDynamic ? decl : cloneStyle(decl)
```
appsforartists: `cloneStyle` already passes through observables.  Is it worth having this handled in both places?

src/rules/StyleRule.js
```diff
@@ -65,7 +65,7 @@ export default class StyleRule implements BaseRule {
   prop(name: string, nextValue?: string): StyleRule|string {
     // The result of a dynamic value is prefixed with $ and is not innumerable in
```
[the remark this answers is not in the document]
kof: yep

src/utils/createRule.js
kof: cloneStyle would skip isDynamic property which is not enumerable, we need to keep it

kof: mb we can remove the check from cloneStyle

kof: decl in this case is actually not an Observable, it is the initialStyle object we pass to createRule from the observables plugin

appsforartists: Maybe fixup the comment then.

appsforartists: > `isDynamic` is set in styles that are entirely specified dynamically (e.g. written to by an Observable).  In that case, we don't need to clone the style.

kof: Maybe I should just create that object with some `InitialStyle`constructor so that I can identify it by constructor check instead of the non-enumerable property.

kof: At the end it still can't be cloned  without setting again the constructor ... so not sure if its a big win.

appsforartists: SGTM

kof:
merged 💥

kof:
Publishing it as 9.1.0 now.

PR #613, opened 2017-11-01
Support current symbol-observable & synchronous observable rules

appsforartists:
I tried using `jss@9.1` to add the `JSSDemo` back to `material-motion`.  I encountered two problems:

1. `link: true` doesn't seem to work if there is more than one observable rule in a style sheet.
2. `isObservable` is always failing, because it depends on an obsolete version of `symbol-observable`.

I added a (currently failing) test for 1) and fixed 2).  Feel free to take this PR over to fix 1). 😄

### DIFF
--- a/src/plugins/observables.js
+++ b/src/plugins/observables.js
@@ -10,7 +10,12 @@ export default {
 
     // Cast `decl` to `Observable`, since it passed the type guard.
     const style$ = (decl: Observable<{[string]: string}>)
-    const initialStyle = {}
+
+    // `style$` needs to be subscribed to before `rule` is created. Otherwise,
+    // any emissions that occur synchronously will be called on an unlinked
+    // `rule`, which will ignore them.
+    let rule: StyleRule
+    let initialStyle = {}
 
     // It can't be enumerable, otherwise it will be rendered.
     Object.defineProperty(initialStyle, 'isDynamic', {
@@ -18,18 +23,19 @@ export default {
       writable: false
     })
 
-    // We know `rule` is a `StyleRule`, and the other types don't have a
-    // `prop` method, so we must explicitly cast to `StyleRule`.
-    const rule = ((createRule(name, initialStyle, options): any): StyleRule)
-
     // TODO
     // Call `stream.subscribe()` returns a subscription, which should be explicitly
     // unsubscribed from when we know this sheet is no longer needed.
     style$.subscribe((style: JssStyle) => {
-      for (const prop in style) {
-        rule.prop(prop, style[prop])
+      if (rule) {
+        for (const prop in style) {
+          rule.prop(prop, style[prop])
+        }
       }
+      else if (Object.keys(style).length) initialStyle = style
     })
+
+    rule = (createRule(name, initialStyle, options): any)
 
     return rule
   },
--- a/src/plugins/observables.js
+++ b/src/plugins/observables.js
@@ -1,7 +1,7 @@
 /* @flow */
-import isObservable from 'is-observable'
 import StyleRule from '../rules/StyleRule'
 import createRule from '../utils/createRule'
+import isObservable from '../utils/isObservable'
 import type {Observable, Rule, RuleOptions, JssStyle} from '../types'
 
 export default {
--- a/src/utils/isObservable.js
+++ b/src/utils/isObservable.js
@@ -1,0 +1,4 @@
+import $$observable from 'symbol-observable'
+
+export default value => value && value[$$observable] && value === value[$$observable]()
+
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
   },
   "dependencies": {
     "is-in-browser": "^1.1.3",
-    "is-observable": "^0.2.0",
+    "symbol-observable": "^1.0.4",
     "warning": "^3.0.0"
   },
   "files": [
--- a/src/rules/StyleRule.js
+++ b/src/rules/StyleRule.js
@@ -63,11 +63,11 @@ export default class StyleRule implements BaseRule {
    * Get or set a style property.
    */
   prop(name: string, nextValue?: string): StyleRule|string {
-    // The result of a dynamic value is prefixed with $ and is not innumerable in
+    // The result of a dynamic value is prefixed with $ and is not enumerable in
     // order to be ignored by all plugins or during stringification.
     const $name = isDynamic(this.style[name]) ? `$${name}` : name
 
-    // Its a setter.
+    // It's a setter.
     if (nextValue != null) {
       // Don't do anything if the value has not changed.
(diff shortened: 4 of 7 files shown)


appsforartists:
Looks like my tests were failing because of quirks in my tests (e.g. `backgroundColor` vs`background-color`).  I've fixed those.

I've also tried to make synchronous observables work.  `initialStyle` is being set correctly, but its values don't seem to be used.

appsforartists:
Fixed initial style - tests should pass now.

appsforartists:
JSSDemo still gets mad at this:

```typescript
    container: {
      touchAction: 'none',
      width: '100vw',
      height: '100vh',
      overflow: 'hidden',
      background: '#F7DF1E',
    },

    foreground: foreground$,
    background: background$,
  }).attach().link();
```

It appears that `cssRule` is never getting created because `toString` is empty because `style.isDynamic` is becoming undefined somewhere.

```
// observable plugin:
initialStyle.isDynamic === true

// StyleSheet.attach
this.rules.index[0].style.isDynamic === undefined
```

appsforartists:
@kof 

What's the advantage of returning an empty string when `result === ''` in `toCss`?  It makes a bunch of tests fail, but unbreaks this demo.  Of course, it would be nice to know why `isDynamic` is vanishing and to resolve that issue, but the early return in `toCss` seems to make JSS more brittle with no obvious gain.  

Is there a good reason to have enough defined-but-empty rule lists that the gain from skipping an empty selector is worth the added complexity?

kof:
> What's the advantage of returning an empty string when result === '' in toCss?

So that nothing gets rendered and the return type is consistent.

kof:
> Is there a good reason to have enough defined-but-empty rule lists that the gain from skipping an empty selector is worth the added complexity?

Not sure if its worth it, lets see first if we can easily fix it. I see your point, there is added complexity.  The use case is not when user simply defines an empty rule and we skip it. The case is when user uses a function which has logic and at some point decides to not return styles so that rule becomes empty but its not obvious from the beginning.

src/utils/isObservable.js
```diff
@@ -0,0 +1,4 @@
+import $$observable from 'symbol-observable'
```
kof: we could have fixed is-observable as well, but yeah

appsforartists: True, but that's more work than I have scope for.  (Opening the PR; writing or generating CJS, ESM, and TS versions; waiting for a merge; updating JSS when it does).  This was faster, simpler, and removes a dependency.

kof: On the other hand, that package could take responsibility of identifying an Observable in the future, one place to have a reliable isObservable. I am not aware if it will ever change again, but then we fix it for every one.

appsforartists:
Would you get the same bug (JSS thinking you hadn't linked) if you returned an empty style the first time your function was called?  It naively seems like you would, but it depends on when/how the functional rules are called.

Debugging why `isDynamic` is disappearing is really frustrating because I have to rebuild JSS, restart my Pundle dev server, and step through a broken source map (thanks Pundle) in the debugger every time I touch JSS.  I would love to use JSS in more demos, but I don't know if I'll have more time to sink into chasing this down today.

In the mean time, this PR should be mergeable.  I wrote the multiple-observable-rule tests to try to catch the problem on the JSS side, but they pass.  I also fixed the observable plugin for observables that emit synchronously (e.g. `Subject`s).

kof:
Your tests are failing now, right? If so I can fix them. I just neeed something reproducible.

kof:
> catch the problem on the JSS side, but they pass.

Oh, so you have a bug in the demo which can't be reproduced by the tests?

appsforartists:
Yeah.

```
Finished in 0.455 secs / 0.237 secs @ 07:30:46 GMT-0700 (PDT)

SUMMARY:
✔ 341 tests completed
$ [ -z "$TRAVIS" ] || codecov
✨  Done in 10.64s.
```

I'm gonna quickly see if I can link material-motion into JSS and debug the failure from JSS's karma setup.

appsforartists:
@kof, take a look at this: http://codereview.cc/D3445

I fixed the issue in `JSSDemo` by using `jss.create().setup(preset()).createStyleSheet(…)` rather than `jss.setup(preset())`.  It appears there's a bug where `isDynamic` is unset when used with `jss.setup(preset())`.

kof:
Thats interesting, because jss.create().setup(preset()) and jss.setup(preset()) is exactly the same, the only difference is that the first one creates a new jss instance.

kof:
The question is still how to reproduce this in a test. Really weird one.

appsforartists:
@kof - thoughts on merging this?  I'd rather depend on jss@9.2 than on my own local branch. 😄

kof:
Wait, I didn't realize the issue is fixed, I thought you are still trying to create a test which reproduces the issue with the demo?

appsforartists:
So there are two things:

1. The initial implementation of Observable rules doesn't handle subjects or the current version `symbol-observable`
2. Multiple Observable rules don't play nicely with `jss.setup()`

The second could use some more attention, but this PR resolves the first and should be merged.

kof:
Got it, testing and then merging. Should we wait with the release for the second issue?

appsforartists:
Knowing that I have an easy workaround for it, I'm not that motivated to invest time in diagnosing and resolving it.  If you want to, go for it, but I don't think we should block 9.2 on it.

appsforartists:
#614 tracks the `isDynamic: undefined` bug with `jss.setup()`.

src/plugins/observables.js
```diff
@@ -10,27 +10,33 @@ export default {
...
     // unsubscribed from when we know this sheet is no longer needed.
     style$.subscribe((style: JssStyle) => {
-      for (const prop in style) {
-        rule.prop(prop, style[prop])
+      if (rule) {
+        for (const prop in style) {
+          rule.prop(prop, style[prop])
+        }
       }
+      else if (Object.keys(style).length) initialStyle = style
```
kof: Trying to figure out the logic. Why do we postpone the rule creation and use style as initialStyle? 

Why not always create the rule first, before subscribe, then just update props with received value, it can go sync, it doesn't matter.

kof: I think I got it.

appsforartists: If you call `rule.prop()` before linking happens, the style will never be set on `cssRule`, so we need somewhere to store the values that are emitted before linking.  `initialStyle` seems to be a good place to do that.

kof: I think you just discovered a completely different bug. Trying to fix it right now.

kof:
Ok got it all fixed, lets continue on #615

PR #615, opened 2017-11-02
Observable rules fixes

kof:


kof:
@appsforartists it probably also fixes the demo, check it out!

appsforartists:
Ugghhh…

I got a bit too cavalier and hard reset to your commit without taking note of my hash.  Neither yours, nor the most recent one I see in `git reflog` before I did that, is working. 🤦‍♂️ 

screen shot 2017-11-02 at 3 24 55 pm

kof:
Lets find out what's wrong, we can chat somewhere, twitter/gitter works.

appsforartists:
You wanna give me push access to this repo, so we don't end up closing and opening PRs every time someone rebases on the other's commits?

Here's my fix:

https://github.com/appsforartists/jss/commit/6e9bdb4c2851315e2bba307b8f8df3c5dfb3fda3

kof:
I think we need a better logic for identifying styles which can't be removed. On thing I realized is that rendering empty rules not a problem if it is all happening on the client. Optimization was mostly made for SSR.

I think we should just allow empty rules, but skip them if we do SSR.

kof:
Another option is if we see that the sheet is linked, we should allow empty rules.

kof:
Ok, I think I got it fixed now in a much better way and created an issue for the future simplification https://github.com/cssinjs/jss/issues/617

PR #618, opened 2017-11-03
Fn values and fn rules plugin

kof:
Extracted entire function values functionality into a core plugin. Based on #615


kof:
cc @lttb

PR #625, opened 2017-11-13
Add class name escaping for the dev mode

### LINKED ISSUE #624: Classes does not work for Components with special chars in display name
@luiz290788 commented on [Mon Nov 13 2017](https://github.com/cssinjs/react-jss/issues/176)

The new prefix for the generated classes that use the display name of the decorated components are not being sanitized.

For example:
Decorated class display name "SomeOtherDecorator(Test)"
Generated classes ".SomeOtherDecorator(Test)-root-0-1"
The parenthesis in the classes is causing the problems here.

Suggested solution: Parse the display name of the component to remove the special characters.

---

@kof commented on [Mon Nov 13 2017](https://github.com/cssinjs/react-jss/issues/176#issuecomment-343913293)

Or escape the chars.

---

@luiz290788 commented on [Mon Nov 13 2017](https://github.com/cssinjs/react-jss/issues/176#issuecomment-343918731)

I was checking the code and thinking about the issue, it seems to be a jss issue instead of a react-jss issue. What do you think?

---

@kof commented on [Mon Nov 13 2017](https://github.com/cssinjs/react-jss/issues/176#issuecomment-343919101)

Well, ultimately yeah, as we are using property names in classes, they can also contain everything. We just never had that issue over there.

---

@kof commented on [Mon Nov 13 2017](https://github.com/cssinjs/react-jss/issues/176#issuecomment-343919693)

Also the escaping needs to be done in dev mode only, because in production we build short class names.

kof:
Fixes #624

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,7 +7,8 @@
   },
   "globals": {
     "benchmark": true,
-    "__VERSION__": true
+    "__VERSION__": true,
+    "CSS": true
   },
   "plugins": ["flowtype"]
 }
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+## 9.2.1 / 2017-11-13
+
+- Add CSS class name escaping for dev mode (#624)
+
 ## 9.2.0 / 2017-11-06
 
 - Allow empty rules when option {link: true} is used.
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -16,6 +16,13 @@ https://unpkg.com/jss/dist/jss.js
 Minified:
 https://unpkg.com/jss/dist/jss.min.js
 
+Polyfills:
+
+Only in development mode:
+
+[CSS.escape](https://github.com/mathiasbynens/CSS.escape)
+
+
 ## Setup with the default preset
 
 Use the [default preset](https://github.com/cssinjs/jss-preset-default) for a quick setup with recommended plugins.
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -18,6 +18,7 @@ module.exports = (config) => {
     files: [
       'node_modules/es5-shim/es5-shim.js',
       'node_modules/es5-shim/es5-sham.js',
+      'node_modules/css.escape/css.escape.js',
       'tests/index.js',
     ],
     preprocessors: {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2270,6 +2270,12 @@
         "is-in-browser": "1.1.3"
       }
     },
+    "css.escape": {
+      "version": "1.5.1",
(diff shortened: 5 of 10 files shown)


kof:
cc @oliviertassinari

oliviertassinari:
I'm assuming you have noticed the implementation we use on Material-UI side:
https://github.com/callemall/material-ui/blob/f10ff11979ac085bb8e2dcd740c8f3e8c57becf8/src/styles/createGenerateClassName.js#L51
I haven't heard any issue limitation from our users so far. CSS escape support is not quite high:
capture d ecran 2017-11-13 a 15 06 54
It might be a pain for people to install this polyfill when developing/fixing bugs on less common browsers.

kof:
This is a very outdated table!

kof:
I saw the regex in mui. It is a half backed escaping. It is broken already.

kof:
See a complete version https://github.com/mathiasbynens/CSS.escape

kof:
Chrome, Firefox and Safari have support already, we can expect a polyfill installed.

kof:
Actually I can add a warning in case it is missing!

kof:
I would not do it, but we need it in dev mode only, so it is kinda ok, right?

kof:
Also you don't know what user can put into the key name) it could be an emoji)

oliviertassinari:
@kof I don't know the probabilities. What's the probability of one using a not supported CSS selection char vs one trying out his interface in dev in a browser requiring the polyfill? Either way, one will discover this limitation.

kof:
Yes, but considered it is dev mode only, perf is not an issue, right?
In that case requiring polyfills is the standard way to go, right?

kof:
I am adding a good warning, which explains whats wrong, so user will have an easy time.

oliviertassinari:
perf is not an issue.

oliviertassinari:
Yeap, the resolution story with the polyfill will be better.

kof:
What happens actually if we write an invalid selector in a regular css file? The rule will be ignored, right?

oliviertassinari:
> What happens actually if we write an invalid selector in a regular css file? The rule will be ignored, right?

The user won't see his style. He will have to dig deep to understand what's going one.

kof:
Another option would be to just escape the component names on a different abstraction level and let user fix his escapings himself. Though as we are generating the class names and use rule keys sort of implicitly, it is probably a better DX to do what I do in this PR.

kof:
If we leave it up to user, it is like saying: "Dude, don't touch class names, we will generate them for you" and then "Dude, you need to make sure your property names are CSS conform."

oliviertassinari:
@kof Hard choice. Either way, we need to take care of it. You solution sounds good to me.

kof:
Just added a test where I use emoji as a key we got a new feature to tweet about haha!

PR #626, opened 2017-11-13
add __webpack_nonce__ support, fix #559

### LINKED ISSUE #559: Support CSP by adding __webpack_nonce__ to the style tag
This is really just a conceptual question to make sure that I understand the philosophy and core technology of the project.

From my cursory overview of the project, it appears to work by inserting and updating style elements into the DOM. This seems to me to be incompatible at a base level with [Content Security Policy security headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP), specifically the [style-src header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src).

For example, In order to get the react integration to work, the tightest security policy I am able to set is `default-src 'self'; style-src 'unsafe-inline' blob:`. For a CSP to provide realistic XSS protection, the `'unsafe-inline'` keyword cannot be included. Is this conflict an implementation detail that could be potentially mitigated, or am I correct in assessing that it's a root conceptual incompatibility?

andrepolischuk:


kof:
well done @andrepolischuk!

kof:
released

andrepolischuk:
cool

PR #627, opened 2017-11-13
fix incorrect expect usage

andrepolischuk:
expect receives only one argument https://github.com/Automattic/expect.js/blob/master/index.js#L30


kof:
thanks!

Issue #629, opened 2017-11-14
Allow function values to utilize the 'priority' parameter of the style.setProperty method and be able to set it to 'important'.


SimplerSoftware:
When using function values, I need to be able to use '!important'.  However, since function values are using the CSSOM 'setProperty' method, the value being set cannot have '!important', see [here](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty).  In order to be able to use '!important', the setProperty method must utilize the 3rd optional parameter and set it to 'important'.  

FYI - This is currently being called in the DomRenderer.js file, line 42.

Thanks!

kof:
good catch!

kof:
released in v9.4.0

PR #631, opened 2017-11-14
Add support for array values while using function values.

SimplerSoftware:
Resolves [#629](https://github.com/cssinjs/jss/issues/629)

### DIFF
--- a/docs/json-api.md
+++ b/docs/json-api.md
@@ -34,6 +34,16 @@ const styles = {
 }
 ```
 
+To use the '!important' modifier with function values, you must use [array syntax](./js-api.md#alternative-syntax-for-space-and-comma-separated-values):
+
+```
+const styles = {
+  button: {
+    color: data => [[data.color, '!important']]
+  }
+}
+```
+
 ## Function rules
 
 Similar to function values, you can use a function to return a dynamic style object. Use [sheet.update(data)](./js-api.md#update-function-values) in order to pass the data object. Sheet option `link: true` is required for this to function.
@@ -42,7 +52,8 @@ Similar to function values, you can use a function to return a dynamic style obj
 const styles = {
   button: (data) => ({
     position: 'flex',
-    color: data.color
+    color: data.color,
+	background: [[data.background, '!important']]
   })
 }
 ```
--- a/src/renderers/DomRenderer.js
+++ b/src/renderers/DomRenderer.js
@@ -37,6 +37,8 @@ function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string
         cssRule.style.setProperty(prop, cssValue, 'important')
         return true
       }
+      cssRule.style.setProperty(prop, cssValue)
+      return true
     }
     cssRule.style.setProperty(prop, value)
   }
--- a/src/renderers/DomRenderer.js
+++ b/src/renderers/DomRenderer.js
@@ -5,6 +5,7 @@ import type StyleSheet from '../StyleSheet'
 import StyleRule from '../rules/StyleRule'
 import type {Rule, InsertionPoint} from '../types'
 import global from '../utils/global'
+import toCssValue from '../utils/toCssValue'
 
 type PriorityOptions = {
   index: number,
@@ -29,9 +30,14 @@ function getStyle(cssRule: HTMLElement|CSSStyleRule, prop: string): string {
  */
 function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string): boolean {
   try {
-    if (value && value.toLocaleLowerCase().indexOf("!important") >= 0) {
-      value = value.substr(0, value.toLocaleLowerCase().indexOf("!important")).trim()
-      cssRule.style.setProperty(prop, value, "important")
+    if (Array.isArray(value)) {
+      let cssValue = toCssValue(value)
+      if (value[value.length - 1] === "!important") {
+        cssValue = cssValue.substr(0, cssValue.length - 10)
+        cssRule.style.setProperty(prop, cssValue, "important")
+      }
+      else
+        cssRule.style.setProperty(prop, cssValue)
     }
     else
       cssRule.style.setProperty(prop, value)
--- a/src/renderers/DomRenderer.js
+++ b/src/renderers/DomRenderer.js
@@ -32,15 +32,17 @@ function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string
   try {
     if (Array.isArray(value)) {
       let cssValue = toCssValue(value)
-      if (value[value.length - 1] === "!important") {
+      if (Array.isArray(value[value.length - 1]) && value[value.length - 1][value[value.length - 1].length - 1] === '!important') {
         cssValue = cssValue.substr(0, cssValue.length - 10)
-        cssRule.style.setProperty(prop, cssValue, "important")
+        cssRule.style.setProperty(prop, cssValue, 'important')
       }
-      else
+      else {
         cssRule.style.setProperty(prop, cssValue)
+      }
     }
-    else
+    else {
       cssRule.style.setProperty(prop, value)
+    }
   }
   catch (err) {
     // IE may throw if property is unknown.
--- a/src/renderers/DomRenderer.js
+++ b/src/renderers/DomRenderer.js
@@ -35,14 +35,10 @@ function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string
       if (Array.isArray(value[value.length - 1]) && value[value.length - 1][value[value.length - 1].length - 1] === '!important') {
         cssValue = cssValue.substr(0, cssValue.length - 10)
         cssRule.style.setProperty(prop, cssValue, 'important')
-      }
-      else {
-        cssRule.style.setProperty(prop, cssValue)
+        return true
       }
     }
-    else {
-      cssRule.style.setProperty(prop, value)
-    }
+    cssRule.style.setProperty(prop, value)
   }
   catch (err) {
     // IE may throw if property is unknown.
--- a/src/renderers/DomRenderer.js
+++ b/src/renderers/DomRenderer.js
@@ -29,7 +29,12 @@ function getStyle(cssRule: HTMLElement|CSSStyleRule, prop: string): string {
  */
 function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string): boolean {
   try {
-    cssRule.style.setProperty(prop, value)
+    if (value && value.toLocaleLowerCase().indexOf("!important") >= 0) {
+      value = value.substr(0, value.toLocaleLowerCase().indexOf("!important")).trim()
+      cssRule.style.setProperty(prop, value, "important")
+    }
+    else
+      cssRule.style.setProperty(prop, value)
   }
   catch (err) {
     // IE may throw if property is unknown.
--- a/tests/functional/function-values.js
+++ b/tests/functional/function-values.js
@@ -122,6 +122,53 @@ describe('Functional: Function values', () => {
     })
   })
 
+  describe('.addRule() with arrays as function values', () => {
+    let style
+    let sheet
+
+    beforeEach(() => {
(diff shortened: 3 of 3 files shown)


src/renderers/DomRenderer.js
```diff
@@ -29,7 +29,12 @@ function getStyle(cssRule: HTMLElement|CSSStyleRule, prop: string): string {
  */
 function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string): boolean {
   try {
-    cssRule.style.setProperty(prop, value)
+    if (value && value.toLocaleLowerCase().indexOf("!important") >= 0) {
```
kof: My problem with this approach is that it will make every update slower, even if it doesn't has an "important". This part of the API is hot, because is used for animations.

kof: Second point is that important in general is a bad thing, it is not needed in a clean design, just a hack.

kof: Potentially we could support something like this: `color: ['red', '!important']` so an important value needs to be written separately by the user and so we don't have to parse it.

SimplerSoftware: First off... wanna say thanks for this awesome library!!!  

While there is definitely some performance hit with the approach, I don't feel that this is or will be the bottleneck of an application when it comes to performance.  Having said that, I don't write graphic intensive applications in Javascript that require thousands of style changes per second.  However, I did setup a test comparison just to show what the impact looks like so that we can talk about it constructively.  The tests can be found [here](https://jsperf.com/jsstolowercasetest/1).  Here's a screenshot as well of a test that I ran:

image

Certainly the test called 'No Check' is the fastest (as expected).  However, it's processing over 3M operations per second!!  The test called 'With check and toLowerCase' which is representative of this PR, is ~41% slower, but is still processing over 1.8M operations per second!!  You'd have to be doing a TON of changes to even get close with this being the bottleneck of the application.  The 'As Array' test I think is representative of what you were proposing above... and is certainly faster than the other  approach, but comes with the cost of not being as straight forward to consume.  If we don't do a check and "assume" that '!important' will always be lowercase already then the 'with check' is just as performant as the array approach.

Anyway, long story short, I feel that even though there is a performance hit with this approach, it is the most straight forward to consume (no changes necessary from a consumer standpoint) and is the way that I think everyone would *expect* to happen.  I appreciate the consideration.

kof: I am also thinking about other plugins and pipelines. Ideally if we do the conversion from string, we should make it in a way that it is done only once and other plugins can rely on it.

I am thinking that we could support !important conversion to an array as a plugin, so that once it is converted all other plugins and DOM module can rely on arrays.

kof: For e.g. we have the same problem in vendor prefixer.

kof: What bothers me and I have no solution for yet:

Function values/rules are in JSS high performance, they are capable of being updated at 60fps. Compared to other libraries, they allow only regular one-time updates, everything else will be generating tons of css.

At the same time they are used for one-time updates in jss as well, for e.g. for theming etc.

So we have 2 different usages of this interface, one is very intensive the other one is not so much.

While for regular "theming" usage, such overhead as important parsing or other plugins like verndor prefixer doesn't matter, for high frequency updates it matters.

So I am still not sure how to proceed with them. Evtl. we need an  API change that allows to distinguish those cases and allow optimizing the frequent updates.

kof: Also in general I need to decide whether we want to support "!important"  as a mainstream feature. I am not convinced it should be, because there should be no need, but if there is, maybe an optional plugin should do the job?

kof: What do you think about supporting array syntax in the core, in setStyle function now. And implement !important   parsing plugin separately?

kof: Also, for space separated values core supports double arrays syntax, because single array is for comma separated values, see[

SimplerSoftware: I think what you propose with the array syntax makes sense.  I guess I didn't realize that having values as an array was already supported in the core.  And to your point about performance, having a separate plugin makes sense as well since users can choose when to use it and when not to.

kof: yep!

kof: would you like to send a PR for arrays syntax support?

SimplerSoftware: Sure... are you looking for something like:
```
    if (Array.isArray(value) && value[1] === "!important")
      cssRule.style.setProperty(prop, value[0], "important")
    else
      cssRule.style.setProperty(prop, value)
```
My issue with this is that it's not truly supporting array syntax.  Array syntax can be used in many other situations and so it would need to be something more like:
```
    if (Array.isArray(value)) {
      let cssValue = utils.toCssValue(value)
      if (value[value.length] === "!important") {
        cssValue = cssValue.substr(0, cssValue.indexOf("!important"))
        cssRule.style.setProperty(prop, cssValue, "important")
      }
      else
        cssRule.style.setProperty(prop, cssValue)
    }
    else
      cssRule.style.setProperty(prop, value)
```
This seems a lot more verbose and less performant but would handle all the different scenarios where an array might  be used as a value.  Although it would really only be less performant for values that are an array and also even less performant for those that use "!important".  I think it's an acceptable trade-off.  Thoughts?

kof: The first version, but checking for the last element in the array instead.

kof: actually wait, you are right, there is more to this

kof: Ok, we def. need to use toCssValue here. I am thinking of 2 different implementations which can have different perf characteristics:

1. always return an array at the end out of toCssValue and relay on Array#toString.
2. create a class CSSValue and return it. It's toString would to what toCssValue does now + you can check for important flag on the instance

kof: On the other hand, you don't need indexOf call, you already know when it is at the end, you can just do substr.

SimplerSoftware: So something like:
```
    if (Array.isArray(value)) {
      let cssValue = utils.toCssValue(value)
      if (value[value.length] === "!important") {
        cssValue = cssValue.substr(0, cssValue.length - 10)
        cssRule.style.setProperty(prop, cssValue, "important")
      }
      else
        cssRule.style.setProperty(prop, cssValue)
    }
    else
      cssRule.style.setProperty(prop, value)
```

kof: yeah, looks correct and introduces only overhead if there is array at first and then !important at second. Also it supports now a full array syntax, which was basically broken otherwise.

SimplerSoftware:
Updated the DomRenderer.js file per our latest conversation.

kof:
Do you care to write tests and lint code as well?

kof:
Also it needs docs

SimplerSoftware:
I can take a stab at the tests and documentation but I'm afraid I don't know the first thing about linting.

kof:
just run `npm run lint`

SimplerSoftware:
I can definitely do that ;)

SimplerSoftware:
I'm getting 5918 errors, all of which are the type: "error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style".  None of them are in files that I've touched though.  Can I just ignore all of those?

kof:
Not getting this. Are you using npm 5? Otherwise it will ignore lock file.

SimplerSoftware:
NPM v5.5.1 - I'm using Visual Studio and not VS Code though.  Not sure if that has anything to do with it.

kof:
Why should a shell command have anything to do with editor?

SimplerSoftware:
Agreed.  Just thought I'd throw that out there just in case.

SimplerSoftware:
The errors appear to be caused by Windows.  I've added the following line to the .eslintrc file in the rules section: "linebreak-style": 0

This fixes the error on Windows but will it now break on other OS's?

SimplerSoftware:
@kof Added the tests and made sure the code passed the linting process.  The logic needed to be updated slightly in order to handle the array structure since it wouldn't be just an array but rather an array or arrays in order to handle '!important'.  I didn't commit the change to the .eslintrc file since I wasn't sure how that might affect other OS's.  Also, I didn't see a good place to update the documentation.  If you can point me where you want me to put it I could write something up for you to review.

kof:
line-break style should not be 0, I guess you can setup your editor to use LF?

src/renderers/DomRenderer.js
```diff
@@ -29,7 +30,19 @@ function getStyle(cssRule: HTMLElement|CSSStyleRule, prop: string): string {
...
 function setStyle(cssRule: HTMLElement|CSSStyleRule, prop: string, value: string): boolean {
   try {
-    cssRule.style.setProperty(prop, value)
+    if (Array.isArray(value)) {
+      let cssValue = toCssValue(value)
+      if (Array.isArray(value[value.length - 1]) && value[value.length - 1][value[value.length - 1].length - 1] === '!important') {
+        cssValue = cssValue.substr(0, cssValue.length - 10)
+        cssRule.style.setProperty(prop, cssValue, 'important')
+      }
+      else {
```
kof: we use early returns to avoid too much nesting and "else"

kof:
Regarding the docs - I think a good place would be to point to the arrays format from function values and function rules in json-api.md

SimplerSoftware:
@kof Updated per your feedback.  Let me know if anything else needs to be changed.

docs/json-api.md
```diff
@@ -42,7 +52,8 @@ Similar to function values, you can use a function to return a dynamic style obj
 const styles = {
   button: (data) => ({
     position: 'flex',
-    color: data.color
+    color: data.color,
+	background: [[data.background, '!important']]
```
kof: is that a tab?

docs/json-api.md
```diff
@@ -34,6 +34,16 @@ const styles = {
 }
 ```
 
+To use the '!important' modifier with function values, you must use [array syntax](./js-api.md#alternative-syntax-for-space-and-comma-separated-values):
+
+```
+const styles = {
+  button: {
+    color: data => [[data.color, '!important']]
```
kof: I am asking myself what more correct:

`[[data.color, '!important']]` vs. `[data.color, '!important']` and `[[data.color],  '!important']`

kof: On one hand, space separated values are separated using array syntax using double arrays. On the other hand !important is not really part of the value itself, it is more like a flag.

WDYT?

SimplerSoftware: I like the idea of it just being a single array: [data.color, '!important'], and that's how I originally had the code structured, but it adds the comma in the text when calling the 'toCssValue' method.  I guess since we're already doing a substring to remove the '!important' from the text, we could just make it remove the comma in front of it as well if we wanted to keep it as a single array. Thoughts?

kof: We could teach toCssValue handle last element properly for e.g., not sure if it is better.

SimplerSoftware: I guess I would lean on your knowledge to decide what is best for the overall architecture. Either way I'm hoping we can release these soon :).  Let me know if you'd like me to change anything.

kof: The thing is if we don't teach toCssValue to correctly stringify a value with !important, the function toCssValue itself won't be correct, because it will convert  `toCssValue(['red', '!important'])` to `'red, !important'`.  So I think we should change it.

SimplerSoftware: That being said, I think perhaps to keep things simple and not add overhead to the 'toCssValue' function we should just leave it as it stands using the double array syntax:  [['red', '!important']].  Modifying the 'toCssValue' method just to add logic for the "!important" modifier might degrade performance.  I'm not sure where and how often it's called and so I would leave that to your judgement. Thoughts?

kof: Need to check what needs to be done in toCssValue for this

kof:
Ok, I digged into it and updated your commits, lets review this pr and release it. https://github.com/cssinjs/jss/pull/646

PR #632, opened 2017-11-15
Use a fallback regex instead of warning

kof:
we don't want to force people to install the polyfil, because it is a
rare one right now and it causes setup overhead

### DIFF
--- a/src/utils/escape.js
+++ b/src/utils/escape.js
@@ -5,18 +5,15 @@ const CSS = (global.CSS: any)
 
 const env = process.env.NODE_ENV
 
+const escapeRegex = /([[\].#*$><+~=|^:(),"'`])/g
+
 export default (str) => {
   // We don't need to escape it in production, because we are not using user's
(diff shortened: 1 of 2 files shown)


src/utils/escape.js
```diff
@@ -5,18 +5,15 @@ const CSS = (global.CSS: any)
...
   if (env === 'production') return str
 
   if (!CSS || !CSS.escape) {
-    warning(
-      false,
-      '[JSS] CSS.escape polyfill in DEV mode is required in this browser, ' +
-      'check out https://github.com/mathiasbynens/CSS.escape'
-    )
-    return str
+    return str.replace(escapeRegex, '\\$1')
```
oliviertassinari: I haven't tested the `'\\$1'` solution. I trust you on this.

kof: I have a trivial test, which shows that it works for "()", I don't test each character though.

kof: That test compares it with CSS.escape.

PR #634, opened 2017-11-16
add `onUpdate` hook to docs

andrepolischuk:


docs/plugins.md
```diff
@@ -92,3 +92,15 @@ You need to register a `plugin` only once per JSS instance. There is a number of
       }
     })
     ```
+
+1. Hook `onUpdate(data, rule, sheet)`.
```
kof: I was keeping it undocumented, because it was used only internally, but yeah, once we start using it outside, we should document it.

kof:
thanks!

PR #641, opened 2017-12-05
Function values docs example

akryvomaz:
It was not clear to me what exactly should be in the update function param and how does it binds to the style rules. I think this example should help.


kof:
lgtm, thanks

PR #643, opened 2017-12-06
Use prefix in className on production

andrepolischuk:


src/utils/createGenerateClassName.js
```diff
@@ -32,11 +32,12 @@ export default (): generateClassName => {
       )
     }
 
+    const prefix = sheet ? (sheet.options.classNamePrefix || '') : ''
```
kof: We could put 'c' as a default prefix if user hasn't provided one

andrepolischuk: We can. I thought, `c` is analogue to `rule.key` in development, but not prefix.

kof: We added "c" because class names can't start with some characters like number, to ensure that. Should have added a comment. In case where user has defined a prefix, I think we can rely on it.

kof:
merged, thanks!

Issue #644, opened 2017-12-06
jss.createStyleSheet() calls collide with independent jss instances


appsforartists:
The docs currently encourage each component to have its own instance of JSS.  However, the instances are independent, which can lead to bugs:

```ts
// component1
const jss = createJSS().setup(preset());
const styleSheet = jss.createStyleSheet({
  scrim: {
    backgroundColor: 'red',
  //...
});

// component2
const jss = createJSS().setup(preset());
const styleSheet = jss.createStyleSheet({
  scrim: {
    backgroundColor: 'blue',
  //...
});
```

It looks like `component1` will have a red scrim and `component2` will have a blue one.  Unfortunately, because they are independent instances, each will generate the same `className` (e.g. `scrim-0-1`), so the two style declarations will conflict with one another.

## Potential solutions
1. Update the documentation to note this issue.  This would lead encourage authors to use the global instance, but could make npm-distributed components less encapsulated.
2. Maintain a `className` registry in jss, independent of the `jss` instance.  This would allow independently-configured `jss` instances to ensure they aren't creating colliding classes.  However, it could complicate garbage collection in server-side rendering.
3. Move `setup` to the `styleSheet` level, rather than the instance level.  This would allow each rendered page to use a single `jss` instance, even if an imported component depends on a different setup.
4. Add a random (hopefully unique) key to each `className`.  Then, the example would result in `scrim1209312031` and `scrim09834234`, which would not conflict.

4 is probably the best trade off between simplicity, minimizing changes, and fixing the issue.

kof:
Omg, no, jss instance should be per app or basically per configuration. Where should we fix it?

appsforartists:
I think you need to add a random hash to each class name.  Otherwise, you get the same problem with

```js
// some-npm-component

const jss = createJSS().setup(myConfig);
const styleSheet = jss.createStyleSheet({
  scrim: // ...
});

// app

import { Component } from 'some-npm-component';

const jss = createJSS().setup(preset());
const styleSheet = jss.createStyleSheet({
  scrim: // ...
});
```

kof:
We have classNamePrefix option for this, as I assume there should not be hundreds of jss instances,  a few of them can be manually prefixed using that option.

kof:
Most of the time you actually have just once instance, second instance may be needed if you have a part of an application which comes from a different package and uses a different jss setup for e.g.

kof:
Where can we enhance the docs for this?

[...]

PR #648, opened 2017-12-11
Use Jss instance counter  in class names generator #644

kof:


src/Jss.js
```diff
@@ -27,7 +27,11 @@ declare var __VERSION__: string
 
 const defaultPlugins = rulesPlugins.concat([observablesPlugin, functionsPlugin])
 
+let instanceCounter = 0
+
 export default class Jss {
+  id = instanceCounter++
```
oliviertassinari: Is it equivalent to having it in the constructor?

src/utils/createGenerateClassName.js
```diff
@@ -33,12 +28,18 @@ export default (): generateClassName => {
...
+      if (sheet.options.jss.id != null) jssId += sheet.options.jss.id
+    }
 
     if (env === 'production') {
-      return `${prefix}${jssCounter}${ruleCounter}`
+      return `${prefix}${moduleId}${jssId}${ruleCounter}`
     }
 
-    return `${prefix + rule.key}-${jssCounter}-${ruleCounter}`
+    return `${prefix + rule.key}-${moduleId}${jssId && `-${jssId}`}-${ruleCounter}`
```
oliviertassinari: Can `jssId` be falsy?

src/utils/createGenerateClassName.js
```diff
@@ -2,12 +2,7 @@
 import warning from 'warning'
 import type {Rule, generateClassName} from '../types'
 import StyleSheet from '../StyleSheet'
-import global from './global'
-
-const ns = '2f1acc6c3a606b082e5eef5e54414ffb'
-if (global[ns] == null) global[ns] = 0
-// In case we have more than one JSS version.
-const jssCounter = global[ns]++
+import moduleId from './moduleId'
```
oliviertassinari: What's the use case for the moduleId? Isn't jssId replacing it?

src/Jss.js
kof: yes, its new property initializers syntax in es-something

src/utils/createGenerateClassName.js
kof: no, we need both, added a better comment about it https://github.com/cssinjs/jss/pull/648/commits/88ffe5e688a61e633a3a70ebc47257a89f15bca9

src/utils/createGenerateClassName.js
kof: excluded in the line above

src/utils/moduleId.js
```diff
@@ -3,5 +3,8 @@ import global from './global'
 const ns = '2f1acc6c3a606b082e5eef5e54414ffb'
 if (global[ns] == null) global[ns] = 0
 
-// In case we have more than one module version.
+// Bundle may contain multiple JSS versions at the same time. In order to identify
```
oliviertassinari: I still don't get it. How is the bundle containing multiple JSS versions on issue? Doesn't the issue coming from the different JSS instance?

kof: If bundle contains multiple JSS package versions, modules are evaluated separately, which means JSS instance counter will start from 0 for each version.

kof: I could replace the module id with a JSS instance counter which always starts using a global counter, like I do with moduleId right now, so there will be no need for module id. I thought this implementation will be easier to reason about.

oliviertassinari:
I came to realize the implication of this PR. It means that people will be able to have multiple version of JSS in their bundle. Effectively, shipping two times the same large CSS-in-JSS runtime to their users. Shouldn't we discourage it by not making it work and by having jss as a peer dependency instead?

kof:
We had support for multiple JSS versions before that as well.

Some time ago we introduced rules counter reset through createGenerateClassName option and this allowed user to have 2 Jss instances with the rules counter starting at 0 and clashing (same JSS module).

So now we just fixed this by adding also Jss instance counter, so that user can have many Jss instances and reset the rules counter whenever they want without clashes.


That being said, we don't have other choice rather than support multiple versions, because user might need this in some situation for migration  purposes. We could add a warning though for the case user is not aware, but we should not forbid this. 

I am happy to add a warning for dev.

PR #655, opened 2017-12-22
Fix added nested rules order

lttb:
Fix an issue from https://github.com/cssinjs/styled-jss/pull/59

The problem is when we add rules by `.addRules`, it has some issues with nested rules order.

```js
const Div = styled('div')({
  padding: 15,
  '&:hover': {
    '& $button': {
      color: ({primary}) => (primary ? 'red' : 'green'),
    },
  },
})

const Button = styled('button')({
  composes: '$button',
})

const App = (props: {primary?: boolean}) => (
  <div>
    <Div primary={props.primary}>
      <Button>Button</Button>
    </Div>
  </div>
)

expect(getCss(sheet)).toBe(removeWhitespace(sheet.toString()))
```

image

### DIFF
--- a/src/StyleSheet.js
+++ b/src/StyleSheet.js
@@ -89,7 +89,7 @@ export default class StyleSheet {
       if (!this.deployed) return rule
       // Don't insert rule directly if there is no stringified version yet.
       // It will be inserted all together when .attach is called.
-      if (queue) queue.push(rule)
+      if (queue) queue.unshift(rule)
       else {
         this.insertRule(rule)
         if (this.queue) {
--- a/tests/integration/plugins.js
+++ b/tests/integration/plugins.js
@@ -1,9 +1,14 @@
 import expect from 'expect.js'
 import {stripIndent} from 'common-tags'
+import jssNested from 'jss-nested'
 import {create} from '../../src'
 import StyleSheet from '../../src/StyleSheet'
-import {createGenerateClassName} from '../utils'
 import PluginsRegistry from '../../src/PluginsRegistry'
+import {
+  createGenerateClassName,
+  getCssFromSheet,
+  removeWhitespace
+} from '../utils'
 
 describe('Integration: plugins', () => {
   let jss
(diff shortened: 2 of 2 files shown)


tests/integration/plugins.js
```diff
@@ -416,4 +421,29 @@ describe('Integration: plugins', () => {
       `)
     })
   })
+
+  describe('jss-nested', () => {
```
kof: this test actually belongs to jss-nested

kof: alternatively we could describe the same scenario using plugins api instead of jss-nested plugin

tests/integration/plugins.js
```diff
@@ -416,4 +421,29 @@ describe('Integration: plugins', () => {
...
+        '&:hover': {
+          '& $b': {
+            color: 'red',
+          },
+        },
+      })
+    })
+
+    it('should save the added nested rules order', () => {
+      expect(getCssFromSheet(sheet)).to.be(removeWhitespace(sheet.toString()))
```
kof: we only use DOM based tests in functional tests, is DOM really needed here or sheet.toString() + comparing with result would be enough?

tests/integration/plugins.js
lttb: I think that this test should be here because it tests the way how rules apply in some cases, so the same scenario instead of jss-nested looks better for me

src/StyleSheet.js
```diff
@@ -89,7 +89,7 @@ export default class StyleSheet {
       if (!this.deployed) return rule
       // Don't insert rule directly if there is no stringified version yet.
       // It will be inserted all together when .attach is called.
-      if (queue) queue.push(rule)
+      if (queue) queue.unshift(rule)
```
kof: I am not sure why this is not breaking anything else. You are reverting the order of added rules for all addRule calls. Maybe there is a missing test which ensures the right order.

We have actually an `options.index` which is set in jss-nested in order to have the control over the order. Maybe jss-nested sets the wrong index?

tests/integration/plugins.js
lttb: yeah, I think we can replace expect with the exact result

kof:
totally forgot this one

kof:
@lttb whats the state here?

PR #666, opened 2018-01-19
fix link to work with global rule

### LINKED ISSUE #664: [jss-global] Link option does not work
I'm unable to use the `link` option mentioned in the JS api documentation with jss-global plugin. Every call to `sheet.update` after `sheet.attach` have been called triggers a

> Warning: Rule is not linked. Missing sheet option "link: true".

even with the right option, and sheet doesn't get refreshed.

I have been able to reproduce this behavior here: https://codesandbox.io/s/vml5lxmvy 

Could be the same as #602 but op didn't said about the `link` option usage.

sambhav-gore:
Related to #664 

As I was also trying to get it work I fixed this.
This works for me and all tests pass, but please check if this causes any side effects.

May be this should be moved to the jss-global plugin ? ( I am not sure about it), I had to use the 'any' type as the actual type of the rule is 'GlobalContainerRule' which is part of jss-global. Let me know if I should improve this PR or you can adapt it.

### DIFF
--- a/src/RuleList.js
+++ b/src/RuleList.js
@@ -155,7 +155,9 @@ export default class RuleList {
       const cssRule = cssRules[i]
       let key = this.options.sheet.renderer.getKey(cssRule)
       if (map[key]) key = map[key]
-      const rule = this.map[key]
+
+      const mappedGlobal: any = this.map['@global']
+      const rule = mappedGlobal ? mappedGlobal.rules.map[key] || this.map[key] : this.map[key]
       if (rule) linkRule(rule, cssRule)
     }
   }


src/RuleList.js
```diff
@@ -155,7 +155,9 @@ export default class RuleList {
       const cssRule = cssRules[i]
       let key = this.options.sheet.renderer.getKey(cssRule)
       if (map[key]) key = map[key]
-      const rule = this.map[key]
+
+      const mappedGlobal: any = this.map['@global']
```
kof: core doesn't know anything about plugins, it needs to be solved locally by a plugin, if plugin is missing what it needs, core should provide a generic solution so that plugin can fix it

sambhav-gore: @kof Thats what I was thinking that this should be moved to the plugin itself. On this weekend if I get some time I will try to move it there. as of right now I am not sure how the plugin should handle this.

aslafy-z:
@sambhav-gore any news on this ?

src/RuleList.js
kof: hey @sambhav-gore I see you got distracted, any chance you can finish this one?

HenriBeck:
Are there any updates on this? @sambhav-gore

aslafy-z:
@kof is this still relevant? We need this to work for a long time now, should we switch to an alternative?

kof:
linking logic has been largely refactored, I need to check if there is still an issue in the next version. note master is in the preparation for v10

HenriBeck:
@kof let's maybe close this pr and add a test for it maybe?

kof:
Lets add the tests first to see if the problem still exists.

HenriBeck:
Seems we already have tests and they are passing: https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-syntax-rule-value-function/src/plugin-global.test.js

kof:
seems like that, yes, @sambhav-gore wanna try the new version directly from master and see if your use case is fixed, since you didn't provide a test, we can't be 100% sure, but it seems like we have this covered and it works.

HenriBeck:
Closing as this seems to be fixed now.

PR #667, opened 2018-01-21
Csp setup

eaardal:
Documentation for configuring CSP, as discussed here https://github.com/cssinjs/jss/issues/559

Please let me know if anything is unclear or misspelled 👍


docs/csp.md
```diff
@@ -0,0 +1,118 @@
+# Setting up Content Security Policy with JSS
+
+Content Security Policy (CSP) is a way of whitelisting what resources the browser should allow to load (and reach out to). There are many good resouces on CSP for more information. [This explanation](https://helmetjs.github.io/docs/csp/) is a good place to start.
+
+In the case of JSS we need to set the `style-src` CSP directive. We don't want to just set it to `unsafe-inline` which will allow everything.
```
kof: directive => attribute?

kof: ah right, directive

docs/csp.md
```diff
@@ -0,0 +1,118 @@
...
+Set the `global.__webpack_nonce__` variable to the same nonce value:
+
+```js
+// webpack.config.js
+const WEBPACK_NONCE = 'N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3'
+module.exports = {
+  /* ... */
+  plugins: {
+    new webpack.DefinePlugin({
+      'global.__webpack_nonce__': JSON.stringify(WEBPACK_NONCE),
```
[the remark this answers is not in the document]
kof: oh right, it adds additional quotes

docs/csp.md
```diff
@@ -0,0 +1,118 @@
...
+    }),
+  },
+}
+```
+
+(`__webpack_nonce__` is a poorly documented variable which, when set, jss will read and apply to `<style />` elements: `<style nonce={webpack-nonce-value} />`).
```
kof: we write jss always capital "JSS"

docs/csp.md
```diff
@@ -0,0 +1,118 @@
...
+
+At this point, the style blocks generated by jss should render with the nonce value (`<style nonce>`) when you inspect the page.
+
+![stylenonce](https://user-images.githubusercontent.com/2316274/35198155-e9472bc6-feea-11e7-9ef8-eac110590c59.PNG)
```
kof: I think we don't really need the screenshot here, otherwise I would place it right near the md file, so that we have a guarantee noone deletes the image, but I think its really just illustrating an attribute on a style tag, we can omit it.

docs/csp.md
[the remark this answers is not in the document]
eaardal: Yeah it's ugly, but the DefinePlugin is a weird beast. From their Usage section:

> - If the value is a string it will be used as a code fragment.
> - If the value isn't a string, it will be stringified (including functions).

They also use `JSON.stringify` in their examples: https://webpack.js.org/plugins/define-plugin/

kof:
Thank you, merged

PR #669, opened 2018-01-23
Secure way to use a CSP

Graham42:
Fixes: #559

This removes the ability to use the webpack value `__webpack_nonce__` as it was insecure, and provides a more secure way to use a CSP.

See #559 for discussion.


Graham42:
I'm getting a strange error in the travis build
```
23 01 2018 18:07:59.872:ERROR [launcher]: Cannot load browser "BS_Safari"!
  Error: Password is required.
```

kof:
Thats normal, I still havn't found any solution to run tests on browserstack from external contributors.

src/renderers/DomRenderer.js
```diff
@@ -248,6 +248,23 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
   getHead().insertBefore(style, prevNode)
 }
 
+/**
+ * Read jss settings from the page if the user has set them.
+ */
+function getSettingsFromInlineJSON() : Object {
+  let config = {}
+  const inlineJsonBlock = document.querySelector('script[type="application/json"][data-jss-selector="jss-settings-json"]')
```
kof: this query can be simplified to an id or simple class name, right?

docs/csp.md
```diff
@@ -41,45 +50,58 @@ app.use(
...
-2. In webpack config:
+1. Now we want to include this value in our page so that JSS can pick it up at runtime.
+
+You can use any templating engine, or if you have SSR that should work too. For this example we'll use the [Nunjucks](https://github.com/mozilla/nunjucks) template engine.
+
+1. First create the template html file.
+
+```html
+<head>
+  <script data-jss-selector="jss-settings-json" type="application/json">
```
kof: type is optional

kof: ah, you are using json in here .... I think in that case browser just leaves it as text, its not a type browser will do something with

src/renderers/DomRenderer.js
```diff
@@ -248,6 +248,23 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
...
+  if (inlineJsonBlock) {
+    try {
+      config = JSON.parse(inlineJsonBlock.textContent)
+    }
+    catch (error) {
+      warning('Error parsing jss-settings-json', error)
```
kof: where is this warning from? mb for the sake of example console.warn?

docs/csp.md
Graham42: Type is important. If there's no type it will default to `"text/javascript"` which will be blocked by a CSP that prevents inline scripts.

kof: actually we can simplify this thing to just use a global variable ...

src/renderers/DomRenderer.js
Graham42: could simplify to `#jss-settings-json` if prefered?

docs/csp.md
kof: or would this be a leak if we just declare global variable with the nonce value?

src/renderers/DomRenderer.js
Graham42: I was going to use `console` but the linter complained about that, so I looked at the other code in the file and copied what I saw at https://github.com/cssinjs/jss/blob/master/src/renderers/DomRenderer.js#L244

docs/csp.md
Graham42: Which I suppose I should add a comment about.

Graham42: This _is_ possible to do with global value, as dav-is suggested here: https://github.com/cssinjs/jss/issues/559#issuecomment-359685112 
However I was wanting avoid any recommendation of inline javascript, as a matter of preference, as I mentioned here:
https://github.com/cssinjs/jss/issues/559#issuecomment-359800360

kof: what about some custom meta tag?

Graham42: We could use an arbitrary tag with a `data-` attribute. 
Example
```html
<span id="jss-config-nonce" data-nonce="NTBkM2U4YmUtYTA5NC00YTBlLTg3NTgtZmZhZDA2ODgxMWVm"></span>
```

Graham42: Could use `<meta ... />` in the head if that's what you meant.

kof: sure, I was just thinking what is nicer, meta tag seems nice, also in both cases no JSON.parse, also I think it should be up to user where he graps the nonce text from, and passes it as option, so that we don't have to deal with transport in the core.

Graham42: True... I could go back to using the global variable and document these different ways of setting the nonce. And then as users to pass it to the global variable.

kof: for e.g. <meta property="nonce" content="asdfasdfaasdf" />

kof: I think meta tag is nicer than global variable

kof: also it is more generic, might be used by some other parts of the application as well

Graham42: Ok, I'll refactor to use a `<meta />` tag

Graham42:
Updated to use meta tag.

docs/csp.md
```diff
@@ -1,36 +1,45 @@
...
 ```js
 // server.js
 import helmet from "helmet";
+import uuidv4 from "uuid/v4"
 
-const WEBPACK_NONCE = "N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3";
-
+app.use(function(req, res, next) {
+  // nonce should be base64 encoded
+  res.locals.nonce = btoa(uuidv4());
```
dav-is: `btoa()` is defined on `window`. We can't use it in the server. Also, we should use the variable `styleNonce` so that one doesn't mistakenly use the style nonce for `script-src` and `style-src`. This nonce should be explicitly used for styles only.

```
res.locals.styleNonce = new Buffer(uuidv4()).toString('base64');
```

kof:
Why not `<meta property="nonce" content="asdfasdfaasdf" />`, isn't it nice?

docs/csp.md
```diff
@@ -1,36 +1,45 @@
...
 app.use(
   helmet.contentSecurityPolicy({
     directives: {
+      defaultSrc: ["'self'"],
       /* ... */
-      styleSrc: ["'self'", `'nonce-${WEBPACK_NONCE}'`]
+      styleSrc: [
+        "'self'",
+        function(req, res) {
+          return "'nonce-" + res.locals.nonce + "'";
```
dav-is: use:
```
return "'nonce-" + res.locals.styleNonce + "'";
```

docs/csp.md
```diff
@@ -1,36 +1,45 @@
...
 
 > 'nonce-{base64-value}'
 > A whitelist for specific inline scripts using a cryptographic nonce (number used once). **The server must generate a unique nonce value each time it transmits a policy. It is critical to provide an unguessable nonce, as bypassing a resource’s policy is otherwise trivial**. See [unsafe inline](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#Unsafe_inline_script) script for an example.
 
-A conundrum for now is how to generate a unique nonce value for every page visit, as webpack must know the nonce at bundle time, while the server must know the nonce value at runtime for including it in the CSP rules. For this tutorial, we'll use a hard-coded Base64-encoded GUID value.
-
-1. Generate a Base64 guid:
+Because the nonce must be generated to be unique and random for every request, this is not something that we can do at compile time. Previously the docs suggested using the `__webpack_nonce__` variable with Webpack. However that is insecure because it never changes, so it could be trivially bypassed by an attacker, as noted in the Mozilla docs above.
```
kof: we can do at ~compile-time~ build time?

docs/csp.md
```diff
@@ -41,45 +50,55 @@ app.use(
...
+1. First create the template html file.
+
+```html
+<head>
+  <meta id="jss-nonce" data-value="{{ noncePlaceholder }}">
+</head>
+...
+```
+
+1. Now update the server to render the template with the `noncePlaceholder` variable to be interpolated with the nonce generated from our middleware `res.locals.nonce`.
```
dav-is: res.locals.styleNonce

docs/csp.md
```diff
@@ -41,45 +50,55 @@ app.use(
...
+1. Now we want to include this value in our page so that JSS can pick it up at runtime.
+
+You can use any templating engine, or if you have SSR that should work too. For this example we'll use the [Nunjucks](https://github.com/mozilla/nunjucks) template engine.
 
-Set the `global.__webpack_nonce__` variable to the same nonce value:
+1. First create the template html file.
+
+```html
+<head>
+  <meta id="jss-nonce" data-value="{{ noncePlaceholder }}">
```
dav-is: Maybe a more universal name is appropriate? Also `noncePlaceholder` should be `styleNoncePlaceholder` or `styleNonce`.

docs/csp.md
```diff
@@ -1,36 +1,45 @@
...
 import helmet from "helmet";
+import uuidv4 from "uuid/v4"
 
-const WEBPACK_NONCE = "N2M0MDhkN2EtMmRkYi00MTExLWFhM2YtNDhkNTc4NGJhMjA3";
-
+app.use(function(req, res, next) {
+  // nonce should be base64 encoded
+  res.locals.nonce = btoa(uuidv4());
+  next();
+});
```
kof: lets stick with no-semicolon style) no hard feelings, just consistency

docs/csp.md
```diff
@@ -41,45 +50,55 @@ app.use(
...
-    new webpack.DefinePlugin({
-      'global.__webpack_nonce__': JSON.stringify(WEBPACK_NONCE),
-    }),
-  },
-}
+const express = require("express");
+const expressNunjucks = require("express-nunjucks");
+//...
+app.get("/", function(req, res) {
+  res.render("index", { noncePlaceholder: res.locals.nonce });
```
dav-is: Don't forget to change to `res.locals.styleNonce`

docs/csp.md
```diff
@@ -41,45 +50,55 @@ app.use(
...
-(`__webpack_nonce__` is a poorly documented variable which, when set, JSS will read and apply to `<style />` elements: `<style nonce={webpack-nonce-value} />`).
+The tag must have `id="jss-nonce"` and have a `data-value` attribute with the string value to use as the nonce.
+
+Now JSS can apply the nonce to `<style />` elements: `<style nonce={webpack-nonce-value} />`
 
 At this point, the style blocks generated by JSS should render with the nonce value (`<style nonce />`) when you inspect the page.
 
 The browser might not show the nonce value inside the style tag when you inspect the page, but it's there regardless.
 
 ## Setting the nonce attribute in styling loaded by Webpack
```
dav-is: Is this still appropriate?

Graham42:
Can do.

docs/csp.md
Graham42: :+1:

docs/csp.md
Graham42: I don't really care what format. But formatting should be done by an auto-formatter.
I didn't see anything in place, so I copied the code into https://prettier.io/playground/ and back. If you have a prefered format, please give me an easy way to format the code.

kof: havn't started to use prettier yet, still rely on eslint here, but docs are not linted, mb there is  a way to check code blocks in md files by eslint as well?

dav-is: https://github.com/eslint/eslint-plugin-markdown

kof: added in master

kof: there are plenty of errors now from md files, just lint try to lint the changed file with lint:staged or over commit hook

kof:
@Graham42 is there anything I can help with so we can finish this one?

Graham42:
I just need time to get to it, hopefully this weekend.

kof:
Great, no worries, just checking with you.

docs/csp.md
Graham42: Not sure what you mean by a "more universal name" ?

kof: Thats what I ment by `<meta property="nonce" content="asdfasdfaasdf" />` nonce can be used for anything else on the page that requires nonce.

dav-is: For instance, in my most recent project, I have a  nonce for js and one for css. They should be different. OAuth and APIs have the concept of nonces as well. By universal I mean someone could just copy this code and it would work. It’s possible someone already has a meta tag “nonce” for something else. Simply being specific is all.

Graham42:
Since we've had a bunch of churn in this PR, rather than add 3 more commits, I decided to rebase and squash. I think this should have all of the things suggested now.

docs/csp.md
kof: Why exactly should they be different? I thought nonce is nonce, no matter for what it is used?

Graham42: You _can_ use different nonces for script vs style, but you don't have to. But just "nonce" on it's own is probably too generic, nonces can be used for anything, not limited to CSP context.

kof: ok, what about "csp-nonce" as a name, it is still for more than JSS, but seems now specific enought to identify properly the nonce value, no?

kof: is there any reason one would use different nonces for js and for css?

kof: Where I am heading with this, in case one uses more than one library on the page, for e.g. JSS and styled-components, they could share the same nonce token.

dav-is: I like your idea. I think `csp-nonce` is probably the best variable to use across the board.

kof: so then `<meta property="csp-nonce" content="asdfasdfaasdf" />`

docs/csp.md
Graham42: I'm going to assume you meant the webpack reference, I removed it now

docs/csp.md
Graham42: Updated

src/renderers/DomRenderer.js
```diff
@@ -248,6 +248,14 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
   getHead().insertBefore(style, prevNode)
 }
 
+/**
+ * Read jss nonce setting from the page if the user has set it.
+ */
+function getNonceFromPage() : ? string {
```
kof: Right now you are running this query when stylesheet is created each time, its a bit of unnecessarily, we can cache it similar to getHead. Also as this is the second place we need memoization, its worth adding a simple local memoze helper:

```js
const memoize = (fn) => {
  let value
  return () => {
    if (!value) value = fn()
    return value
  }
}
```

```js
const getNonce = memoize(() => {
   ....return the nonce
})

const getBody = memoize(() => {
   ....return the body
})
```

Graham42: Can do.

Graham42: I've cached the value as you suggested.

However, Flow was giving me a bunch of trouble with that `memoize` function. It probably requires generics or something and I'm really not familiar with Flow. Perhaps this can be done as an future enhancement  by someone who is more familiar with it.

kof: just don't add typings to memoize

kof: we should be adding typings anyways to public api only

Graham42: Even without adding typings, the type inference causes problems.

kof: does using any solve it?

Graham42: I'm going to say it's better to repeat some trivial code twice rather than lose the type checking. If this was being done 3+ times then maybe

kof: I can check it out if you commit with --no-verify

Graham42: there you go

kof:
alright, fixed the typing, lets continue here #672

PR #670, opened 2018-01-23
Use Phantom for PR builds

Graham42:
This might fix PR builds. Opening the PR to trigger build to find out.


kof:
what about core maintainers pull requests?

Graham42:
Going to have to give this more thought

iamstarkov:
I would recommend puppeteer for more reliable tests

kof:
Ideally would be to still run tests in all browsers like we normally do.

Issue #651, opened 2017-12-17
Start using prettier


kof:
1. Start using https://github.com/prettier/eslint-config-prettier
1. Check settings of https://github.com/cssinjs/eslint-config-jss if they are compatible with prettier
1. Start using prettier via mrm https://www.npmjs.com/package/mrm-task-prettier

Graham42:
PR opened https://github.com/cssinjs/jss/pull/671

kof:
closing this for core, needs to be done for all plugins though

PR #671, opened 2018-01-27
Use prettier to autoformat code magically

Graham42:
Fixes: #651

Tried to choose configuration that matched previous style (single quotes, no semi)

Also remove markdown plugin for eslint for now because it's causing a whole bunch of errors and the whole reason it was added was because I requested a formatter. But Prettier solves this problem better.

The only conflict I ran into with eslint was in a few tests where strings were previous concatenated over multiple lines to look like css, and then prettier put them all on one line. Solved with a couple comments. For more details and a future alternative see the [commit message](https://github.com/cssinjs/jss/pull/671/commits/f6fc0db81748f07923bb1cbdb83a1ae52b57415c). 

If at any time this PR needs to be updated, can easily be done with an interactive rebase:
<details>

* choose to drop the commit that is "Format code with prettier"
*  choose to edit the commit before

Should look like:

```
pick f9c5d6d Treat package-lock.json like binary
edit bf9ba69 Configure prettier
drop 4c63cc2 Format code using prettier
pick d1018bd Add empty comments to make eslint happy with prettier
pick 9430d7e Check prettier status in pre-commit hook
```

Then when the rebase stops after the "Configure" commit, run `npm run format`, and create a new commit "Format code using prettier". 

The rebase should then continue with out any issues
</details>


kof:
merged, thanks!

PR #678, opened 2018-02-05
Remove property if  null passed to .prop() and null|undefined returned from fn values

kof:


kof:
Fixes #514

kof:
cc @jaminthorns, @AleshaOleg @lttb @iamstarkov

kof:
The question is though if we should handle `false` and empty strings same way

kof:
I think we should at least support `false` because  people will use it like this: `{color: (data) => isActive && data.activeColor}`

PR #694, opened 2018-04-02
Fix flow type definition of StyleSheet.update function

HenriBeck:
Fixed StyleSheet.update param types to be the same as RuleList.update


HenriBeck:
Yeah, no need for a separate release, just noticed it when I wrote my own implementation and thought I would just do a quick fix for the future.

PR #703, opened 2018-04-17
Issue #702

### LINKED ISSUE #702: Clarify array syntax and "!important"
# Motivation
Array syntax description is misleading. In particular, the examples with `!important` are not working as described:

```js
const styles = {
  button: {
    color: ['red', '!important'],
    margin: [[5, 10], '!important']
  }
}
```

Should be (note the second brackets around `'red'`):

```js
const styles = {
  button: {
    color: [['red'], '!important'],
    margin: [[5, 10], '!important']
  }
}
```

# Tasks
- [ ] Update link to array notation
- [ ] Update example

davidbonnet:
Fixes #702

### DIFF
--- a/docs/json-api.md
+++ b/docs/json-api.md
@@ -36,12 +36,12 @@ const styles = {
 
 ### Support of "!important"
 
-To use the `!important` modifier with function values, you must use [array syntax](./js-api.md#alternative-syntax-for-space-and-comma-separated-values):
+To use the `!important` modifier with function values, you must use [array syntax](#alternative-syntax-for-space-and-comma-separated-values):
 
 ```javascript
 const styles = {
   button: {
-    color: data => [data.color, '!important']
+    color: data => [[data.color], '!important']
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/json-api.md
```diff
@@ -352,7 +352,7 @@ Compiles to:
 ```javascript
 const styles = {
   button: {
-    color: ['red', '!important'],
+    color: [['red'], '!important'],
     margin: [[5, 10], '!important']
```
kof: this one works

docs/json-api.md
```diff
@@ -352,7 +352,7 @@ Compiles to:
 ```javascript
 const styles = {
   button: {
-    color: ['red', '!important'],
+    color: [['red'], '!important'],
```
kof: oh I just realized, it depends on jss-expand, margin: [5, 10, '!important'] would work, but not color: ['red', '!important']

kof:
merged, thanks

docs/json-api.md
davidbonnet: Oh, so shouldn't we document this then?

kof: it is basically jss-expand documentation, maybe we should cut it a little bit in the json-api.md as well?

PR #712, opened 2018-05-13
upgrade to flow 0.72.0

atabel:


src/PluginsRegistry.js
```diff
@@ -69,7 +69,7 @@ export default class PluginsRegistry {
   /**
    * Call `onUpdate` hooks.
    */
-  onUpdate(data: Object, rule: Rule, sheet: StyleSheet): void {
+  onUpdate(data: ?Object, rule: ?Rule, sheet: StyleSheet): void {
```
kof: where is it comming from? I don't think those can be optional

TrySound: Seems like data argument can be `undefined`. I added checks and integration tests are failed when data is always object.

atabel: `data` is optional here https://github.com/cssinjs/jss/blob/master/src/RuleList.js#L137

kof:
closing this one as we just merged this https://github.com/cssinjs/jss/pull/716

Issue #649, opened 2017-12-16
Try to bundle with microbundle


kof:
https://github.com/developit/microbundle

kof:
we went for using rollup directly

PR #713, opened 2018-05-13
Bundle UMD with rollup

TrySound:
Here's size comparison

webpack

jss.js 86.04K
jss.js.map 103.0K
jss.min.js 33.71K

rollup

jss.js 59.53K
jss.js.map 97.22K
jss.min.js 21.15K

For further track of bunlde sizes I added size snapshot plugin.

I temporary inlined babel config to rollup. Will fix that in the next PR.


TrySound:
Not sure why the test is failed.

kof:
Wanna try microbundle? https://github.com/cssinjs/jss/issues/649

TrySound:
I wouldn't call it "micro". It contains a huge dependency tail which I would like to omit. Configuration is actually dead simple and will be even more simple with a couple of next PRs.

kof:
Yeah and no. What microbundle depends on is not very important to me, what I have to do is more important, because I have to maintain a bunch of repositories and microbundle seems to be targeted for the libs, so there might be things you are not considering right now.

TrySound:
Let me take care about all repos and their dependencies and then you could decide the necessity of that huge abstraction.

kof:
Even if you help with updating all repos (huge deal). Please take a look at microbundle more closely. It seems to have way smaller configuration footprint. Its dependencies is not that big deal!

kof:
Microbundle was created to reduce configuration optimized for libraries on top of rollup.

TrySound:
I saw it and see a couple of problems.

buble+flow-remove-types over babel is not the best option. I don't actually believe in buble stability (I worked on it for some time).

It shows only gzipped size. My size snapshot plugin creates json which may see every user and shows some more info like the size which will consume user bundler, the size which will load browser and the size which will parse browser. Also for esm bundles it shows treeshakability of the project by both webpack and rollup. All these metrics are important.

My opinion is that microbundle just increases complexity and maintainability over simple rollup config.

kof:
Hmm, what if we create own "microbundle" using dependencies you wanted? So that we can keep repos kompletely clean and only use 1 dependency and 1 cli command for the build?

TrySound:
Let's create "a perfect" config for all repos at first and then will try to reuse it. I will completely handle it.

kof:
I am in

package.json
```diff
@@ -92,8 +102,7 @@
     "all": "npm run lint && npm run flow && npm run test && npm run build && npm run size",
     "build": "npm run clean && npm run build:lib && npm run build:max && npm run build:min && npm run build:tests",
     "build:lib": "babel src --out-dir lib && npm run flow:copy-src",
-    "build:max": "cross-env NODE_ENV=development webpack src/index.js dist/jss.js",
-    "build:min": "cross-env NODE_ENV=production webpack src/index.js dist/jss.min.js",
+    "build:umd": "rollup -c",
```
kof: will this generate minified and non-minified version?

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+    input,
+    output: {
+      file: 'dist/jss.min.js',
+      format: 'umd',
+      exports: 'named',
+      name
+    },
+    plugins: [
+      nodeResolve(),
+      commonjs({include: /node_modules/}),
```
kof: why the difference with line 31?

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+      babel(getBabelOptions()),
+      replace({'process.env.NODE_ENV': JSON.stringify('production')}),
+      sizeSnapshot(),
+      uglify()
+    ]
+  }
+]
```
kof: seems like we can reuse most of it, use production flag via cli, add uglify for production

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+
+  {
+    input,
+    output: {
+      file: 'dist/jss.min.js',
+      format: 'umd',
+      exports: 'named',
+      name
+    },
+    plugins: [
```
kof: seems like we can reuse most of it, use production flag via cli, add uglify for production

rollup.config.js
```diff
@@ -0,0 +1,55 @@
+import nodeResolve from 'rollup-plugin-node-resolve'
+import commonjs from 'rollup-plugin-commonjs'
+import babel from 'rollup-plugin-babel'
+import replace from 'rollup-plugin-replace'
+import uglify from 'rollup-plugin-uglify'
+import {sizeSnapshot} from 'rollup-plugin-size-snapshot'
+
+const input = './src/index.js'
+
+const name = 'jss'
```
kof: we can read it from the package.json automatically, right?

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+
+const getBabelOptions = () => ({
```
kof: why is it a function?

kof: can we use babelrc we already have?

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+      sourcemap: true,
+      exports: 'named',
+      name
+    },
+    plugins: [
+      nodeResolve(),
+      commonjs({include: '**/node_modules/**'}),
+      babel(getBabelOptions()),
+      replace({'process.env.NODE_ENV': JSON.stringify('development')}),
+      sizeSnapshot()
```
kof: do we still need size-limit utility? Can sizeSnapshot exit with error?

package.json
TrySound: Yes. I use multiconfig, array of configs.

rollup.config.js
TrySound: I write these configs for tens of project last months and this will be overreusing. It's more readable to just duplicate a couple of characters but reuse important names and babel options. Let's keep it this way for now. I'm gonna change it in next PRs.

rollup.config.js
TrySound: It's usually not the same name. package.json contains dashed name. This `name` is camel cased for window object. So not sure we should reuse it.

rollup.config.js
TrySound: It's temporary I will tidy up babel configs in the next PR.

TrySound:
Done

rollup.config.js
kof: we can camelize

TrySound: It's not always fit to camalize. For example there is cytoscape.js project which creates Cytoscape global variable, or react-powerplug which creates ReactPowerPlug global variable. It's better to set it manually.

kof: In our case it will work, because naming is consistent.

TrySound:
Let me first propose more changes.

TrySound:
So can I merge this?

kof:
Oh I didn't realize you wanted to merge it in this repo, I thought we would go directly into separate package?

TrySound:
I would like to achieve the best configuration at first and then move what we can to that repo.

kof:
What do you think is still missing?

TrySound:
I'd like to upgrade to babel7, change flow distribution and discuss flat bundles. Let's discuss everything in different PRs step by step to not create a mess.

kof:
agreed

PR #717, opened 2018-05-20
Simplify flow distribution

TrySound:
Since we already have all types in `src` we can reuse them instead of
copying. And we don't even need a tool for that.

I add src to published files and generate this file with simple `echo`
```js
// @flow

export * from '../src';
```

Prevented starting server on flow check which could lead to performance
problems and crashing in CI.


kof:
merged

package.json
```diff
@@ -98,16 +97,16 @@
...
   },
   "scripts": {
-    "all": "npm run lint && npm run flow && npm run test && npm run build && npm run size",
+    "all": "npm run lint && npm run typecheck && npm run test && npm run build",
     "build": "npm run clean && npm run build:lib && npm run build:umd && npm run build:tests",
-    "build:lib": "babel src --out-dir lib && npm run flow:copy-src",
+    "build:lib": "babel src --out-dir lib && npm run build:flow",
     "build:umd": "rollup -c",
     "build:tests": "cross-env NODE_ENV=test webpack tests/index.js tmp/tests.js",
+    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" > lib/index.js.flow",
```
kof: wait, does echo work on windows?

TrySound: It always worked for me there. For my previous coworkers on windows this way also worked afaik.

Issue #714, opened 2018-05-14
Mono Repo and NPM org


HenriBeck:
I would like to propose to move the whole Jss project into a [Lerna](https://github.com/lerna/lerna) mono repo like babel and other projects did. I would also then move to a fixed mode for the versioning (like babel). This would also solve some problems with maintaining all of the dev dependencies of the repositories.

This would include the following repositories:
- jss
- react-jss, styled-jss and aphrodite-jss
- all of the currently provided plugins by Jss

Another idea would be to create npm organization and publish most packages under this organization.
 - jss -> `@jss/core`
 - react-jss (integrations) -> `@jss/react`
 - jss-expand (plugins): `@jss/plugin-expand`

oliviertassinari:
I can only approve the idea 👍

TrySound:
And this will solve the problem of reusing builder.

kof:
I would need someone to commit finishing this migration, who can?

TrySound:
So do we go with proposed jss scope?

HenriBeck:
I would also include the provided integrations from the jss team in the monorepo. That way we could easily share code between them as well (Updating styled-jss to use the same JssProvider for example).

I could migrate it, but I won't have time before next week.

kof:
After next week is fine, just need to be sure we can finish it and not leave it in half migrated state for months.

The reason I am not sure I would put them all into monorepo is because they have partially different setup, separate issue trackers and it feels like they are entirely separate projects. 

We need a clear decision framework on how to decide.

One possibility: everything that depends on JSS core, should be part of the monorepo, everything else - not.

kof:
Ok, we got merged the first pr towards monorepo, now we need to move plugins to it

kof:
@HenriBeck are you up for this?

HenriBeck:
All of the plugin repos are now imported. Any other repos that should be imported?

kof:
I checked, all plugins are here, is there anything else that is topping us from jss 10 release?

HenriBeck:
I think we should need to archive the plugin repos and add a message to the readme.

kof:
yes, immediately after release

kof:
oh btw. we should check all the docs and fix the links to old repos

HenriBeck:
Yeah, we need to update the links in the package.json files

kof:
also mb add docs regarding esm builds? and generaly whats in the dist?

kof:
[...]

PR #720, opened 2018-05-28
Monorepo

HenriBeck:
Started on converting this repo into a monorepo.

Some things to discuss:
- Fixed or independent version mode?
- Which repositories to import?
- Moving to an NPM org as well? (like babel)

#714


package.json
```diff
@@ -1,142 +1,5 @@
 {
-  "name": "jss",
```
TrySound: Keep the `name` and add `private: false`.

TrySound:
I'm thinking maybe it will be better to move all plugins first and provide them build step and then work with `jss` itself. Because with current strategy we move a lot of stuff deeper and then we will a lot of stuff hoist in the root.

TrySound:
Also it would be good to move each plugin in separate PRs so we could review faster.

TrySound:
Also we probably should not change package names and publish one patch with new links so nobody tried to look at old repos.

HenriBeck:
I would rather move `jss` in this pr and setup build and test scripts in this pr and then move the other packages in separate PRs.

@TrySound so you are against switching to an npm org?

package.json
```diff
@@ -1,142 +1,5 @@
...
-  "description": "A lib for generating Style Sheets with JavaScript.",
-  "version": "9.8.1",
-  "author": {
-    "name": "Oleg Slobodskoi",
-    "email": "oleg008@gmail.com"
-  },
-  "repository": {
```
TrySound: `author` and `repository` should probably also stay here

package.json
```diff
@@ -1,142 +1,5 @@
...
-    "css",
-    "components",
-    "composable",
-    "css in js",
-    "css-in-js"
-  ],
-  "license": "MIT",
-  "main": "./lib/index.js",
-  "engines": {
-    "node": ">=4"
```
TrySound: This should stay here to validate devDependencies

package.json
```diff
@@ -1,142 +1,5 @@
...
-  },
-  "devDependencies": {
```
TrySound: All development stuff should stay here. It's cleaner when packages have only production dependencies.

package.json
```diff
@@ -1,142 +1,5 @@
...
-  "files": [
-    "dist",
-    "lib",
-    "src",
-    "flow-typed",
-    "readme.md",
-    "changelog.md",
-    "LICENSE"
-  ],
-  "lint-staged": {
```
TrySound: `lint-staged` should also be run from the root

kof:
btw. we don't own jss namespace on npm, we do https://www.npmjs.com/~cssinjs though

package.json
HenriBeck: Yeah, I still need to move all the dev deps and setup scripts in the root package.

kof:
Wait, we don't need the user name to publish under @jss scope

kof:
Just looked at babel, they are not using the namespace. What are the pros/cons?

HenriBeck:
https://www.npmjs.com/docs/orgs/migrating-a-user-account.html I think thats the problem, there is already a user with the jss name.

Babel switched to namespaces with v7

kof:
I asked the person, lets think of alternatives.

HenriBeck:
Well, we don't have to switch, it was more of an idea. I think it just unifies the packages a bit more.

kof:
We can migrate with current package names first, then if we get a good namespace we can change that as a separate step. Lets focus on things which are necessary

kof:
@HenriBeck we got some conficts and the comments from @TrySound

HenriBeck:
A few things about the monorepo:

- The packages will have their own build scripts
- The packages will have their own test scripts as those might differ
- I moved some configs like the rollup config into the root folder and made created a config function to be used by the packages
- Everything else lives in the root folder now (lint etc.)

Some problems which still exists for some reason:
- Babel always throws an error, this makes the test and the build scripts fail

kof:
What if we unify all the build stuff and put only compatible repos into this monorepo? For e.g. jss and plugins only for now?

kof:
I have the idea to move to the same testing library everywhere, but for this we need to extract browser specific part into a separate library which can keep the current complex setup with karma.

kof:
the thing is if we keep each package with its own build, there is not much value in monorepo, I think the big value is to reduce maintenance cost of updating each repo, so we need a unification.

wdyt?

TrySound:
You may keep individual builds. I can hoist them in the next PR then.

kof:
lets do it before the merge @TrySound you can commit directly to this branch as well

TrySound:
I just would like to do atomic pull requests to simplify review and see the state of work.

kof:
that is true, unless master gets into a state where we can't release from, also you need to know that cssinjs.org website is dynamically fetching markdown files from each repo and it will need to be fixed too

kof:
So if we have these points, we can merge:
- Can continue releasing JSS core
- Can continue releasing plugins
- Didn't break the website, can continue changing text
- Didn't break the CI

HenriBeck:
About the build step, currently, each package has its own build npm script. We could do an approach like storybook did with having js script files which we can simply call, but each package would still need an npm script.

I've already extracted the configuration for the rollup config into a function so the different plugins have a unified config, the only things that change between them are the names.

kof:
Yeah, its a good first step I think

.flowconfig
```diff
@@ -3,6 +3,7 @@
 [include]
 
 [libs]
+packages/jss/flow-typed
```
TrySound: already removed

.eslintignore
```diff
@@ -1 +1,4 @@
-flow-typed/
+packages/jss/flow-typed/
```
TrySound: already removed

TrySound:
@HenriBeck Could you rebase so we could focus on actual changes?

HenriBeck:
Yeah will rebase and finish work on this tomorrow

HenriBeck:
As there are a lot of changes here, which when rebasing would completely revert my changes so far, I decided to start clean on a new branch too make commits a bit tidier

HenriBeck:
#729

PR #725, opened 2018-06-10
Move cssom types to sources

TrySound:
This fixes missing global types in userland.
Previously types in `flow-typed` was treated as local and
didn't extend global namespace.


kof:
merged 🎉

kof:
released

Issue #736, opened 2018-06-21
Resolve global to window


kof:
Webpack has this built in, we broke it during migration to rollup. Should we use https://www.npmjs.com/package/rollup-plugin-node-globals ?

Originates in #735

TrySound:
I would use this explicitly
https://github.com/calvinmetcalf/rollup-plugin-node-globals/blob/master/src/global.js

kof:
wanna add it?

TrySound:
Sure

kof:
released 9.8.7

PR #741, opened 2018-06-23
Use rollup globals plugin #736

kof:


oliviertassinari:
You might want to set `ignoreGobals` to `true`. But 💯 sure about what's it doing but I think that it can avoid conflict: https://github.com/mui-org/material-ui/pull/11878/files#diff-81aa76013a7bbe44ca75536783d9df84R21.

kof:
@oliviertassinari thanks, added

kof:
merged

PR #748, opened 2018-07-06
Add jss preset default

HenriBeck:
Add jss-preset-default to the monorepo.

Do not merge yet because there is a problem with rollup.


changelog.md
```diff
@@ -1,14 +1,14 @@
 ## 9.8.7 / 2018-06-24
 
-* Fix `global` access in the dist version (#736)
```
kof: I wonder why we have this all the time, is someones editor using wrong config?

kof: these things should have been fixed when you run the linter and prettier over cli, did you get  this changes from editor?

kof:
can we merge the repo without loosing the history and contributors?

packages/jss-preset-default/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "changelog.md",
+    "LICENSE"
+  ],
+  "license": "MIT",
+  "main": "./lib/index.js",
+  "devDependencies": {
+    "jss": "^9.8.0"
+  },
+  "peerDependencies": {
+    "jss": "^9.7.0"
```
kof: probably not your fault but versions should be the same

changelog.md
HenriBeck: Probably, I didn't change that explicitly.

packages/jss-preset-default/src/index.js
```diff
@@ -0,0 +1,40 @@
+// @flow
```
kof: if you add flow as a separate step, you will speed up merging

HenriBeck:
Yes we can import the whole history

packages/jss-preset-default/src/index.js
```diff
@@ -0,0 +1,40 @@
...
+import nested from 'jss-nested'
+import compose from 'jss-compose'
+import camelCase from 'jss-camel-case'
+import defaultUnit from 'jss-default-unit'
+import expand from 'jss-expand'
+import vendorPrefixer from 'jss-vendor-prefixer'
+import propsSort from 'jss-props-sort'
+
+type Options = {
+  template?: {},
```
kof: in theory once all other plugins are typed, you could import each type from there and use it here

packages/jss/src/types/jss.js
```diff
@@ -9,7 +9,7 @@ import SimpleRule from '../rules/SimpleRule'
 import FontFaceRule from '../rules/FontFaceRule'
 import type {CSSStyleRule} from './cssom'
 
-export type Classes = { [string]: string }
```
kof: same here, please check why we get these changes, back and forth

HenriBeck: Pretty sure that this was prettier because we set the bracket-spacing to false

kof: why wasn't it fixed before through lint-staged?

kof: I mean in previous releases

packages/jss-preset-default/src/index.js
HenriBeck: I added flow here already because it's a really small change. I bigger plugins I will definitely create a separate PR for flow

packages/jss/src/types/jss.js
kof: I don't even see it with spaces in master

HenriBeck: https://github.com/cssinjs/jss/blob/master/packages/jss/src/types/jss.js#L12 I do🤔

kof: oh, sorry, npm run format fixed it for me locally and I didn't notice

HenriBeck:
Yeah, will do when I fixed rollup. It's not creating the umd bundle for the jss-preset-default

HenriBeck:
@TrySound could you help me out here? For some reason, rollup doesn't build the jss-preset-default package. But it correctly bundles the jss package.

TrySound:
Will take a look

HenriBeck:
@TrySound did you already look at the problem? Because I literally have no idea. It's not even throwing an error or logging anything out.

HenriBeck:
@kof, I fixed the build by removing the `jss-global` plugin which caused the `umd` the build to fail silently. I currently have no idea why that's case.

kof:
Lets figure out what it is, we can't go without this plugin

HenriBeck:
Interestingly enough, when I copy the jss-global package and build it using the new setup and symlink it, it gives me an error:

[!] Error: 'default' is not exported by ../jss-global/lib/index.js

HenriBeck:
will be recreated once #767 is merged.

PR #749, opened 2018-07-09
[WIP] jss-observable

kof:


HenriBeck:
Needs to wait for #748 as the rollup build is currently broken.

HenriBeck:
This should be done.

Publishing this would mean a breaking change. Currently the observable plugin is not automatically applied, because I wanted to avoid a dependency loop (this might cause some weird issues).

What is done:
- Moved all of the observable code out of the jss package
- Moved all of the tests as well

What could be better:
- Remove the `jss` dependency from the `jss-observable` package. This would solve the issue with the breaking change. Currently the only thing being used from the `jss` package is the `createRule` function.

kof:
Plugins should only have peerDependencies on jss. We need to release this one first. Then release jss-preset-default with this plugin and at the same time release jss without the built-in version.

packages/jss-observable/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "files": [
+    "dist",
+    "lib",
+    "src",
+    "readme.md",
+    "changelog.md",
+    "LICENSE"
+  ],
+  "dependencies": {
+    "is-observable": "^1.1.0"
```
kof: there was a problem with is-observable, see in the issues, thats why we have own version of this function. I think it is small enough to just be copied

HenriBeck: Yeah, just saw this on travis. Where should we put this function, because jss uses that in the `cloneStyle` function.

packages/jss-observable/src/index.js
```diff
@@ -0,0 +1,47 @@
+/* @flow */
+import isObservable from 'is-observable'
+import type StyleRule from 'jss/src/rules/StyleRule'
+import createRule from 'jss/lib/utils/createRule'
+import type {Rule, RuleOptions, JssStyle} from '../../jss/src/types'
```
kof: Is that ok to import types that way? I thought they have to be part of public interface and imported differently

HenriBeck:
Yeah, but this would be mean a breaking change. The problem was when I tried to import the `jss-observable` package in the actual jss to avoid a breaking change.

packages/jss-observable/src/index.js
HenriBeck: Well, the types are going to be stripped out anyway. It's not the best solution but exporting everything isn't either.

packages/jss-observable/package.json
kof: Yeah I wonder how to separate it properly. Ideally it should be entirely encapsulated in jss-observable

HenriBeck: Yeah, also had that utility function inside `jss-observable` but that created the dependency cycle again and having it inside jss isn't right either. Could we create a PR to not use const in the `is-observable` package?

packages/jss-observable/src/index.js
kof: Well, we could evtl go with this for now and rethink later

packages/jss-observable/package.json
kof: no, is-observable package is for node only, see closed issues there, we talked to the maintainer

kof: one way is to create utilities package and have it there, another way is to duplicate this function, its one line

HenriBeck: Yeah, let's go with duplicating the line for now until we find a better way.

kof: yep

kof:
LGTM, just go ahead and merge when ready

PR #753, opened 2018-07-12
Updated build script

HenriBeck:
Because we have problems with the `jss-global` and `jss-preset-default` package, I extracted the changes to the build system out so we can work on other packages before solving those problems.

- Fixed rollup build for other packages
- Fixed inline the correct version
- Fixed Travis tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "lint": "eslint scripts/ packages/ && npm run typecheck && npm run format",
     "lint:staged": "lint-staged && npm run typecheck",
     "format": "prettier --write *.{js,md,json} {docs,packages,scripts}/**/*.{js,md,json}",
-    "test": "cross-env NODE_ENV=test karma start --single-run",
+    "test": "cross-env NODE_ENV=test VERSION=\"1\" karma start --single-run",
     "test:watch": "cross-env NODE_ENV=test karma start",
     "posttest": "lerna exec -- cross-env SNAPSHOT=match rollup --config ../../rollup.config.js && [ -z \"$TRAVIS\" ] || codecov",
     "codecov": "codecov",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,4 @@
 {
-  "name": "jss",
-  "description": "A lib for generating Style Sheets with JavaScript.",
-  "version": "9.8.7",
   "author": {
     "name": "Oleg Slobodskoi",
     "email": "oleg008@gmail.com"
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -10,18 +7,28 @@
     "type": "git",
     "url": "git@github.com:cssinjs/jss.git"
   },
-  "keywords": [
```
kof: keywords gone?

kof: ah I see, they are in the jss package

HenriBeck: Yeah, I removed some unnecessary things in the root package.json

HenriBeck:
Okay, tests are fixed now.

HenriBeck:
Is this ready to be merged? I would work on the observable and function plugin first.

rollup.config.js
```diff
@@ -25,34 +31,32 @@ const base = {
     globals(),
     replace({
       'process.env.NODE_ENV': JSON.stringify('development'),
-      __VERSION__: JSON.stringify(pkg.version)
+      'process.env.VERSION': JSON.stringify(pkg.version)
     }),
     sizeSnapshot({
       matchSnapshot,
-      snapshotPath: path.join(packagesPath, 'jss/.size-snapshot.json')
+      snapshotPath: './.size-snapshot.json'
```
kof: why would you change things like that, I just recently made them all consistently full paths, since in some cases relative path was not properly resolved

webpack.config.js
```diff
@@ -12,7 +12,6 @@ const isTest = env === 'test'
 
 const plugins = [
   new webpack.DefinePlugin({
-    'process.env.NODE_ENV': JSON.stringify(env),
```
kof: are you sure about that?

kof: It could be in any other package, like in the "warning" package we are using

kof:
Please avoid adding changes to the code in the same PR which are not related to the goal of the PR. It makes it hard to review.

kof:
inline-version worked before, I am not aware it was broken

rollup.config.js
kof: Oh is it supposed to resolve relative to each package dir?

webpack.config.js
kof: please get this one back before merging

HenriBeck:
Inline version doesn't work with a monorepo. That's why I had to switch to env variables and use the babel plugin for inlining them.

kof:
I see, thanks!

rollup.config.js
HenriBeck: Yes, before we got a `.size-snapshot.json` file in the root folder. That way we get an individual file for every package.

webpack.config.js
HenriBeck: Ohhh, yeah, removed that because we are using the babel-plugin-inline-env-variables now, but forgot that this could still be in node_modules.

PR #755, opened 2018-07-13
Jss plugin syntax rule value function

HenriBeck:
Tests inside the other packages which depend on the function plugin are currently broken. I don't see an easy way of fixing those tests though. Only when importing the other repositories.


packages/jss-plugin-syntax-rule-value-function/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "jss",
+    "style",
+    "sheet",
+    "stylesheet",
+    "css",
+    "components",
+    "composable",
+    "css in js",
+    "css-in-js",
+    "observable",
```
kof: these last 3 are from observable, I think we can use: "dynamic", "function", "reactive", "animation"

packages/jss-plugin-syntax-rule-value-function/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "jss-plugin-syntax-rule-value-function",
+  "description": "JSS plugin for function syntax",
```
kof: for function value and rule syntax

packages/jss-plugin-syntax-rule-value-function/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "version": "0.1.0",
```
kof: I guess it should correspond the jss version?

HenriBeck: Lerna will take care of it when we publish them.

packages/jss-plugin-syntax-rule-value-function/src/integration.test.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable no-underscore-dangle */
```
kof: we can just call it index.test.js

packages/jss/src/Jss.js
```diff
@@ -22,7 +21,7 @@ import type {
   generateClassName
 } from './types'
 
-const defaultPlugins = rulesPlugins.concat([functionsPlugin])
+const defaultPlugins = rulesPlugins
```
kof: we don't need defaultPlugins constant any more

packages/jss-plugin-syntax-rule-value-function/package.json
HenriBeck: https://github.com/lerna/lerna#publish

packages/jss/tests/integration/plugins.js
```diff
@@ -387,7 +387,7 @@ describe('Integration: plugins', () => {
       })
       sheet = jss.createStyleSheet({
         a: {
-          color: data => data.color
```
kof: how is it possible that this test works? method .update triggers onUpdate only for functions ... hmmm

packages/jss-plugin-syntax-rule-value-function/.size-snapshot.json
```diff
@@ -0,0 +1,32 @@
+{
+  "dist/jss-observable.js": {
```
kof: why do we have this one here?

packages/jss/tests/integration/plugins.js
HenriBeck: onUpdate is being called for every plugin which implements onUpdate. That's why it works. It's not actually bound to the function plugin which is good.

kof: uhh, my bad

PR #756, opened 2018-07-14
Added jss-expand package

HenriBeck:
There was an error while importing the package, which means we can't import the whole git history.


kof:
I think this would be a good name for it jss-plugin-syntax-expand

HenriBeck:
Renamed to jss-plugin-syntax-expand

kof:
Unrelated: trying to run tests on master, it fails in many places, are there any steps one should do to run the tests? missing docs?

HenriBeck:
Well yeah, some tests are failing because we moved out the function plugin and some tests inside the plugins actually depend on the function values.

kof:
so how do we fix it? moving function values tests to the corresponding plugins?

HenriBeck:
They will be fixed once we move the plugins to the monorepo and can set up the function plugin for the tests. I already did that for this plugin: jss-plugin-syntax-expand

kof:
Alright, lets get through this, because we can't release anything right now

kof:
sure and add a hint to the new repo

kof:
wait, we need to migrate all links first

kof:
and release a new version with migrated plugins

kof:
a todo?

HenriBeck:
which links to need to be migrated?

kof:
everything that refers to the old repo, docs mostly

HenriBeck:
Yeah, I think we can only update links when most of the repos are migrated because they contain links to other repos.

Issue #579, opened 2017-09-07
Static CSS files extraction


kof:
We can create a  babel plugin and let all styles be extracted to just css files. 

- Dynamic values wouldn't be possible of course.
- Static vendor prefixer need to be used.
- We could put styles in Component.css files and add an `import classes from './Component.css'` to the Component.js file.

StephanBijzitter:
Perhaps inspiration can be gotten from [babel-plugin-css-in-js](https://www.npmjs.com/package/babel-plugin-css-in-js).

The only significant downside with this plugin is that usage with webpack-dev-server (and hot module reloading enabled) does reload the page, as compilation is done in two steps.

1. user edits styles in a js file
2. webpack detects changes to .js and hard-reloads page (new .css is now created)
3. webpack detects changes to .css and hot-reloads page

It's definitely a good improvement, but it can also be made better.

kof:
also found this https://www.npmjs.com/package/extract-jss-webpack-plugin

mehrdaad:
any updates on this?

kof:
Additional thoughts: if we are not allowed to use any dynamic capabilities like function values etc which can access render time variables, why not put js styles in separate files, name them something.styles.js and just write a small script to extract them into css files, it is a few lines of code and not worth creating a package/library for it.

kof:
This way you can still use JSS for actual dynamic stuff at the same time, by just keeping dynamic CSS inside of the files with components or .js files, so that they stay untouched.

kof:
Mb I should just add an example for something like that and close this issue.

kof:
If we do extraction using babel from js files, we have an issue that it might be confusing: "why with static extraction certain things don't work", which might be not obvious to people who are new to all this.

kof:
If you really want to put jss styles into the same js file and still have an extraction, you can build stuff in a way that you can require every file without side effects and export styles, so that you can do the extraction same way, like we can do with separate files.

```js
// MyButton.js
export const styles = {}
export default MyButton = () => ()
```

```js
jss.createStyleSheet(require('MyButton').styles).toString()
```

kof:
One problem you still have, in case you don't want to generate styles at runtime at all, is the `injectSheet()` fn call if react-jss for e.g. is used. 

In that case injectSheet should simply be replace by a dummy function which simply returns the component itself.

```js
const injectSheet = (styles) => (InnerComponent) => InnerComponent
```

StephanBijzitter:
[...]

PR #757, opened 2018-07-14
[WIP] Babel plugin for preprocessing

kof:
Issue https://github.com/cssinjs/jss/issues/579


packages/babel-plugin/src/index.js
```diff
@@ -0,0 +1,122 @@
+import {declare} from '@babel/helper-plugin-utils'
+import {create as createJss} from 'jss'
+
+const rawRuleName = '@raw'
+const defaultIdentifiers = ['createStyleSheet', 'injectSheet']
+
+export default declare(({types: t, ...api}, {identifiers = defaultIdentifiers, jssOptions}) => {
+  api.assertVersion(7)
```
HenriBeck: Should we maybe update our own babel packages to v7 as well? Would make testing easier.

kof: sure, if you get time, v6 and 7 are largely compatible

kof:
@cssinjs/core anyone wants to review this one? I am in the testing phase yet, but I want already start with making this code high quality readable

HenriBeck:
Can you rename the directory from `babel-plugin` to `babel-plugin-jss`?

packages/babel-plugin/src/testUtils.js
```diff
@@ -0,0 +1,14 @@
+import * as babel7 from '@babel/core'
```
HenriBeck: Maybe rename to just babel?

kof: I am working on tests to ensure babel6 works, so it will be needed

kof:
> Can you rename the directory from babel-plugin to babel-plugin-jss?

yep

kof:
> Also, can we have one standard for where the test files should be located? Because right now sometimes *.test.js is used and sometimes all of the tests are inside a tests folder.

There are 2 different testing scenarios. One is unit tests, where I actually test a specific module. The other one is more like an integration tests. In this case I 100% did integration tests and I had them all in index.test.js but then the amount of tests grew and file became unreadable, so I split them up in separate suits.

I think its ok to not have this consistency. If you are testing a specific module, then it should be module.test.js, otherwise if a test uses multiple modules it should go into __tests__. Both patterns are supported by jest by default without changing the config.

kof:
Remaining todo:
-  find a way to identify function calls without relying on a name, because this doesn't work after certain babel plugins, it be `something.default()`
- find a way to extract imported static variables
- evaluate module code in a separate context instead of `eval`
- add tests ensuring the proper variable scoping when evaluating (use same variable name to detect wrong scoping)
- use evaluation more instead of nodes traversing, seems like a more robust appraoch

packages/babel-plugin-jss/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "@babel/core": "^7.0.0-beta.54",
+    "babel-core": "^6.26.3",
+    "babel-plugin-syntax-decorators": "^6.13.0",
+    "jest": "^23.4.0"
+  },
+  "dependencies": {
+    "@babel/helper-plugin-utils": "^7.0.0-beta.53",
+    "babel-generator": "^6.26.1",
+    "babel-types": "^6.26.0",
+    "jss": "^9.8.7",
```
HenriBeck: Shouldn't jss be a peer dependency?

kof: seems valid

jahglow:
guys, any progress  on this? Like css extraction I mean...

PR #758, opened 2018-07-14
Moved jss-default-unit

HenriBeck:
Signed-off-by: Henri Beck <henribeck.dev@gmail.com>


HenriBeck:
Any suggestions for a name change? @kof

kof:
I think its also syntactic, so jss-plugin-syntax-default-unit

packages/jss-plugin-syntax-expand/.size-snapshot.json
```diff
@@ -8,5 +8,15 @@
     "bundled": 11240,
     "minified": 4432,
     "gzipped": 1544
+  },
+  "dist/jss-plugin-syntax-expand.js": {
```
kof: wrong name

kof: ah its a change from a different package here

HenriBeck: Yeah, forgot to build and it created them only now with the correct names.

PR #764, opened 2018-07-14
Added jss-vendor-prefixer

HenriBeck:
This fixes the tests🙌


kof:
jss-plugin-vendor-prefixer?

HenriBeck:
yeah, I waited for the name change because I wasn't quite sure about this one.

packages/jss-vendor-prefixer/src/index.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+    beforeEach(() => {
+      sheet = jss.createStyleSheet({
+        a: {animation: 'yyy'}
+      })
+    })
+
+    it('should generate correct CSS', () => {
+      if (isIEorEdge) {
+        expect(sheet.toString()).to.be('@keyframes a {\n}')
```
HenriBeck: Why will this be `keyframes` @kof? Doesn't make sense to me. This test fails in the BrowserStack tests but runs successfully locally.

packages/jss-vendor-prefixer/src/index.test.js
```diff
@@ -0,0 +1,165 @@
...
+  describe('@keyframes', () => {
+    let sheet
+
+    beforeEach(() => {
+      sheet = jss.createStyleSheet({
+        '@keyframes a': {}
+      })
+    })
+
+    it('should generate correct CSS', () => {
```
HenriBeck: Also, this one fails in the BrowserStack this test fails as well.

packages/jss-vendor-prefixer/src/index.test.js
kof: doesn't make sense to me, cc @AleshaOleg https://github.com/cssinjs/jss-vendor-prefixer/commit/739a34128c2d84395fc441093710bf1cedd4a9b8

packages/jss-vendor-prefixer/src/index.test.js
HenriBeck: Apparently, the keyframes aren't being prefixed in IE 9.0.0 (Windows 7.0.0), IE 10.0.0 (Windows 8.0.0), IE 11.0.0 (Windows 8.1.0.0), Edge 15.15063.0 (Windows 10.0.0)

HenriBeck: So they assertion is wrong.

kof: the assertion is not running, see early exit before on ln 24

kof: wait, early exit is only IE9, others should not add any prefixes afaik, cc @AleshaOleg how did the tests even pass?

HenriBeck: That's a different test. Two tests are failing right now

HenriBeck: I think the jss-vendor-prefixer plugin wasn't tested with BrowserStack and locally those tests succeed.

HenriBeck: Ohhh I know whats going on here...

The check for IE and Edge is supposed to be in this test and not in the other test.

That way the tests make sense...

HenriBeck: @kof tests seem to pass now

kof:
already approved

PR #766, opened 2018-07-14
Added jss-props-sort plugin

HenriBeck:
Renamed it to jss-plugin-props-sort

Removed jss plugins from dev dependencies


packages/jss-plugin-syntax-nested/src/index.test.js
```diff
@@ -2,7 +2,7 @@
 
 import expect from 'expect.js'
 import {stripIndent} from 'common-tags'
-import jssExtend from 'jss-extend'
+// import jssExtend from 'jss-plugin-syntax-extend'
```
HenriBeck: The problem is that jss-plugin-syntax-extend also has tests for jss-plugin-syntax-nested and this would create a dependency cycle. I turned off the tests for now. We might need some integration tests for the plugins which do not live in the actual plugin packages. So that the tests inside the plugins don't depend on other plugins.

Or we might just import the plugins needed for the tests directly.

PR #770, opened 2018-07-15
Removed DOM tests from function and observable plugin

HenriBeck:
I removed the DOM tests from the function and observable plugins because updating the DOM isn't done in those packages.

We could extract some testing utils into a private package, for example, the `createGenerateClassName` is being used in every package.

### DIFF
--- a/packages/jss-plugin-syntax-rule-value-function/src/function-rules.test.js
+++ b/packages/jss-plugin-syntax-rule-value-function/src/function-rules.test.js
@@ -3,9 +3,8 @@ import {stripIndent} from 'common-tags'
 
 import {create} from 'jss'
 import functionPlugin from './'
-import {createGenerateClassName, getStyle, getCss, removeWhitespace} from '../../jss/tests/utils'
 
-const settings = {createGenerateClassName}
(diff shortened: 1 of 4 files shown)


packages/jss-plugin-syntax-rule-value-function/src/function-rules.test.js
```diff
@@ -30,7 +28,7 @@ describe('jss-plugin-syntax-rule-value-function: Function rules', () => {
           {link: true}
         )
         .attach()
-      style = getStyle()
+      // style = getStyle()
```
kof: forgotten?

kof:
Lets see if any of those test scenario can be implemented in the dom tests from the core without using an observable, but just using rule.prop setter

kof:
if we have all prop() dom tests, then those dom tests are really unnecessary

HenriBeck:
Yeah, a plugin shouldn't worry about the DOM, I think.

packages/jss-plugin-syntax-rule-value-function/src/function-rules.test.js
HenriBeck: yes

PR #771, opened 2018-07-15
Add ES Modules support

HenriBeck:
- Added compiling to ES Modules with babel
- Upgraded to babel 7

Some babel 6 packages are still needed because of babel-plugin-rewire, once we moved the tests to jest we can use jest mocking functionality and remove babel-plugin-rewire


kof:
wait @TrySound was already working on esm build for the dist

TrySound:
Yes, I want to provide do a bundles instead. I can send a pr this evening.

HenriBeck:
Are there any advantages over a flat bundle? I assume you can load it in the browser as well, right?

TrySound:
A bunch of
- more controllable size and output
- slightly faster bundling
- faster browser loading if somebody will use it
- better debugging in development mode without sourcemaps (less webpack stuff)
- hidden internals which may cause breaking changes if user decide to use them

HenriBeck:
Okay, sounds great. 👍

PR #772, opened 2018-07-16
UMD distribution fixes

TrySound:
In this diff I refactored rollup config in favor of composition
instead of mixins.

- made createRule part of a public api; this allows to get rid from
interals usage in plugins
- peer dependencies are global in umd; this reduces plugins
size significantly
- changed development back to production mode in umd.min

### DIFF
--- a/packages/jss/src/index.js
+++ b/packages/jss/src/index.js
@@ -20,6 +20,11 @@ export {default as getDynamicStyles} from './utils/getDynamicStyles'
 export {default as toCssValue} from './utils/toCssValue'
 
 /**
+ * Create a rule instance.
+ */
+export {default as createRule} from './utils/createRule'
+
+/**
  * SheetsRegistry for SSR.
  */
 export {default as SheetsRegistry} from './SheetsRegistry'
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -5,7 +5,7 @@ import babel from 'rollup-plugin-babel'
 import replace from 'rollup-plugin-replace'
 import uglify from 'rollup-plugin-uglify'
 import {sizeSnapshot} from 'rollup-plugin-size-snapshot'
-import globals from 'rollup-plugin-node-globals'
+import nodeGlobals from 'rollup-plugin-node-globals'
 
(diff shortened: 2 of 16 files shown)


rollup.config.js
```diff
@@ -17,46 +17,77 @@ function toCamelCase(name) {
...
-      name: toCamelCase(pkg.name)
-    }
-  }),
-  Object.assign({}, base, {
-    plugins: [].concat(base.plugins, uglify()),
+      name,
+      globals
+    },
+    external: Object.keys(globals),
+    plugins: [
```
kof: I rewrote last time this thing to have same plugins in one place, because there were already problems, since one needs to do everything twice with the plugins and can be easily forgotten

can we please have just one place for the plugins?

TrySound: In that refactoring you replaced NODE_ENV with `development` for both umd and umd.min bundles. The case is that plugins can be different. It's cleaner to keep them separated and reuse only options.

I'm gonna add cjs and esm bundles in the next PR and their configs will be also different.

kof: oh, i just forgot to make different replace plkugins for both …  we will still have many same plugins though

TrySound: And I wouldn't add more plugins. These lists won't change for a long time.

kof: I bet someone will break next time one of the environments again if we don't put the same plugins with setup in one place

packages/jss/src/index.js
```diff
@@ -19,6 +19,11 @@ export {default as getDynamicStyles} from './utils/getDynamicStyles'
  */
 export {default as toCssValue} from './utils/toCssValue'
 
+/**
+ * Create a rule instance.
+ */
+export {default as createRule} from './utils/createRule'
```
kof: Its a public interface now, I wonder if we should document it

rollup.config.js
TrySound: In the bright future we will remove `commonjs` and `node-globals` plugins

packages/jss/src/index.js
TrySound: Could you fix that? I just don't know what it is.

PR #773, opened 2018-07-16
Generate cjs and umd flat bundles

TrySound:
A few reasons to bundle

- more controllable size and output
- slightly faster bundling
- faster browser loading if somebody will use it
- better debugging in development mode without sourcemaps (less webpack stuff)
- hidden internals which may cause breaking changes if user decide to use them


kof:
> Guys, let's use squash and merge button. It provides less noise in commit history.

I am for it, usually I simply forget to press it lol

TrySound:
Me too. There should be an option in project settings to restrict with only that button. Could you try it?

kof:
Cool, didn't know it was an option, just activated it

HenriBeck:
Looks good👍

What is the code property inside the size snapshot files? And why is there a difference between rollup and webpack?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

TrySound:
`code` is the size of treeshaked and minified code
`import_statements` is the size of import statements in the code
So when both this values are equal we have perfect treeshakability. Import statements just considered as side effectful and not treeshaked.

Webpack is different because its treeshaking abilities are less perfect than rollup ones + webpack adds runtime and doesn't have esm output.

PR #774, opened 2018-07-16
Upgrade babel to 7 beta for build

TrySound:
Size is slightly increased because of new helpers


package.json
```diff
@@ -34,10 +34,14 @@
     "node": ">=4"
   },
   "devDependencies": {
+    "@babel/core": "^7.0.0-beta.53",
+    "@babel/plugin-proposal-class-properties": "^7.0.0-beta.53",
+    "@babel/preset-env": "^7.0.0-beta.53",
+    "@babel/preset-flow": "^7.0.0-beta.53",
     "babel-cli": "^6.18.0",
```
HenriBeck: Do we still need the babel-cli?

package.json
```diff
@@ -34,10 +34,14 @@
...
-    "babel-core": "^6.20.0",
+    "babel-core": "^6.26.3",
     "babel-eslint": "^8.2.3",
-    "babel-loader": "^6.2.9",
+    "babel-loader": "^7.1.5",
     "babel-plugin-external-helpers": "^6.22.0",
```
HenriBeck: We also don't need the external helpers plugin and the inline version anymore.

HenriBeck: Also, I think we should be able to remove babel-plugin-transform-es3-member-expression-literals, babel-plugin-transform-es3-property-literals, babel-plugin-transform-flow-strip-types and the old babel preset packages

TrySound: I know. I left everything for tests yet

package.json
TrySound: Nope

Issue #777, opened 2018-07-18
Expose public typings for plugins and abstractions like react-jss


kof:
For plugins and for e.g. react-jss we need the typings and its better to reuse them rather than duplicating.

Exporting all typings we have in the core would expose many private things, mb we should export only typings that we know are needed.

kof:
Subscribe #795 for updates about release date.

PR #778, opened 2018-07-18
Added exporting types from jss

HenriBeck:
#777

### DIFF
--- a/packages/jss/src/index.js
+++ b/packages/jss/src/index.js
@@ -7,7 +7,39 @@
  * @license MIT
  */
 import Jss from './Jss'
-import type {JssOptions} from './types'
+import StyleSheet from './StyleSheet'
+import type {
+  JssOptions,
+  StyleSheetOptions,
+  JssStyle,
+  Plugin,
+  generateClassName,
+  Rule,
+  RuleOptions,
+  Classes,
+  Renderer,
+  RuleListOptions,
+  JssValue
+} from './types'
+
+/**
+ * Export types for better typing inside plugins and integrations.
+ */
+export type {
+  StyleSheetOptions,
+  JssValue,
+  JssOptions,
+  JssStyle,
+  Plugin,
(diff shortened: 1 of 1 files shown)


packages/jss/src/index.js
```diff
@@ -7,7 +7,39 @@
...
+  Plugin,
+  generateClassName,
+  RuleListOptions,
+  Rule,
+  Renderer,
+  RuleOptions,
+  Classes,
+  Jss,
+  StyleSheet
+}
```
TrySound: Does it work?
```
export type {} from './types';
```

HenriBeck: Will try it out.

HenriBeck: Yes, it works. Good catch 👍

packages/jss/src/index.js
```diff
@@ -7,8 +7,27 @@
...
+ * Export types for better typing inside plugins and integrations.
+ */
+export type {
+  StyleSheetOptions,
+  JssValue,
+  JssOptions,
+  JssStyle,
+  Plugin,
+  generateClassName,
+  RuleListOptions,
```
kof: Is this one needed somewhere?

packages/jss/src/index.js
```diff
@@ -7,8 +7,27 @@
...
+  Rule,
+  Renderer,
```
kof: Is this one needed?

packages/jss/src/index.js
HenriBeck: Maybe, I don't know. It might be needed at some time especially because we export RuleList.
Also, I would like to export those now and not have to add exports every time later on.

packages/jss/src/index.js
HenriBeck: That depends on wether or not we allow people to write there own renderers?

PR #779, opened 2018-07-18
Added typing to the cache plugin

HenriBeck:
@kof could it be that the `RuleOptions` is missing the `parent` property because we are accessing it in the plugin. That's also why the CI is failing.


HenriBeck:
I think the third option is the RuleOptions but the parent is missing. I think it's supposed to get the InternalRuleOptions.

kof:
Ok I checked, parent is passed as an option only in ConditionalRule and KeyframesRule and RuleOptions is simply missing it

PR #780, opened 2018-07-18
Added typing to the isolate plugin

HenriBeck:
Same problem with the `RuleOptions` as in #779.

Also, I'm importing the type of the StyleRule from the src. We should maybe export the Rules as well as they are needed for better typing in some of the plugins.

### DIFF
--- a/packages/jss-plugin-isolate/src/index.js
+++ b/packages/jss-plugin-isolate/src/index.js
@@ -39,15 +39,10 @@ const getStyle = (option = 'inherited') => {
   return inheritedInitials
 }
 
-const ignoreParents = {
-  keyframes: true,
-  conditional: true
-}
-
 const shouldIsolate = (rule: StyleRule, sheet, options) => {
   const {parent} = rule.options
 
-  if (parent && typeof parent.type === 'string' && ignoreParents[parent.type]) {
+  if (parent && (parent.type === 'keyframes' || parent.type === 'conditional')) {
     return false
   }
 
--- a/packages/jss-plugin-isolate/src/index.js
+++ b/packages/jss-plugin-isolate/src/index.js
@@ -47,7 +47,7 @@ const ignoreParents = {
 const shouldIsolate = (rule: StyleRule, sheet, options) => {
   const {parent} = rule.options
 
-  if (parent && ignoreParents[parent.type]) {
+  if (parent && typeof parent.type === 'string' && ignoreParents[parent.type]) {
     return false
   }
 
--- a/packages/jss-plugin-isolate/src/index.js
+++ b/packages/jss-plugin-isolate/src/index.js
@@ -1,5 +1,13 @@
+// @flow
 import inheritedInitials from 'css-initials/inherited'
 import allInitials from 'css-initials/all'
+import type {Plugin} from 'jss'
+import type StyleRule from 'jss/src/rules/StyleRule'
+
+type Options = {
+  isolate?: boolean | string,
+  reset?: 'all' | 'inherited' | Object | ['all' | 'inherited', Object]
+}
 
 const resetSheetOptions = {
   meta: 'jss-isolate',
@@ -37,7 +45,7 @@ const ignoreParents = {
   conditional: true
 }
 
-const shouldIsolate = (rule, sheet, options) => {
+const shouldIsolate = (rule: StyleRule, sheet, options) => {
   const {parent} = rule.options
 
   if (parent && ignoreParents[parent.type]) {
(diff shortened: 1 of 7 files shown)


packages/jss-plugin-isolate/src/index.js
```diff
@@ -1,5 +1,13 @@
+// @flow
 import inheritedInitials from 'css-initials/inherited'
 import allInitials from 'css-initials/all'
+import type {Plugin} from 'jss'
+import type StyleRule from 'jss/src/rules/StyleRule'
```
kof: I think we should stop importing from non-public types, since we decided to export types. In this case we should either as well export all rule types or user the Rule type https://github.com/cssinjs/jss/blob/master/packages/jss/src/types/jss.js#L19

HenriBeck: Yeah, but I need to say that explicitly it's a StyleRule, so we need to export the rules as well

HenriBeck: https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-syntax-rule-value-observable/src/index.js#L4 we have the same problem here

kof: Then we have to export them all I guess

packages/jss-plugin-isolate/src/index.js
```diff
@@ -48,7 +47,7 @@ const ignoreParents = {
 const shouldIsolate = (rule: StyleRule, sheet, options) => {
   const {parent} = rule.options
 
-  if (parent && ignoreParents[parent.type]) {
+  if (parent && typeof parent.type === 'string' && ignoreParents[parent.type]) {
```
kof: would it work with a `type in parent` or just `parent.type` check? I wonder if string check is required

HenriBeck: parent.type doesn't work. Flow has a problem resolving the type for the `parent.type` and that's why the string check is necessary.

kof: type in parent too?

kof: The only place where type is not defined is when parent is a StyleSheet, in theory we could just typecast to a Rule, so this check won't be needed. Its a shame we have to add runtime checks for the sake of types

HenriBeck: We could also do this:

```js
  let {parent} = rule.options
  
  parent = ((parent: any): ConditionalRule | KeyframesRule)

  if (parent && ignoreParents[parent.type]) {
    return false
  }
```

kof: Yes, thats what I ment with casting, not sure how its called properly, I think its better, we basically manually tell the type system what it is, so it doesn't throw

HenriBeck: This works as well:
```js
  if (parent && (parent.type === 'keyframes' || parent.type === 'conditional')) {
    return false
  }
```

kof: types get removed from the runtime, these checks - don't

HenriBeck: Yeah but the above solution will result in `result = result`, or is it also being removed? When we check against the strings we don't need the type casting

kof: `parent = ((parent: any): ConditionalRule | KeyframesRule)` would becompletely removed afaik, whatever check you do in actual js code will not

kof: oh actually I am not sure if its gonna be removed after flow types are removed, but for sure after uglify

HenriBeck: Agree with you. That's why I would go for `parent.type === 'keyframes' || parent.type === 'conditional'` because this doesn't require the typecasting and we have two property access operations anyway.

kof: Huh, I thought actually that you would add with this 2 runtime checks, but with this its basically as it was before, just typechecked `result=result` check only in non-minified version

```
  parent = ((parent: any): ConditionalRule | KeyframesRule)

  if (parent && ignoreParents[parent.type]) {
```

kof: oh wait, you are going to replace with this ignoreParents[parent.type] checks ... yes thats not adding anything its 2 checks vs. 2 checks

kof: yeah it seems better

PR #781, opened 2018-07-19
Added typings to props-sort plugin

HenriBeck:


packages/jss-plugin-props-sort/src/index.js
```diff
@@ -11,8 +14,8 @@ export default function jssPropsSort() {
 
     const newStyle = {}
     const props = Object.keys(style).sort(sort)
-    for (const prop in props) {
-      newStyle[props[prop]] = style[props[prop]]
+    for (const prop of props) {
```
kof: this generates a lot of code, which is an overkill for a library: https://babeljs.io/repl/#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYewTgBAFAxiB2BnALhADmEaImOzaiAlBAN4BQEV8ApgO4DKyAngDY0DaGWAuhALwQUbTtzQ9yAXyA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&sourceType=module&lineWrap=false&presets=es2015%2Ces2016%2Ces2017%2Cstage-0&prettier=false&targets=&version=6.26.0&envVersion=

HenriBeck: Didn't know it generated that much code. I replaced it with a forEach now. The reason I changed it from a `for ... in` is because it's not supposed to be used with arrays.

kof:
The only problem with for in on arrays is this https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea which does not apply to our case, because we we just called Object.keys before.

I didn't use any callback based iterators in hot plugins because they are much slower than for loops on a micro perf level, which is kinda important for the plugins.

HenriBeck:
I replaced it with a normal for loop as it should be faster than a for-in loop and flow doesn't complain anymore.

kof:
True, much faster https://jsperf.com/for-loop-vs-for-in-loop

PR #783, opened 2018-07-19
Added typing to the compose plugin

HenriBeck:


packages/jss-plugin-syntax-compose/src/index.js
```diff
@@ -26,7 +28,9 @@ function registerClass(rule, className) {
     return registerClass(rule, className.split(' '))
   }
 
-  const {parent} = rule.options
+  const {parent} = ((rule.options: any): {parent: StyleSheet})
+
+  // parent = ((parent: any): StyleSheet)
```
HenriBeck: @kof which syntax should we use? The second one requires the parent to declare it with let.

PR #784, opened 2018-07-20
Add typings to global plugin

HenriBeck:
- Update return type of onCreateRule hook for plugins to only return a BaseRule interface
- Added a ParentRule interface which specifies rules to be a RuleList
- Made the parent prop in RuleOptions and RuleListOptions to be of type ParentRule
- Updated interfaces of ConditionalRule and KeyframesRule to be ParentRule

### DIFF
--- a/packages/jss-plugin-syntax-global/src/index.js
+++ b/packages/jss-plugin-syntax-global/src/index.js
@@ -1,12 +1,24 @@
-import {RuleList} from 'jss'
+// @flow
+import {
+  RuleList,
+  type Plugin,
+  type RuleOptions,
+  type ParentRule,
+  type StyleRule,
+  type BaseRule
+} from 'jss'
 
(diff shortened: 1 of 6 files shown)


packages/jss-plugin-syntax-global/src/index.js
```diff
@@ -1,12 +1,24 @@
...
+  type ParentRule,
+  type StyleRule,
+  type BaseRule
+} from 'jss'
 
 const propKey = '@global'
 const prefixKey = '@global '
 
-class GlobalContainerRule {
+class GlobalContainerRule implements ParentRule {
```
kof: Mb instead of ParentRule call it ContainerRule?

HenriBeck: yeah that's better 👍

PR #785, opened 2018-07-20
Moved react-jss into monorepo

HenriBeck:
- ESLint is broken because react-jss uses globals inside the tests
- Tests are not run yet


kof:
do you want to merge this state?

HenriBeck:
No definitely not, just created it so @gomesalexandre could recreate the PR for the types

HenriBeck:
I added the react-jss to the test files, and when I run them, some of them fail because there are always exactly two style tags too many. Weirdly enough those are inside the body and aren't being removed from the dom

kof:
Sounds like some clean up task is not done, you can easily find that out by excluding tests, finding the right once and then looking why there is no cleanup

HenriBeck:
the tests run locally but don't in the CI and that's because we use sinon inside react-jss tests which doesn't support some of the browsers that we are testing with

HenriBeck:
for now, yeah

HenriBeck:
@kof what should we do? Once we switched to jest this won't be a problem anymore.

kof:
What exactly is failing?  I checked, we are only using the spy function, it can't be the reason.

HenriBeck:
yes, but we are including the whole sinon libary and I don't think it's tree shaking it.

HenriBeck:
Mobile Safari 9.0.0 (iOS 9.3.0) ERROR
  SyntaxError: Unexpected token '>'
  at webpack:///~/sinon/pkg/sinon-esm.js:5212:0 <- packages/react-jss/src/injectSheet.test.js:40905

I just noticed that we need to include a raf polyfill

kof:
That is strange, tried upgrading the version? Token ">" is potentially the arrow function? Do you see whats there? I tried to find the line on cdn version but it points me to something else. Did you try to just upgrade to the latest version?

HenriBeck:
It should be on the latest version because I installed it manually with the other missing dev dependencies.

HenriBeck:
yeah, it's an arrow function.

kof:
Lets try to raise the issue on sinon, they seem to be actively developing it. If we can't fix it quickly there, we only use spy, so we can just replace it, but I would rather keep it for the sake of using the right tool for the job.

HenriBeck:
I don't think they will do anything about it because they don't support that browser anymore.

kof:
btw. you are using esm build, mb there are others?

HenriBeck:
I think webpack is using esm, we really don't have any control over that.

kof:
http://sinonjs.org/releases/v6.1.4/


There should not be any issues with using Sinon v6.1.4 in newer versions of the same runtimes.

If you need to support very old runtimes that have incomplete support for ES5.1 you might get away with using loading es5-shim in your test environment. If that fails, we recommend getting a legacy releases of Sinon.

kof:
It seems that the Set polyfill is still missing in the ci build

HenriBeck:
Now the tests are failing because of formatting issues. But the tests run successfully locally.

HenriBeck:
@kof any ideas on why the tests are failing?

kof:
Some browser apply different formatting rules, in jss core I was stripping all whitespaces just for the sake of being able to compare.

kof:
🎉🎉🎉

PR #788, opened 2018-07-23
Added typings to the nested plugin

HenriBeck:


packages/jss-plugin-syntax-nested/src/index.js
```diff
@@ -52,6 +58,7 @@ export default function jssNested() {
     // Options has been already created, now we only increase index.
     if (options) return {...options, index: options.index + 1}
 
+    // $FlowFixMe: Flow can't know that nestingLevel actually may exist here
     let {nestingLevel} = rule.options
```
kof: are we missing a typo? Mb we should add it?

HenriBeck: The `nestingLevel` only added by the nested plugin. So I think it shouldn't be in the core

kof: Right, in that cae we need jss-nested own rule options obejct which extends the original one, but I am also ok withnot doing it now

HenriBeck: Well we would need to extend the rule because we are accessing the options on the rule and not the options arg which was passed

kof: Yes we would have to typecast over any in that case

PR #794, opened 2018-07-30
Added typings to the expand plugin

HenriBeck:


HenriBeck:
Didn't find any good solutions to the problem

PR #810, opened 2018-08-03
Added typing to expand plugin

HenriBeck:
Fixed typing of the `prop` method in `StyleRule` to allow null and false as values

### DIFF
--- a/packages/jss/src/rules/StyleRule.js
+++ b/packages/jss/src/rules/StyleRule.js
@@ -69,7 +69,7 @@ export default class StyleRule implements BaseRule {
   /**
    * Get or set a style property.
    */
-  prop(name: string, value?: JssValue): StyleRule | string {
+  prop(name: string, value?: JssValue | null | false): StyleRule | string {
     // It's a getter.
     if (value === undefined) return this.style[name]
 
--- a/packages/jss/src/rules/StyleRule.js
+++ b/packages/jss/src/rules/StyleRule.js
@@ -69,7 +69,7 @@ export default class StyleRule implements BaseRule {
   /**
    * Get or set a style property.
    */
-  prop(name: string, value?: JssValue | null | false): StyleRule | string {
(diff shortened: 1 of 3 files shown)


packages/jss/src/rules/StyleRule.js
```diff
@@ -69,7 +69,7 @@ export default class StyleRule implements BaseRule {
   /**
    * Get or set a style property.
    */
-  prop(name: string, value?: JssValue): StyleRule | string {
+  prop(name: string, value?: JssValue | null | false): StyleRule | string {
```
kof: didn't know we support null/false there

kof: value is optional because .prop can be used as a getter if value is not passed, but we have an explicit === undefined check, so null and false would act as a setter, so you would land it in the CSSOM

HenriBeck: Yeah, null and false are used to remove a value if necessary, that's why I added it.

kof: Oh wait, true

kof: Should we evtl allow them in JssValue directly? At render time empty values are getting removed anyways, I believe

HenriBeck: The only real reason to use them directly in the styles definition is with function values to get those styles removed. That's why I would tend not to allow them in JssValue as the user should only use those for the function values.

```js
const styles = {
  // There is no reason for allowing display: false here
  container: { display: false }
};
```

kof: Thats the only obvious reason, but sometimes people use functions with logic which may accidentally return also a null: for eg. `{color: getColor()}` and getColor in some specific case returns a null. So the question would be independent of the .prop method, if you get a full typing power, this is actually allowed, since we handle this scenario, right?

PR #811, opened 2018-08-03
Added typing to jss-preset-default

HenriBeck:
I had to manually export and import the options because Flow doesn't seem to know how to extract function parameters sadly.


kof:
Let me know when I should look into this PR, seems like a WIP because you removed the options?

HenriBeck:
Well, I removed the unnecessary options because most plugins don't have options

kof:
Oh right and now flow is complaining …

PR #812, opened 2018-08-05
Added jss keyframes helpers package

HenriBeck:
What this package does:
 - Create keyframes with a unique id
 - Reuse stylesheets for the same jss instance
 - Support multiple jss instances
 - Server-side rendering support by collecting the stylesheets in a sheet registry


kof:
The benefits of having it this way

```
  '@keyframes fadeIn': {
    from: {opacity: 0},
    to {opacity: 1}
  }
})

sheet.getRule('@keyframes fadeIn').id // @keyframes fadeIn-12378676134
// or
sheet.keyframes.fadeIn // @keyframes fadeIn-12378676134
```
1. familiar syntax
2. no separate style sheet required

kof:
What would be a benefit for the approach in this PR?

HenriBeck:
I would say this is also a familiar syntax especially for people coming from `styled-component` and `emotion`. And I find it a little bit easier for new users to reason about. They don't need to know how it works under the hood, they just get a string back which they can use instead of getting the generated name on their own.

Another problem I see with the above API is that: How would you use the keyframe in the same stylesheet?

kof:
Yeah, I see familiarity from other cssinjs libs as well as a point.

> How would you use the keyframe in the same stylesheet?

Same like referencing rules in jss-nested - $keyframe. We even saw someone assuming this works this way already for some reason.

kof:
Another point: the way keyframes work currently is in conflict with core philosophy - everything scoped by default, I think this needs to be fixed in any case.

kof:
Mb we should have both things in place? The API I suggested built in in the core + your utility function as a package, which will leverage the core implementation. As an optional utility.

packages/jss-keyframes/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+const stylesheets = new WeakMap()
+
+function getStyleSheet(jss: Jss) {
+  const existingStylesheet = stylesheets.get(jss)
+
+  if (existingStylesheet) {
+    return existingStylesheet
+  }
+
+  const stylesheet = jss.createStyleSheet({}, {meta: 'JSS Keyframes'})
```
kof: wouldn't it make sense to just have one style sheet?

kof: In case one wants to render that style sheet in a specific place in the head, like all jss style sheets, there should be a way to pass other options.

packages/jss-keyframes/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+import defaultJss, {SheetsRegistry, type Jss} from 'jss'
+import warning from 'warning'
+
+type Options = {
+  jss?: Jss,
+  name?: string
+}
+
+let counter = 0
+const registry = new SheetsRegistry()
```
kof: in case of react-jss and actually even in case with working with just core, adding another registry complicates things, since one needs to be aware of a separate registry for the keyframes and not forget to ssr it properly?

packages/jss-keyframes/src/index.js
```diff
@@ -0,0 +1,52 @@
...
+export default function createKeyframes(keyframes: {}, options: Options = {}) {
+  counter++
+
+  warning(
+    counter <= 10000,
+    'You might have a memory leak because the keyframes counter grew up to 10000'
+  )
+
+  const {name = 'animation', jss = defaultJss} = options
+  const keyframesName = process.env.NODE_ENV === 'production' ? `c${counter}` : `${name}-${counter}`
```
kof: we could actually leverage generateClassName function from the core, since if user wants to customize it, now he has to also take care of this one.
I regret now for not having called it generateId or something more generic, but maybe its not too bad if we use it here as well, even though its not a css class name what we generate.

kof:
Another thing is suboptimal: if you use react-jss, you assume only styles that are used also generated. Now with this function you will generate all keyframes at once, so during ssr you will either render them all or none at all.

I am not sure it even makes sense to use keyframes in ssr code, but who knows, we don't know if there is a good use case for it.

kof:
An option could be that keyframes() doesn't render anything, it "creates" the keyframes, but to render them one needs to pass its result with the styles:

```js
const keyframesId = keyframes({
  from: {transform: 'scale(1)'},
  to: {transform: 'scale(0)'}
})

createStyleSheet({
  '@keyframes': keyframesId,
  button: {
    animation: `${keyframesId} ...`
  } 
})
```

Not sure though if its not too complex for the user.

packages/jss-keyframes/src/index.js
HenriBeck: yeah, the idea behind that was to support multiple versions of jss

HenriBeck: But I have an idea on how to actually allow the user to pass options to the sheet.

kof: Yes, its possible, but its still a complication.

HenriBeck:
Yeah, I think this might be a little bit to complex, and you can't use two keyframes in the same sheet.

kof:
> you can't use two keyframes in the same sheet.

you can for e.g. like this:

```js
createStyleSheet({
  '@keyframes': [keyframesId],
  button: {
    animation: `${keyframesId} ...`
  } 
})
```

But still not sure about complexity. Benefit is really that you render it in a specific style sheet and in case of react-jss, it can control the injection and sheets registry.

HenriBeck:
Yeah, that seems too complicated.

Let's do this for the core maybe:
```js
  '@keyframes fadeIn': {
    from: {opacity: 0},
    to {opacity: 1}
  },

  button: {
    animation: '$fadeIn',
  },
})
```

This should be easier for users who already use jss-nested with referencing rules.

kof:
Yeah, also if one really wants a keyframes() fn, it would be basically (pseudocode): 

```js
const sheet = jss.createStyleSheet().attach()
export default (styles) = sheet.addRule('@keyframes x', styles).id
```

HenriBeck:
The only problem is with the `jss-nested` like syntax is that it would be a breaking change.

kof:
I am not really in favor of increasing the core size, but in this case idk, keyframes is core css functionality. An option would be to extract those more rarely used rules in separate packages and include them in the default preset, so in case someone doesn't need them, no need to load them in the main bundle. Also we could show case a smaller core  haha

kof:
> The only problem is with the jss-nested like syntax is that it would be a breaking change.

Yes, but since we are releasing v10, we can introduce a breaking change

HenriBeck:
Okay, let's maybe go with both solutions and give the user a choice. We might need to write some guide where we describe the pros and cons of both solutions.

Keyframe Helper:
Pros: Reusability even between different StyleSheets

Scoped Keyframes:
Pros: Only available to the current StyleSheet, no leakage (reuse the same animation name in different sheets)

kof:
Would you create the helper in jss package or separately?

kof:
Btw. there is also another way https://github.com/Khan/aphrodite#animations 

Basically instead of using @keyframes directly, using keyframes definition object as an animationName value

HenriBeck:
I would leave the helper in a separate package because not everyone will actually need this.

About the integration into react-jss: I don't think there is an easy way of doing it. That's why the package is exposing a `setRegistry` function to add the created sheets to a registry. For Server-Side Rendering, you would need to Registry (one for the keyframes with can be the same across all requests and one for the JssProvider). 

About the way Aphrodite does animations, I'm not a big fan of it because of those downsides. Also, I think the referencing style `$fadeIn` is more suitable because users already use it and the migration would be relatively simple.

HenriBeck:
But I would maybe rename the package from `jss-keyframes` to `jss-keyframes-helper` maybe.

kof:
Mb react-jss should have this helper integrated and already work with the registry out of the box?

kof:
jss-keyframes-helper sounds good

HenriBeck:
I don't think we can make a perfect integration into react-jss and server-side rendering because the registry is being passed in the context and I don't see an easy way on how we would access it.

kof:
Yeah, this isn't quite easy, mb we don't need the helper package standalone after all, but built in react-jss somehow

kof:
React-JSS needs to be able to control the registry from keyframes helper

HenriBeck:
I don't think it's a good idea to built this feature into react-jss directly because it should be usable just with jss core. The main benefit of this package is the possibility of sharing the keyframes between multiple sheets with random name generation which should not be tied to react / integration in my opinion.

karma.conf.js
```diff
@@ -41,7 +41,8 @@ module.exports = config => {
       './packages/react-jss/src/index.test.js',
       './packages/react-jss/src/compose.test.js',
       './packages/react-jss/src/JssProvider.test.js',
-      './packages/react-jss/src/injectSheet.test.js'
+      './packages/react-jss/src/injectSheet.test.js',
+      'packages/jss-keyframes-helper/src/index.test.js'
```
kof: inconsistent path

packages/jss-keyframes-helper/src/index.js
```diff
@@ -26,30 +26,29 @@ function getStyleSheet(jss: Jss, registry: SheetsRegistry | null, sheetOptions):
...
 
+  // Attach the stylesheet to the DOM
   stylesheet.attach()
 
   return stylesheet
 }
 
-export function setRegistry(registry: SheetsRegistry | null) {
-  sheetsRegistry = registry
+export function resetRegistry() {
```
kof: could export the registry directly

packages/jss-keyframes-helper/src/index.js
```diff
@@ -8,10 +8,10 @@ type Options = {
   registry?: SheetsRegistry
 }
 
-let sheetsRegistry = null
+const sheetsRegistry = new SheetsRegistry()
 const stylesheets = new WeakMap()
```
kof: we can get rid of this stylesheets map, it is bad because you will accumulate during ssr. 
For e.g. you could use sheetsRegistry.registry.indexOf instead

kof: second thought: why not just use a single style sheet?

packages/jss-keyframes-helper/src/index.js
```diff
@@ -0,0 +1,50 @@
...
+import defaultJss, {SheetsRegistry, type Jss, type StyleSheet, type StyleRule} from 'jss'
+
+type Options = {
+  jss?: Jss,
+  name?: string,
+  sheet?: StyleSheet,
+  registry?: SheetsRegistry
+}
+
+export const sheetsRegistry = new SheetsRegistry()
```
kof: I would seriously consider to use always one styleSheet, generate and attach it lazily once createKeyframes is called the first time and export it

HenriBeck: I added the multiple stylesheets for different jss versions. They are lazily attached and reused between the same jss instance though.

HenriBeck:
@kof can I merge this?

packages/jss-keyframes-helper/README.md
```diff
@@ -0,0 +1,80 @@
+# jss-keyframes-helper
+
+Helper for generating keyframes with a unique id in jss
```
kof: JSS

packages/jss-keyframes-helper/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Installation
+
+> yarn add jss-keyframes-helper
```
kof: we had always npm in docs, should we stay consistent with this?

packages/jss-keyframes-helper/README.md
```diff
@@ -0,0 +1,80 @@
...
+  // This will not add the stylesheet to the internally used SheetsRegistry of this package.
+  // You will need to add it to your own SheetsRegistry when doing SSR.
+  sheet: StyleSheet,
+}
+```
+
+## SSR with React
+
+Because we may create multiple stylesheets for different jss instances, we add those to an internal SheetsRegistry which you can import.
+Keyframes should be static, so you don't need to recreate them for every requests and reuse them between requests.
```
kof: What if they want to use fn values inside of them? Not sure if it even works.

packages/jss-keyframes-helper/README.md
HenriBeck: Yes, just wrote yarn because I mostly use yarn

packages/jss-keyframes-helper/README.md
HenriBeck: Ohh yeah, function values do not work with this package, though I feel like you usually don't need function values inside keyframes.

kof:
I am still trying to see the value of having the helper function vs implementing it in the stylesheet with namespaces.

kof:
Are you ok with not merging this at all and instead introduce scoping in the core?

HenriBeck:
For me the pros are:
- Easier migration/learning curve from `styled-components`
- The user doesn't need to know anything about `jss.createStyleSheet` to create reusable animations between multiple sheets

I think the scoped keyframes is a great idea when you want to use the keyframes in the same sheet., but this API is much simpler for reusing keyframes between different stylesheets.

Having said that, this was just an idea for a package based on some `styled-components` API (which I think a lot of people are familiar with) and some requests for scoping keyframes better. Not merging it is fine with me.

kof:
The difference with SC is that they run a single stylesheet mode, which means user doesn't need to do a separate step to SSR keyframes. Also they don't support multiple versions, so user doesn'T have to have the jss and sheet option. 

It becomes way more complex than the one in SC.

kof:
If we could provide this function in a way that would have no need for all the options, I would be for it.

HenriBeck:
Well, we could remove the options for `sheet` and `jss`, though that would hardly limit the user.

This package also runs single stylesheet mode when no extra config etc. is used.

The problem with SSR is that there is no way to access the `SheetsRegistry` from the `JssProvider`.
Also, the keyframes are only called once, and not on every render.

kof:
Pretty sure the next issue would be to allow function values in keyframes …

HenriBeck:
Well, you can't use function values with this syntax. This syntax solves using keyframes between multiple sheets with random name generation. Supporting function values would be a nightmare.

The scoped keyframes are more suited for this.

In general:
If you need/want to reuse keyframes between sheets, use the keyframe helper.
If you only need the keyframes in the same sheet or function values inside use the scoped keyframes.

kof:
yeah, that could be a use case, evtl though the DX difference between this and createStyleSheet is not big enough to add a separate utility function.

PR #816, opened 2018-08-06
fix(jss): Wrong sheet order when the insertion point is a DOM element.

### LINKED ISSUE #752: Wrong styles order by DomRenderer
This bug was found and described here: https://github.com/cssinjs/react-jss/issues/278

Demo project: https://github.com/bungu/jss-wrong-style-order-example

If you define custom insertion point as DOM element (`typeof element.nodeType === 'number'`) and insertion point is the last child of `<head>`, then sheets with same highest index (except first) will be rendered at insertion point, instead of first highest sheet.

I'm not sure how to write test for this bug properly, so I made simple demo project.

electrikdevelopment:
When a new sheet is added with an index higher than all of those that already exist and the insertion point is an element the style sheets are actually added to the top of the list of sheets instead of the bottom.

The logic for inserting into the head vs inserting into a different element should be the same but currently isn't.

The modified test fails without the change and passes after.

This bug is particularly impactful when using the ShadowDOM as styles need to be added to an element within the shadow tree instead of the head and currently prevents using material-ui components inside a shadow root.

Fixes #752

### DIFF
--- a/packages/jss/src/renderers/DomRenderer.js
+++ b/packages/jss/src/renderers/DomRenderer.js
@@ -263,7 +263,7 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
     // https://stackoverflow.com/questions/41328728/force-casting-in-flow
     const insertionPointElement: HTMLElement = (insertionPoint: any)
     const {parentNode} = insertionPointElement
-    if (parentNode) parentNode.insertBefore(style, insertionPointElement.nextSibling)
+    if (parentNode) parentNode.insertBefore(style, prevNode)
     else warning(false, '[JSS] Insertion point is not in the DOM.')
     return
   }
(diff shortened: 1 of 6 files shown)


packages/jss/src/renderers/DomRenderer.js
```diff
@@ -263,7 +263,7 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
     // https://stackoverflow.com/questions/41328728/force-casting-in-flow
     const insertionPointElement: HTMLElement = (insertionPoint: any)
     const {parentNode} = insertionPointElement
-    if (parentNode) parentNode.insertBefore(style, insertionPointElement.nextSibling)
+    if (parentNode) parentNode.insertBefore(style, prevNode)
```
HenriBeck: `prevNode` will always be `null` here.

electrikdevelopment: Indeed it will, as will it in the "getHead().insertBefore" line below. Perhaps using appendChild for both is more descriptive?

electrikdevelopment:
ok i'll look at adding a few more test cases and a different fix over the next week.

HenriBeck:
Your current fix should be reverted. Also `nextElementSibling` is actually fine in line 232

HenriBeck:
Solved by #819

PR #818, opened 2018-08-07
[react-jss] Add flow typings

gomesalexandre:
This is a work-in-progress PR, this first commit is pretty much just the work done before the switch to monorepo. 

TODO:

- 100% flow coverage for react-jss
- adding script for copying source files


package.json
```diff
@@ -99,5 +99,8 @@
     "sinon": "4.5.0",
     "webpack": "^2.3.3",
     "zen-observable": "^0.6.0"
+  },
+  "dependencies": {
+    "flow-copy-source": "^2.0.2"
```
HenriBeck: We don't copy the source files. We already do the step to use flow typings.

packages/react-jss/.flowconfig
```diff
@@ -0,0 +1,11 @@
+[ignore]
```
HenriBeck: This file shouldn't exist because there is a flow config in the root for all packages

packages/react-jss/package.json
```diff
@@ -33,7 +33,8 @@
     "decorator"
   ],
   "scripts": {
-    "prepare": "node ../../scripts/prepare.js"
+	"prepare": "node ../../scripts/prepare.js",
+	"flow": "flow"
```
HenriBeck: Again, we don't need to run flow here as it's being run from the root directory

packages/react-jss/src/createHoc.js
```diff
@@ -32,7 +34,19 @@ const dynamicStylesNs = Math.random()
  *
  */
 
-const getStyles = (stylesOrCreator, theme) => {
+type Props = {
+  classes: ?{},
+  innerRef?: () => {},
```
HenriBeck: should be `(comp: ComponentType<{}> | null) => void`. You could make the component type more exact with moving the props into the generic types of `createHOC`

packages/react-jss/src/createHoc.js
```diff
@@ -119,9 +138,9 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
       }
     }
 
-    componentDidUpdate(prevProps, prevState) {
+    componentDidUpdate(prevProps: Props, prevState: State) {
       // We remove previous dynamicSheet only after new one was created to avoid FOUC.
-      if (prevState.dynamicSheet !== this.state.dynamicSheet && prevState.dynamicSheet) {
```
HenriBeck: ` && prevState.dynamicSheet` should not be removed!

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+import type {
+  InternalStyleSheetOptions,
+  Rule,
+  ToCssOptions,
+  RuleOptions,
+  JssStyle
+} from 'jss/lib/types'
+import RuleList from 'jss/lib/RuleList'
```
HenriBeck: should be imported from `jss` as a `type`

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+import type {
+  InternalStyleSheetOptions,
+  Rule,
+  ToCssOptions,
+  RuleOptions,
+  JssStyle
+} from 'jss/lib/types'
```
HenriBeck: All of those should be imported from `jss`

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,48 @@
...
+import RuleList from 'jss/lib/RuleList'
+import {create} from '../jss'
+
+export interface JssSheet {
```
HenriBeck: No need to redefine the StyleSheet here. Just import it from `jss` as a type

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,48 @@
...
+  link(): JssSheet;
+  update(name?: string | Object, data?: Object): JssSheet;
+  toString(options?: ToCssOptions): string;
+}
+export type Options = {
+  theming?: {
+    themeListener: any => any
+  },
+  inject?: Array<'classes' | 'themes' | 'sheet'>,
+  jss?: create,
```
HenriBeck: Should be a Jss instance and not a function which creates an instance

packages/react-jss/src/createHoc.js
```diff
@@ -176,25 +199,15 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
-        : staticSheet.classes
-      const classes = {
-        ...defaultClasses,
-        ...jssClasses,
-        ...userClasses
-      }
+      const sheet = dynamicSheet || staticSheet
+      // $FlowFixMe defaultProps are not defined in React$Component
+      const defaultClasses = InnerComponent.defaultProps ? InnerComponent.defaultProps.classes : {}
+      const classes = {...defaultClasses, ...sheet.classes, ...userClasses}
```
HenriBeck: Why did you change the logic of this code?

HenriBeck:
Also please add typings for the other files when you have time. Especially for the JssProvider

package.json
gomesalexandre: Fair enough, I didn't see the `babel-plugin-transform-flow-strip-types"`, removed

packages/react-jss/package.json
gomesalexandre: Removed

packages/react-jss/src/createHoc.js
gomesalexandre: For now this is essentially a copy and paste from previous PR. Given the diff shows there are some changes (such as this one) that haven't been integrated in it, I'll take care of it top-priority, before working on anything else

packages/react-jss/src/createHoc.js
gomesalexandre: Same diff reason, fixed

packages/react-jss/src/types/index.js
gomesalexandre: Fixed

packages/react-jss/src/types/index.js
gomesalexandre: Fixed

packages/react-jss/src/createHoc.js
gomesalexandre: Reverted

packages/react-jss/src/createHoc.js
```diff
@@ -161,10 +182,12 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
...
-        staticSheet[dynamicStylesNs] = getDynamicStyles(styles)
-      }
-
-      const dynamicStyles = staticSheet[dynamicStylesNs]
+        dynamicStyles = compose(
+          staticSheet.classes,
+          getDynamicStyles(styles)
+        )
+        staticSheet[dynamicStylesNs] = dynamicStyles
+      } else dynamicStyles = staticSheet[dynamicStylesNs]
```
HenriBeck: Again, no logic changes here please

packages/react-jss/src/createHoc.js
```diff
@@ -62,20 +76,25 @@ let managersCounter = 0
  * @param {Object} [options]
  * @return {Component}
  */
-export default (stylesOrCreator, InnerComponent, options = {}) => {
+export default function createHOC<P>(
+  stylesOrCreator: StylesOrThemer,
+  InnerComponent: ComponentType<P>,
+  options: Options = {}
+): ComponentType<any> {
```
HenriBeck: The returning component props should be typed exactly

packages/react-jss/src/createHoc.js
gomesalexandre: Used `(comp: ComponentType<{}> | null) => void` in the current implementation, will implement generics in the next PR, it's still a bit of overhead for me still

packages/react-jss/src/compose.js
```diff
@@ -1,3 +1,5 @@
+// @flow
+type Classes = {[string]: string}
```
HenriBeck: Should be also imported from `jss` https://github.com/cssinjs/jss/blob/master/packages/jss/src/index.js#L33

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import type {StyleSheetOptions} from 'jss'
+import jss from '../jss'
```
HenriBeck: Import the `Jss` type also from `jss`

packages/react-jss/src/types/index.js
```diff
@@ -0,0 +1,19 @@
...
+  },
+  inject?: Array<'classes' | 'themes' | 'sheet'>,
+  jss?: jss,
+  ...$Shape<StyleSheetOptions>
+}
+export type Theme = {}
+export type Styles = {[string]: {}}
+export type ThemerFn = (theme: Theme) => Styles
+export type StylesOrThemer = Styles | ThemerFn
+export type Classes<S> = {|
```
HenriBeck: Should be also imported from `jss` https://github.com/cssinjs/jss/blob/master/packages/jss/src/index.js#L33

HenriBeck:
Quick guideline on which props the HOC should accept:
- All of the props of the injected Component
- An additional `innerRef` prop which needs to be a function with the injected Component as an argument

I don't think we need to use the `$Diff` operator for props like `classes` because the user is actually allowed to overwrite those props.

packages/react-jss/src/compose.js
gomesalexandre: Fixed

packages/react-jss/src/types/index.js
gomesalexandre: Fixed

packages/react-jss/src/types/index.js
gomesalexandre: Fixed

gomesalexandre:
@HenriBeck Do we really need to init options to empty object literal `{}` if not passed to `createHOC`? It produces a bit of flow errors because of l.85 i.e
`const {theming = defaultTheming, inject, jss: optionsJss, ...sheetOptions} = options`.

Then of course, accessing a property of undefined `options` object with no default would produce errors but that could be mitigated.

gomesalexandre:
@HenriBeck Regarding the HOC, the returned component is stripped of the `classes` prop isn't it ? That's the logic behing the use of the `$Diff` util here

HenriBeck:
No, I was wrong about that before, see https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/createHoc.js#L191

User classes override the from the stylesheet.

.eslintrc
```diff
@@ -4,7 +4,8 @@
   "rules": {
     "import/no-unresolved": 0,
     "flowtype/define-flow-type": 1,
-    "flowtype/use-flow-type": 1
+    "flowtype/use-flow-type": 1,
+	"semi": ["error", "never"]
```
romelperez: Should we use spaces instead of tabs?

HenriBeck: I think this can be reverted anyway because prettier takes care of it. I actually write the code with semicolons and prettier removes them for me

HenriBeck: But yes, definetly spaces

gomesalexandre: Using the indent rule would avoid such issue, I have automatic linting fixes on save in VSCode. Plus we would get errors thrown by the hook, should I add it ?

HenriBeck: We also run prettier in the precommit git hook so it's unnecessary

HenriBeck:
@kof types should be done

packages/react-jss/src/createHoc.js
```diff
@@ -57,41 +74,49 @@ let managersCounter = 0
 /**
  * Wrap a Component into a JSS Container Component.
  *
+ * IProps: Props of the InnerComponent.
+ * OProps: The Outer props the HOC accepts.
+ *
  * @param {Object|Function} stylesOrCreator
```
kof: I think we can remove those jsdoc comments now

packages/react-jss/src/createHoc.js
```diff
@@ -57,41 +74,49 @@ let managersCounter = 0
 /**
  * Wrap a Component into a JSS Container Component.
  *
+ * IProps: Props of the InnerComponent.
```
kof: InnerProps would be probably a better name, no?

packages/react-jss/src/JssProvider.js
```diff
@@ -1,11 +1,23 @@
-import {Component, Children} from 'react'
+// @flow
+import {Component, Children, type Element, type ElementType} from 'react'
 import PropTypes from 'prop-types'
+import type {Jss, generateClassName as GenerateClassNameType, SheetsRegistry} from 'jss'
```
kof: why mapping from lower case to capital?

kof: ohh, I think we need to rename it to upper case in the core package in the first place

kof: Once it starts uppercase you don't need to add Type suffix either

packages/react-jss/src/createHoc.js
HenriBeck: already using InnerProps though, that's why I had to name it IProps 🤔

packages/react-jss/src/createHoc.js
HenriBeck: Will do some cleanup in another PR. Have some small fixes etc. in mind

packages/react-jss/src/createHoc.js
kof: lets find a better convention, maybe InnerPropsType?

packages/react-jss/src/createHoc.js
```diff
@@ -176,6 +209,7 @@ export default (stylesOrCreator, InnerComponent, options = {}) => {
         })
       }
 
+      // $FlowFixMe InnerComponent can be class or stateless, the latter doesn't have a defaultProps property
```
kof: it may have defaultProps property same way as class based

packages/react-jss/src/types.js
```diff
@@ -0,0 +1,43 @@
...
+  sheet?: {}
+}
+// Needs to be hard coded for stricter types
+export type Context = {
+  '64a55d578f856d258dc345b094a2a2b3'?: Jss,
+  d4bd0baacbc52bbd48bbb9eb24344ecd?: SheetsRegistry,
+  b768b78919504fba9de2c03545c5cd3a?: {[key: number]: SheetsManager},
+  '6fc570d6bd61383819d0f9e7407c452d': StyleSheetFactoryOptions & {disableStylesGeneration?: boolean}
+}
+export type OuterProps<IP, C> = IP & {innerRef: (instance: ElementRef<C> | null) => void}
```
kof: Lets not use these abbreviations, they are hard to understand

packages/react-jss/src/types.js
```diff
@@ -0,0 +1,43 @@
...
+export type Styles = {[string]: {}}
+export type ThemerFn = (theme: Theme) => Styles
+export type StylesOrThemer = Styles | ThemerFn
```
kof: ThemerFn -> ThemeCreator

packages/react-jss/src/createHoc.js
HenriBeck: yeah but flow doesn't have them https://github.com/facebook/flow/blob/master/lib/react.js#L99

packages/react-jss/src/types.js
HenriBeck: ThemeCreator seems a little bit misleading

packages/react-jss/src/createHoc.js
kof: weird, we should look for an issue there for explanation or raise one .... really weird

PR #819, opened 2018-08-09
Fixes sheet ordering

HenriBeck:
Fixes the sheet ordering as described in #816 
- [x] Fix is implemented
- [x] Updated a test case which failed before

`prevNode` returns now an object because `nextSibling` may return null when it's the last element which results in not being able to get the parent for the insertion.

### DIFF
--- a/packages/jss/src/renderers/DomRenderer.js
+++ b/packages/jss/src/renderers/DomRenderer.js
@@ -273,8 +273,8 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
   const {insertionPoint} = options
   const nextNode = findPrevNode(options)
 
-  if (nextNode !== false) {
-    if (nextNode.parent) nextNode.parent.insertBefore(style, nextNode.node)
+  if (nextNode !== false && nextNode.parent) {
+    nextNode.parent.insertBefore(style, nextNode.node)
 
     return
   }
--- a/packages/jss/src/renderers/DomRenderer.js
+++ b/packages/jss/src/renderers/DomRenderer.js
@@ -216,33 +216,54 @@ function findCommentNode(text: string): Node | null {
   return null
 }
 
+type PrevNode = {
+  parent: ?Node,
+  node: ?Node
+}
+
 /**
  * Find a node before which we can insert the sheet.
  */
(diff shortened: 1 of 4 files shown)


packages/jss/src/renderers/DomRenderer.js
```diff
@@ -216,45 +216,66 @@ function findCommentNode(text: string): Node | null {
...
-function findPrevNode(options: PriorityOptions): ?Node | null {
+function findPrevNode(options: PriorityOptions): PrevNode | false {
   const {registry} = sheets
 
   if (registry.length > 0) {
     // Try to insert before the next higher sheet.
     let sheet = findHigherSheet(registry, options)
-    if (sheet) return sheet.renderer.element
+    if (sheet) {
+      return {
```
kof: Seems like calling parentNode in advance doesn't bring much value other than makes findPrevNode semantics wrong. Mb I oversee  something but to me it looks like we could keep the return value as it was.

packages/jss/src/renderers/DomRenderer.js
```diff
@@ -268,7 +289,7 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
     return
   }
 
-  getHead().insertBefore(style, prevNode)
```
kof: it worked basically before same way, because the insertBefore api allows to pass prevNode === null and it will be inserted in the end of the parent, so the same as appendChild

packages/jss/src/renderers/DomRenderer.js
HenriBeck: The problem is when returning nextSibling and when it returns `null` we can't get the parentNode

packages/jss/src/renderers/DomRenderer.js
HenriBeck: seems a bit simpler to me

packages/jss/src/renderers/DomRenderer.js
HenriBeck: Not necessarily, it was supposed to work with appending it at the end of the parent of the stylesheet

kof:
We should have supported an insertion point as node as an only argument, I think this would simplify it.

PR #824, opened 2018-08-16
Added running linting, flow and prettier on the CI 

HenriBeck:
This makes sure the user didn't skip and we don't merge anything without running prettier etc.


package.json
```diff
@@ -5,7 +5,9 @@
     "typecheck": "flow check --max-warnings=0",
     "lint": "eslint scripts/ packages/ && npm run typecheck && npm run format",
     "lint:staged": "lint-staged && npm run typecheck",
-    "format": "prettier --write *.{js,md,json} {docs,packages,scripts}/**/*.{js,md,json}",
+    "lint:ci": "eslint scripts/ packages/ && npm run typecheck && npm run format:ci",
+    "format": "prettier *.{js,md,json} {docs,packages,scripts}/**/*.{js,md,json} --write",
+    "format:ci": "prettier *.{js,md,json} {docs,packages,scripts}/**/*.{js,md,json} --list-different",
```
kof: if you can use --write and --list-different at the same time, then you might not need separate format:ci command, but instead just run npm run format -- --list-different

kof: In that case you also won't need separate lint:ci too

HenriBeck: Yeah, I was able to remove `format:ci` but not `lint:ci` because we can't pass the flag to the `format` npm script from the travis file

PR #825, opened 2018-08-16
Move compose classes func

HenriBeck:
I moved the `compose` function which was previously in `react-jss` into `jss` and added support for composing multiple classes objects. This will allow others to merge multiple class objects more easily.

As of not increasing the `jss` core size, we now support ES modules which allows for tree shaking and because it's not a function which is being used in `jss`should be removed most of the time automatically.


kof:
Hmm, trying to think of a use case, where it might be needed besides of react-jss. Do you have any in mind?

Another thing: bundlephobia will report still the full size and many people use it. Also many people use older bundler  versions without tree shaking. 

I was actually thinking to go the opposite direction and move everything out of the core that is not needed by majority and create separate small packages, since we can now easily add packages.

It could be even a jss-utilities package or something if we have more stuff like that

HenriBeck:
Creating a separate package like `jss-utils` for example would work too.

My use case would be to provide a fast option for anyone who might need that.
I think it falls into the same category as `getDynamicStyles`.

I was planning to add support for configuring how the classes inside `react-jss` would be merged, that would be a use case. For example `material-ui` also merges the user classes like that.

kof:
So lets create a utilities package. Not sure mui is not doing extra logic,
but if we can cover it easily and replace there, I guess it would be good.

HenriBeck:
Closed this as the `compose` function won't be moved into the core

PR #826, opened 2018-08-18
Fix CI

HenriBeck:
Currently, the CI isn't working. It is because prettier finds more files on the CI to format as locally.
Those files aren't being formatted locally and cause prettier to fail on the CI. Those files only are mostly json files.


HenriBeck:
The problem was that locally it used the glob expanding of the system because we didn't put the globs into `""`. This meant that it could differ which files prettier formats between different OS.

This was also causing the mismatch between the CI and my local personal machine.

Issue #576, opened 2017-08-29
jss.setup should extend instead of override


kof:
People are wondering why they call .setup and options they passed over create() are gone.

kof:
released with v9

PR #827, opened 2018-08-20
Update js-api.md

eps1lon:
- Fix note about setup overwrite

Looking at the source https://github.com/cssinjs/jss/blob/dd20297caba69ac4c390f1460dca0034691f4a2a/packages/jss/src/Jss.js#L47 options are merged on setup not completely overwritten.

I'm also asking because `react-jss` expects the options to be public readonly and the typings over at DefinitelyTyped do not have options listed either. 

All of which is no direct concern of you but I'd like to get some feedback before I work on PRs in those packages. Might also be a good idea to document the options property in the api docs.

### DIFF
--- a/docs/js-api.md
+++ b/docs/js-api.md
@@ -37,7 +37,7 @@ Options:
 - `virtual` - if true, JSS will use VirtualRenderer.
 - `insertionPoint` - string value of a DOM comment node which marks the start of sheets or a rendered DOM node. Sheets rendered by this Jss instance are inserted after this point sequentially.
 
-**Note**: Each `jss.setup()` call will perform a shallow merge with the old options except for `plugins`. Passed `plugins` will be added to the existing plugins if they do not already exist.
+**Note**: Each `jss.setup()` call will perform a shallow merge with the old options except for `plugins`. Passed `plugins` will be added to the existing plugins.
 
 
 See [setup examples](./setup.md#specify-dom-insertion-point).
--- a/docs/js-api.md
+++ b/docs/js-api.md
@@ -37,6 +37,9 @@ Options:
 - `virtual` - if true, JSS will use VirtualRenderer.
 - `insertionPoint` - string value of a DOM comment node which marks the start of sheets or a rendered DOM node. Sheets rendered by this Jss instance are inserted after this point sequentially.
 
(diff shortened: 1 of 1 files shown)


kof:
LGTM

kof:
Actually, most options do overwrite previously defined options, plugins option though doesn't uninstall previous plugins, it adds the new plugins additionally to previously defined once. Mb worth writing it that way down?

eps1lon:
I don't know. It just sounded like setup had the following behavior
```js
jss.setup({ foo: 'foo' });
// jss.options.foo === 'foo'
jss.setup({ bar: 'bar' })
// jss.option.foo === undefined or whatever was default
```

HenriBeck:
Let's maybe not remove the complete line but only the `all` word? 
This seems to be the reason for the mistaking.

eps1lon:
Behavior was changed in #576. Probably just forgot to update docs?

kof:
Yes we fixed it at some point, so it actually extends

eps1lon:
I've rewritten the wording. Maybe this is clearer. At least the behavior concerning plugins was not described correct before.

docs/js-api.md
```diff
@@ -37,7 +37,8 @@ Options:
 - `virtual` - if true, JSS will use VirtualRenderer.
 - `insertionPoint` - string value of a DOM comment node which marks the start of sheets or a rendered DOM node. Sheets rendered by this Jss instance are inserted after this point sequentially.
 
-**Note**: Each `jss.setup()` call will overwrite all previous settings.
+**Note**: Each `jss.setup()` call will perform a shallow merge with the old options except for `plugins`. Passed `plugins` will be added to the existing plugins if they do not already exist.
```
kof: We have no duplicate detection right now

eps1lon: It does call `#use` which does
https://github.com/cssinjs/jss/blob/dd20297caba69ac4c390f1460dca0034691f4a2a/packages/jss/src/Jss.js#L131

This seems like duplicate detection.

kof:
True, doesn't work though always, because usually you have a function that returns an object with hooks. In that case ref will be a new one, so this thing would only help if user saves the reference. It is so unlikely that this check is almost useless.

kof:
merged, thank you!

PR #831, opened 2018-08-21
Provide "global" shim for each bundle

TrySound:
I'm currently bundle my production app with rollup and found unexpected
`global` stuff in consumed modules. I think this should be handled in
library side and provide browser friendly bundle. So the only
requirement could be shimming/replacing `NODE_ENV`.


TrySound:
Here's my current solution

image

packages/jss/.size-snapshot.json
```diff
@@ -10,21 +10,21 @@
     "gzipped": 5827
   },
   "dist/jss.cjs.js": {
-    "bundled": 48623,
-    "minified": 20945,
-    "gzipped": 5971
+    "bundled": 54298,
+    "minified": 24068,
+    "gzipped": 6908
```
HenriBeck: Why is there such a significant increase in size?

TrySound: Ooops, and here's the problem. node-globals unconditionally polyfills everything. Even `process` object which is used by us for NODE_ENV.

TrySound: Added PR
https://github.com/calvinmetcalf/rollup-plugin-node-globals/pull/16

HenriBeck: I mean we only need to shim `process.env.NODE_ENV`, correct? 
And this should always be replaced by the user.

TrySound: We only need to shim `global` but `process.env.NODE_ENV` should be replaced on user side.

kof: shit I didn't see the size increase

TrySound: Seems like umd bundles size is also increased since `nodeGlobals` is used before `replace`.

TrySound: The simplest work around is own ponyfill which we could reuse in modules.

HenriBeck: are we using anything from `global` other than `process.env.NODE_ENV` or `process.env.VERSION`?

TrySound: `global` is used to create unique identifiers and nothing more. `process.env.VERSION` is replaced statically on our side. `process.env.NODE_ENV` should be replaced statically on user side.

TrySound:
Seems like preventing polyfilling `process` object significantly reduced jss+plugins size.

kof:
awesome!

TrySound:
At least umd bundles are smaller

Issue #802, opened 2018-08-03
[react-jss] prepare for react 17 (but maintain backward compat)


HenriBeck:
_From @pradel on April 11, 2018 15:23_

From the react docs: `React version 17 will deprecate several of the class component API lifecycles: componentWillMount, componentWillReceiveProps, and componentWillUpdate.`

Would be great to have this library compatible :)

Can be migrated using https://github.com/reactjs/react-lifecycles-compat

Thanks!

_Copied from original issue: cssinjs/react-jss#231_

HenriBeck:
_From @damianobarbati on June 2, 2018 9:50_

```
    <React.StrictMode>
        <ThemeProvider theme={theme}>
            <Router />
        </ThemeProvider>
    </React.StrictMode>
```

=> 

```
Warning: Unsafe lifecycle methods were found within a strict-mode tree:
    in App

componentWillMount: Please update the following components to use componentDidMount instead: Jss(Home), Jss(Layout), Jss(Nav), SideEffect(NullComponent), ThemeProvider

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: Jss(Home), Jss(Layout), Jss(Nav), ThemeProvider

componentWillUpdate: Please update the following components to use componentDidUpdate instead: Jss(Home), Jss(Layout), Jss(Nav)

Learn more about this warning here:
https://fb.me/react-strict-mode-warnings
```

HenriBeck:
_From @RusinovAnton on July 5, 2018 11:4_

Hey, I would like to help with this one. I need some heads up, is it correct that using following element for tests should cover all possible components with lifecycles?

```
<React.StrictMode>
  <ThemeProvider theme={theme}>
    <div />
  </ThemeProvider>
</React.StrictMode>
```

HenriBeck:
_From @kof on July 5, 2018 11:8_

also JssProvider and the biggest one injectSheet() HOC creator

damianobarbati:
Hey @HenriBeck do you need with this?

HenriBeck:
I assume you are asking if I/we need help with this and yes, feel free to create a PR.

felixputera:
Hi @HenriBeck,

I was looking into this matter and noticed that `react-jss` is also using the legacy context API which is also called in `componentWillReceiveProps`:
```js
    componentWillReceiveProps(nextProps: OuterPropsType, nextContext: Context) {
      this.context = nextContext
      const {dynamicSheet} = this.state
[...]

PR #834, opened 2018-08-29
react-jss: Remove usage of deprecated lifecycle methods (#802)

felixputera:
Addressing #802 

I had removed (almost) all of the usage of old lifecycle methods. I had not removed the `componentWillReceiveProps` because it is currently used for `context`, hence I also skipped the new `StrictMode` test.

For eslint, I had turned off `react/no-did-mount-set-state` and `react/no-did-update-set-state` (as written in [React's blog post](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#fetching-external-data-when-props-change)):

> Sometimes people use componentWillUpdate out of a misplaced fear that by the time componentDidUpdate fires, it is “too late” to update the state of other components. This is not the case. React ensures that any setState calls that happen during componentDidMount and componentDidUpdate are flushed before the user sees the updated UI. 

However, I am not sure if that's the case for versions prior `16.3`.

### DIFF
--- a/packages/react-jss/src/createHoc.js
+++ b/packages/react-jss/src/createHoc.js
@@ -110,9 +110,6 @@ export default function createHOC<
       const theme = isThemingEnabled ? themeListener.initial(context) : noTheme
 
       this.state = this.createState({theme, classes: {}}, props)
-    }
-
-    componentWillMount() {
       this.manage(this.state)
     }
 
@@ -121,23 +118,22 @@ export default function createHOC<
         this.unsubscribeId = themeListener.subscribe(this.context, this.setTheme)
       }
     }
-
+    // TODO: remove this lifecycle when migrating to new Context API
     componentWillReceiveProps(nextProps: OuterPropsType, nextContext: Context) {
       this.context = nextContext
-      const {dynamicSheet} = this.state
-      if (dynamicSheet) dynamicSheet.update(nextProps)
-    }
-
-    componentWillUpdate(nextProps: OuterPropsType, nextState: State) {
-      if (isThemingEnabled && this.state.theme !== nextState.theme) {
-        const newState = this.createState(nextState, nextProps)
-        this.manage(newState)
-        this.manager.unmanage(this.state.theme)
-        this.setState(newState)
-      }
     }
 
     componentDidUpdate(prevProps: OuterPropsType, prevState: State) {
+      const {dynamicSheet} = this.state
+      if (dynamicSheet) dynamicSheet.update(this.props)
+
+      if (isThemingEnabled && this.state.theme !== prevState.theme) {
+        const newState = this.createState(this.state, this.props)
+        this.manage(newState)
+        this.manager.unmanage(prevState.theme)
+        this.setState(newState)
+      }
+
       // We remove previous dynamicSheet only after new one was created to avoid FOUC.
       if (prevState.dynamicSheet !== this.state.dynamicSheet && prevState.dynamicSheet) {
         this.jss.removeStyleSheet(prevState.dynamicSheet)
--- a/packages/react-jss/src/createHoc.js
+++ b/packages/react-jss/src/createHoc.js
@@ -118,10 +118,6 @@ export default function createHOC<
         this.unsubscribeId = themeListener.subscribe(this.context, this.setTheme)
       }
     }
-    // TODO: remove this lifecycle when migrating to new Context API
-    componentWillReceiveProps(nextProps: OuterPropsType, nextContext: Context) {
-      this.context = nextContext
-    }
 
     componentDidUpdate(prevProps: OuterPropsType, prevState: State) {
       const {dynamicSheet} = this.state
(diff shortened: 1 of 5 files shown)


packages/react-jss/src/createHoc.js
```diff
@@ -121,23 +118,22 @@ export default function createHOC<
         this.unsubscribeId = themeListener.subscribe(this.context, this.setTheme)
       }
     }
-
+    // TODO: remove this lifecycle when migrating to new Context API
     componentWillReceiveProps(nextProps: OuterPropsType, nextContext: Context) {
       this.context = nextContext
```
HenriBeck: I'm not sure what this line does, maybe try removing it and run the tests to see if anything breaks.

packages/react-jss/src/injectSheet.test.js
```diff
@@ -37,6 +37,21 @@ describe('injectSheet', () => {
       })()
     })
 
+    it.skip('should work in StrictMode without error on React 16.3+', () => {
```
HenriBeck: I really like the test, but why are you skipping it?

packages/react-jss/src/createHoc.js
```diff
@@ -121,23 +118,22 @@ export default function createHOC<
...
     }
 
-    componentWillUpdate(nextProps: OuterPropsType, nextState: State) {
-      if (isThemingEnabled && this.state.theme !== nextState.theme) {
-        const newState = this.createState(nextState, nextProps)
+    componentDidUpdate(prevProps: OuterPropsType, prevState: State) {
+      const {dynamicSheet} = this.state
+      if (dynamicSheet) dynamicSheet.update(this.props)
+
+      if (isThemingEnabled && this.state.theme !== prevState.theme) {
```
HenriBeck: I think this can be moved into a function called `handleThemeUpdate` which is the callback to the themeListener.subscribe

packages/react-jss/src/createHoc.js
felixputera: Hmm yeah its weird, I just commented it out and nothing breaks.

packages/react-jss/src/injectSheet.test.js
felixputera: Was commenting it out for the one leftover componentWillReceiveProps, but removing it because it seems that I can just remove the componentWillReceiveProps.

packages/react-jss/src/createHoc.js
felixputera: Correct me if im wrong, but themeListener.subscribe only accept one callback. Do you mean to write another themeListener.subscribe? 

Another question: in the callback, how can I access to the previous theme (`prevState.theme`) to unmanage it?

HenriBeck: yes, themeListener.subscribe accepts one argument. I think the current one is called `setTheme`.  Make it so that the whole creating of the state and sheets when the theme changes are in there. The new theme is being passed as an argument.

The only thing which need to stay in componentDidUpdate is the `unmanage` of the old theme.

felixputera: Hi, so I was trying out your suggestion. Apparently, i can't call `createState` in `setTheme` because there's  some side effects from calling `createState` (I am not sure what precisely the side effect(s) is/are). Moving them causes some test fail.

TrySound:
I would migrate to the new context api first

felixputera:
@TrySound should I open a new issue? I'm very unfamiliar with the process here.

TrySound:
No process actually. If you want just open new pr and rebase this one after merge.

felixputera:
Sure, I'll take a look and submit a PR when its ready

packages/react-jss/src/createHoc.js
```diff
@@ -122,22 +119,17 @@ export default function createHOC<
       }
     }
 
-    componentWillReceiveProps(nextProps: OuterPropsType, nextContext: Context) {
```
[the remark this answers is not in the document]
HenriBeck: You are talking about the updating of the dynamic sheet, right?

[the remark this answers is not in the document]
HenriBeck: We could move the logic into `static getDerivedStateFromProps` and merely return null but update the dynamic sheet with the new props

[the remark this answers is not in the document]
kof: Wait, I just recalled a discussion I had about this thing, componentDidUpdate and componentDidMount are called synchronously after commiting  to the DOM, so sync CSSOM updates are still done before the the recalc is scheduled. We are good.

kof:
thanks @felixputera @HenriBeck and @TrySound

HenriBeck:
@felixputera you are working on the new context as well, right? We should use [this](https://github.com/jamiebuilds/create-react-context) package for backward compatibility.

TrySound:
@HenriBeck Do we really need backward compatibility with major release?

HenriBeck:
Just created an issue for that. Let's discuss there.

Issue #836, opened 2018-09-01
[react-jss] Drop support for react 0.13 and 0.14


HenriBeck:
We currently still support React 13 and 14 as defined in the peer dependencies in the package.json. I think we can drop those versions without any significant problems.

I think we should hold off on dropping React 15 because most other react libaries still support it.

kof:
What is the benefit of dropping it? We migrated away from componentWillReceiveProps already.

HenriBeck:
We need to drop those versions (at least 13) for using the create context polyfill.

HenriBeck:
Implemented in #851

PR #851, opened 2018-09-11
[react-jss] Drop React 0.13

HenriBeck:
__What would you like to add/fix?__ Drop react 0.13 to eventually switch to the new context API polyfill.

__Corresponding issue (if exists):__ #836


kof:
merged

Issue #710, opened 2018-05-08
[jss-nested] before/after pseudo elements do not seem to support functions as rule values


nicholasrice:
I'm running into an issue where assigning a function value to a property inside a pseudo element causes the following error:

`Warning: Rule is not linked. Missing sheet option "link: true".`

Strangely, it does not seem to repro with pseudo classes.

I've isolated a repo here: https://codesandbox.io/s/kor8kkzjzv

kof:
Looks like a bug related to jss-nested plugin.

kof:
Any chance you could look into it?

kof:
Interesting, that might be even a bug in the core, potentially CSSOM related, because of rule linking.

nicholasrice:
@kof Yea I'm digging into it. I also noticed that non-standard media queries also throw the same warning when assigned a function value:

```
"@media screen and (-ms-high-contrast: active)": {
    foo: {
        background: config => config.background
    }
}
``` 

My hunch is the two are related, unless its known that browser-specific media queries wont work. But the fact that the above repros without jss-nested makes me think both are related to core.

Good call @HenriBeck - I didn't try the *correct* syntax (bad habit on on my part)

nicholasrice:
Okay I tracked the issue down:
- `StyleSheet.link()` retrieves the  `CSSRuleList` from the renderer (and subsequently from the DOM) and passes it to `RuleList.link`.
- `RuleList.link()` iterates over the CSSRuleList and tries to match them to `StyleRules` stored in the `RuleList`. If a match is found, the rule is linked, otherwise it is ignored.

The last part is where the issue is originating from. It seems that the `CSSStyleRule` and `CSSMediaRule` `selectorText` properties are getting mutated by the browser:
- `.foo-0-0-1:after` becomes `.foo-0-0-1::after`
- `@media screen and (-ms-high-contrast: active)` becomes `@media not all`

So `RuleList.map` is storing rules by one string (`.foo-0-0-1:after`) and the selectorText of the CSSStyleRule that is created *from* that `StyleRule` is mutated (`.foo-0-0-1::after`). This means JSS is not able to match the JSS rule with the CSS rule, which results in the error that the rule has not been linked whenever the stylesheet is updated.

I'm trying to think of a more bullet-proof way to associate JSS rules with CSS rules but I don't have any stellar ideas yet - let me know if you have any thoughts on it.

kof:
[...]

PR #857, opened 2018-09-16
Always use insertRule API with option link: true

### LINKED ISSUE #815: "Rule is not linked" error for function rules with non-canonical selector formatting
Minimal reproduction: https://codesandbox.io/s/016vzn15ol

If I define two dynamic rules like this:

```javascript
  nospace: {
    "@media (min-width:1px)": {
      color: ({ color }) => color
    }
  },
  space: {
    "@media (min-width: 1px)": {
      color: ({ color }) => color
    }
  },
```

The first one does not work, but the second one does. The difference is the missing space between `min-width:` and `1px`.

I’ve tracked this down to [RuleList.link()](https://github.com/cssinjs/jss/blob/3ccac24bc31837d2727f4b51838f2b629e46871d/packages/jss/src/RuleList.js#L157) where the key given by `DomRenderer.getKey()` is sanitised by the browser, whereas the key originally used in `this.map` was not. 

I’m not sure the best way to fix this. Now that I know the problem, the workaround is obvious, but it seems like something we should fix because it was quite difficult to figure out why my rules weren’t working.

This is the same problem first reported by @jpetitcolas [here](https://github.com/cssinjs/jss/issues/664#issuecomment-370729498), in whose case the error was caused by using `:after` instead of the more correct `::after`.

kof:

Current linking logic causes too many bugs. This approach simplifies it. We are not linking any more using selector, we create rules using insertRule API always when we need to link a rule.

- [x] Need to test which currently open issues it fixes.
- [x] Need to check how linking works with nesting and container rules like media queries.

Fixes #815, #710, #664


kof:
This PR seems to be ready

kof:
Ok added a test with media query, its not ready, fixing it.

packages/jss/src/StyleSheet.js
```diff
@@ -34,7 +31,6 @@ export default class StyleSheet {
   constructor(styles: Object, options: StyleSheetOptions) {
     this.attached = false
     this.deployed = false
-    this.linked = false
```
HenriBeck: Why did you remove the linked property?

kof: because we don't need it any more, if options.link is true, then it is always linked, there is no separate sheet.link() call.

HenriBeck: Yeah makes sense

kof:
Should fix (still verifying) #815, #710, #664
@TrySound @HenriBeck please review the last commit

TrySound:
Looks good to me

kof:
Merged

PR #858, opened 2018-09-16
Migrate to yarn

TrySound:
One of NPM maintainers shows some disrespect to others which kills a
trust to them.

Yarn is more stable client with more full set of features so I suggest
to migrate to it.

As you can see yarn.lock is two times smaller because of less verbose
syntax. Lockfile is not required for every package in monorepo, thanks to
yarn workspaces.


TrySound:
Original package-lock.json was gracefully converted to `yarn.lock` by its `yarn import` command.

kof:
especially in the sight of getting rid of node_modules, I am now all in.

kof:
Though if its really about npm inc, we should stop using registry as well haha

kof:
What is the reason to actually use npm registry?

TrySound:
Supporting own registry is quite expensive since all packages should be synchronised or lazy loaded from the same npm. So yarn registry for is is just an alias.

kof:
No, I mean why not just use directly github?

TrySound:
Because in this case you should push build artifacts which is verbose for history. There is also a problem with loading all this stuff. Repos may be quite big. And monorepo's won't work with github links.

kof:
true

kof:
I will talk to https://unpkg.com/ guys to see what we can do there.

kof:
We just need to be able to publish to a cdn like unpkg without npm registry and thats it.

kof:
can you please also update docs where npm was used? just remove npm completely

scripts/build.js
```diff
@@ -25,7 +24,7 @@ function build(options = {}) {
 
   if (!fs.existsSync('src')) {
     if (!silent) {
-      log.error('No src dir')
+      console.error('No src dir')
```
kof: linter warnings, need to add eslint-disable statement

kof:
Merged!!!! 🎉🎉🎉🎉

HenriBeck:
I think yarn is better client though not publishing to npm and only to unpkg would be a mistake in my opinion. I think we would lose a lot of users through that.

kof:
We can publish, but we could avoid using it ourselves.

kof:
Btw. does yarn has its own registry or is it just using npm registry?

HenriBeck:
The yarn registry being a mere mirror of the npm registry. Though this is being removed in yarn v2 I think.

TrySound:
I think the best is having central registry with packages and if as a fallback distributed database like torrent.

TrySound:
Yarn registry is already removed I think. Currently it's just a proxy.

Issue #856, opened 2018-09-14
[react-jss] theme prop without styled themes


HenriBeck:
__Is your feature request related to a problem? Please describe.__ Get the theme as a prop even if the styles aren't themed.

__Describe the solution you'd like__
We either allow for an option like `withTheme` or always subscribe to theme updates and pass the theme, though this solution would require some investigation if it creates performance issues. 

Initially created in cssinjs/react-jss#291

HenriBeck:
@kof and I decided that we will implement this feature by simply only passing the theme down to the component when it's inside the `inject´ sheet. We will subscribe to the theme when the styles are themed, or the theme is explicitly passed down.

kof:
@HenriBeck means when "inject" option contains "theme": e.g. `injectSheet(styles, {inject: ['classes', 'theme']})`.

Also we discussed that we should turn off theme passing to the child component by default. I couldn't recall why I decided to pass the theme by default to the child anyways. If we find that out we could rethink. Right now I don't know why child component needs to be aware of the theme over props without explicitly asking for that.

PR #864, opened 2018-09-24
Added withTheme option

HenriBeck:
__What would you like to add/fix?__ Add an option for manually subscribing to the theme and passing it down to the component, even if the styles aren't themed.

__Fixed:__ Currently, there are still multiple stylesheets for different themes generated even though the styles aren't different. 

__Corresponding issue:__ #856


HenriBeck:
Will implement this feature when we switch to the new `theming` version. This will be much simpler

PR #868, opened 2018-09-27
[react-jss] Drop React 14

HenriBeck:
__What would you like to add/fix?__ Drop React 14 support

### DIFF
--- a/packages/react-jss/package.json
+++ b/packages/react-jss/package.json
@@ -36,7 +36,7 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "peerDependencies": {
-    "react": ">=0.14"
+    "react": ">=0.15"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


kof:
To give others some context: we are switching theming package to the new context api, which requires react 15+

packages/react-jss/package.json
```diff
@@ -36,7 +36,7 @@
     "prepare": "node ../../scripts/prepare.js"
   },
   "peerDependencies": {
-    "react": ">=0.14"
+    "react": ">=0.15"
```
jdreesen: Is this correct? I think it should be `>=15` as React switched to major versions with React v15 (see: https://reactjs.org/blog/2016/04/07/react-v15.html)

kof: good catch, yes its true!

PR #872, opened 2018-09-28
Update json-api.md

shaneporter:
Code sample

prefixed `random()` call with `Math` so that it works.


HenriBeck:
Thanks. Merged 👍

PR #873, opened 2018-10-01
Use pascal case for global namespace

kof:


TrySound:
This will make jss be accessed with `Jss`, `ReactJSS`, `JssPluginCache` etc. Is it okay?

kof:
Damn, checked old exports from jss

```
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports) :
  typeof define === 'function' && define.amd ? define(['exports'], factory) :
  (factory((global.jss = {})));
}(this, (function (exports) { 'use strict';

```

kof:
Also react-jss was using "reactJss" 

```
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory(require("React"));
	else if(typeof define === 'function' && define.amd)
		define(["React"], factory);
	else if(typeof exports === 'object')
		exports["reactJss"] = factory(require("React"));
	else
		root["reactJss"] = factory(root["React"]);
```

TrySound:
I think it's okay for breaking change. Let's go with it?

kof:
Its just confusing that Jss is a namespace but its not a constructor, I think React isn't either.

TrySound:
Then we need to provide an explicit globals object. And convert `name` with normal camel case.

kof:
I wonde when will be the next time we have same problem with some peer, we can't actually decide globally what global namespace any of the peer dependencies may use.

TrySound:
We can actually. Just take a look at peer umd
https://unpkg.com/react@16.5.2/umd/react.development.js

kof:
updated to use for jss packages camel case, but for peers - pascal case

TrySound:
jss is also peer dependency, isn't it?

HenriBeck:
On react-jss not. On most of the plugins yes.

TrySound:
Should be on react-jss too.

kof:
Shit :)

kof:
Any ideas how to do this without maintaining a whitelist?

TrySound:
No way. Lib authors may not follow only one pattern.

TrySound:
It's not so big list. react, react-dom and jss. Am I miss something?

kof:
Ok, whitelist with a warning it is now

PR #877, opened 2018-10-08
Update users.md

sammi:

### DIFF
--- a/docs/users.md
+++ b/docs/users.md
@@ -31,3 +31,4 @@
 [Trioxis](https://www.trioxis.com/)
 
 [Real World React](https://realworldreact.com) ([source](https://github.com/realworldreact/realworldreact-website))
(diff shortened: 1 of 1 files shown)


docs/users.md
```diff
@@ -31,3 +31,4 @@
 [Trioxis](https://www.trioxis.com/)
 
 [Real World React](https://realworldreact.com) ([source](https://github.com/realworldreact/realworldreact-website))
+[Bootstrp JSS](https://sammi.github.io/bootstrap-jss)([source](https://github.com/sammi/bootstrap-jss))
```
HenriBeck: Typo, should be `Bootstrap`

kof: yes, there is a typo. I am thinking that we can place this lib and mui together on the first readme section, so that it is easier to discover, but I think this lib needs better storybook integration, currently there is only a button. As soon as all components are in storybook.

sammi: Thanks guys, working on more stories.

kof:
merged, thanks, let me know when your showcase is ready with all components!

Issue #682, opened 2018-02-12
Support full syntax in function rules and values


kof:
Currently the only plugin that we run over function values/rules is jss-default-unit. The problem with running every plugin over it is that we would add performance overhead to something that is supposed to be usable for animations.

Originates in https://github.com/cssinjs/react-jss/issues/185

Cases:

- [x] fallbacks in fn rules
- [x] media queries inside of function rules
- [x] media query as a fn rule nested 
- [x] function values inside of media rules
- [x] jss-nested inside of fn rules
- [x] jss-nested as a fn rule
- [x] jss-global should support function values and rules
- [x] jss-compose should support linking function rules
- [x] jss-compose should support linking rules with fn values
- [x] jss.use(pluginObservable({process: false})), test for disabling plugins

kof:
I got an idea what we can do here: due to the fact that other cssinjs libs have also functions and they don't expect those to be updatable at 60 FPS we can do this:

1. We assume a mainstream assumption is that those styles don't update at 60FPS
1. We allow all required plugins to react on the values/rules changes like they do for normal values
1. Introduce a modified interface for "fast" functions: 

```js
{
  button: {
   // If value is an array and first item is a function, we got a function value.
    top: [props => props.top]
  }
}
```

```js
{
   // If rule is an array and first item is a function, we got a fast function rule.
  button: [(props) => ({
    top: props.top
  })
}
```

Afditionally we could allow passing options:

```js
{
  button: {
   // If value is an array and first item is a function, we got a function value.
    top: [props => props.top, {someOptions}]
  }
}
```
    

I am open for any other interface suggestions.

kof:
Fast functions being ones which can be updated frequently using `.update()` method without being slow for animations or other types of frequent updates.

Basically current function values/rules implementation IS the "fast" version, which is why I couldn't add all plugins to the results they return, because that would slow them down.

So we need  a way to use the entire JSS syntax for more "static" use cases, where static doesn't mean it can be done only once, but it shouldn't be updated frequently. This is the way styled-components is using them. At the same time I want to keep the "fast" version for frequent updates.

So we need to distinguish those 2 versions somehow.

kof:
@Mati365 we always post updates, subscribe!

AleshaOleg:
[...]

Issue #746, opened 2018-07-05
[jss-nested][jss-observable] Observables not working for '& div'


brianmhunt:
With JSS of:

```css
{
  alpha: {
    '& div': observable
  }
}
```

The first application will correctly use the value of the `observable`; however, when `observable` changes e.g. `{ height: '123' }`, the result never updates.

The observable is being monitored and the styleRule updated per [`onProcessRule`](https://github.com/cssinjs/jss/blob/master/packages/jss/src/plugins/observables.js#L38) as expected, however the `styleRule.prop` does not overwrite the prior value of the `& div` in the `<style>`.

Noting related issues #735 #737

HenriBeck:
Could you create a codesandbox with an example?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
Similar problem as with jss-expand, core is supposed to take out observable objects and replace them with values once they are received from subscription, but I was testing only with more simple examples without nested structures.

From the API perspective it is logical to support this, but on the other hand, the use cases are so rare and it is always possible to use it on the first level.

brianmhunt:
@henribeck do you have a precedent I can build on?

HenriBeck:
No sorry. The sandbox should only contain the code for reproducing the issue.

kof:
I think we fixed it already in #682, but I added this case as a task, and will add a test for it asap, follow #682 for updates

kof:
@brianmhunt actually would it be that bad if observables were supported only in top level? The way I was thinking about observables in JSS context is that you would want to use them when you need to update an animation very frequently, otherwise you can just use static objects or fn values/rules if you want to update them from time to time.

Observables was to me a specific case for animations.

kof:
The problem with supporting them everywhere right now is that we need to make every plugin that transforms styles aware of observables.

kof:
One idea: we could have a utility fn: `isDynamicValue()` which returns true for observables and functions and keeps them as they are during processing in all plugins.

kof:
I am thinking right now if we should handle Observables  consistently same way as fn values and set process: true by default. This way it is consistent and we don't make assumptions on how people use it. 

The reason is that in knockout or any other streams based framework, observables are not just a corner case but a default way of updating things, even for non-frequent changes.

cc @HenriBeck

kof:
Update: we decided to support plugins processing in observables by default

brianmhunt:
[...]

PR #878, opened 2018-10-09
Dynamic values full syntax

kof:
#682


packages/jss-plugin-syntax-rule-value-function/src/index.js
```diff
@@ -13,51 +22,59 @@ export default function functionPlugin() {
...
+
+      // If we have a style function, the entire rule is dynamic and style object
+      // will be returned from that function.
+      if (fnRule) {
+        styleRule.style = fnRule(data)
+
+        // We need to run the plugins in case style relies on syntax plugins.
+        const process = options ? options.process : true
+        if (process) {
+          styleRule.options.jss.plugins.onProcessStyle(styleRule.style, styleRule, sheet)
```
kof: don't really like this here, that api was supposed to be private, but  I haven't found a way around so far.

packages/jss/src/RuleList.js
```diff
@@ -127,16 +138,36 @@ export default class RuleList {
   /**
    * Update the function values with a new data.
    */
-  update: Update = (name, data) => {
+  update(...args: UpdateArguments): void {
```
kof: previous implementation was generating arrow method, which transpiles to a less performant inlined function in constructor

packages/jss-plugin-syntax-rule-value-function/src/index.js
HenriBeck: Can't this be moved into the core and then diff it against the previous style object and only update the properties which are needed?

HenriBeck: Also I thought we would go with the route of `fastValue((props) => props.color)` to make fast function values?

kof: strict comparison of previous and current style reference before the hook and after might actually be a good idea! Gonna try this:

`fastValue` we can still add as a separate feature later on, I think we can skip this for now, just because we still have observables for "fast by default" and its not clear if anyone actually using function values as "fast by default" besides of my demos.

kof: also  `sheet.update(data, {process: true})` is an equevalent, so we might not need `fastValue()`  at all

kof: > We could also make a shallow comparison right?

Yeah, can be done if needed, I think we only need to check style references currently.

> Also what do you think of only supporting fn values as fast and not fn rules?

The problem is that values with jss-expand can be complex too, for e.g. `margin: {top, left}`, so you will have a painful inconsistency if you make them fast by default.

HenriBeck: No. not make them fast by default but instead only allowing 'fast' functions as function values.

packages/jss/src/RuleList.js
HenriBeck: 👍

packages/jss-plugin-syntax-rule-value-function/src/index.js
kof: same thing, no? `margin: () => (top, left)`

kof:
Ok I am done, please take a look again

kof:
Lets  review and merge this one for now, there are 2 issues we need to think how to solve separately.

changelog.md
```diff
@@ -2,6 +2,12 @@
 
 - Fix multiple cases where linking CSS rules didn't work (#815, #710, #664)
 
+### Breaking changes
+
+- Observables, function values and rules are now standalone packages, not part of the core. They are still part of the default preset though.
+- Function values and rules apply plugins by default now, which means they are slower by default. To speed them up use `sheet.update(data, {process: false})` (#682)
+- Observables will not apply plugins by default now, to make them even faster. If you want them to use plugins, can use an option when using the plugin: `jss.use(pluginObservable({process: true}))` (#682)
```
kof: https://github.com/cssinjs/jss/issues/746#issuecomment-429647786

packages/jss/src/RuleList.js
```diff
@@ -127,16 +143,64 @@ export default class RuleList {
...
+    const {style} = styleRule
+
+    plugins.onUpdate(data, rule, sheet, options)
+
+    // We rely on a new `style` ref in case it was mutated during onUpdate hook.
+    if (options.process && style && style !== styleRule.style) {
+      // We need to run the plugins in case new `style` relies on syntax plugins.
+      plugins.onProcessStyle(styleRule.style, styleRule, sheet)
+      // Update, remove and add props.
+      let nextValue
```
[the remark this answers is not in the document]
kof: we do, I was thinking we should avoid creating more variables in hot places to reduce garbage collection, but on the other hand, transpiled variable will be hoisted anyways, so there is no difference and we can do it in a more readable way

packages/jss/src/RuleList.js
```diff
@@ -127,16 +143,64 @@ export default class RuleList {
...
-    if (typeof name === 'string') {
-      plugins.onUpdate(data, this.get(name), sheet)
-    } else {
-      for (let index = 0; index < this.index.length; index++) {
-        plugins.onUpdate(name, this.index[index], sheet)
+
+    const styleRule: StyleRule = (rule: any)
+    const {style} = styleRule
+
+    plugins.onUpdate(data, rule, sheet, options)
```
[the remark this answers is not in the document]
kof: dunno, onUpdate doesn't mean onUpdateStyles, it means user called .update(), so not sure what would be a good return there, thats why I didn't return anything

kof:
🎉🎉🎉

Issue #882, opened 2018-10-11
Support Houdini Typed OM


kof:
Houdini typed values API can be [polyfilled](https://github.com/csstools/css-typed-om) with a fallback to strings, so we can actually start using in the latest versions and get perf benefits.

https://ishoudinireadyyet.com/
https://developers.google.com/web/updates/2018/03/cssom

### Current way: 
```js
  rule.style.margin = '5px';
  rule.style.margin // 5px
```
### New way: 
```js
rule.attributeStyleMap.set('margin', CSS.px(10))
rule.attributeStyleMap.get('margin') // CSSUnitValue {value: 10, unit: 'px'}
```

We can easily feature test it and use the new API always when supported. We don't care about all the CSS.whatever methods, they should be polyfilled in a separate product, we should just use `attributeStyleMap` when available, it also supports old-style strings.

## Todo

- [x] add conditional to DomRenderer
- [x] add documentation

Issue #885, opened 2018-10-12
[jss-default-unit] use typed cssom values when possible


kof:
When user relies on default unit, we don't have to convert it to string, we can actually use cssom:

```js
{
  margin: 10
}
```

should convert to

```js
{
  margin: CSS.px(10)
}
```

PR #887, opened 2018-10-12
Typed CSSOM values support

kof:
#882 
Branch is based on https://github.com/cssinjs/jss/pull/878 so it needs to be merged first for a sane diff.


HenriBeck:
Could we maybe also use the `CSS.px` etc. functions for the default unit plugin?

HenriBeck:
Is this branch supposed to have all of the changes of the dynamic values PR or was it an accident?

kof:
> Could we maybe also use the CSS.px etc. functions for the default unit plugin?

totally, added a separate issue for it https://github.com/cssinjs/jss/issues/885

> Is this branch supposed to have all of the changes of the dynamic values PR or was it an accident?

Just wanted to avoid conflicts, since the other PR is going to be merged soon

PR #888, opened 2018-10-13
Scoped keyframes

kof:
#346


packages/jss/src/rules/KeyframesRule.js
```diff
@@ -20,6 +28,10 @@ export default class KeyframesRule implements ContainerRule {
 
   constructor(key: string, frames: Object, options: RuleOptions) {
     this.key = key
+    // TODO make it more robust
+    this.name = key.substr(this.type.length + 2)
+    const fakeRule: Rule = ({key: this.name}: any)
+    this.id = options.jss.generateClassName(fakeRule, options.sheet)
```
kof: this is not very nice, I actually want id generation to be outside of a rule and passed to it

kof:
Thought about this too as well as about a new interface. The problem here
is that key in this rule unusable for id, so I had to create a fake rule.

Alternatively the key can be changed to something usable for id, then
interface can stay.

packages/jss/src/plugins/rules.js
```diff
@@ -30,4 +31,27 @@ const plugins: $ReadOnlyArray<Plugin> = Object.keys(classes).map((key: string) =
   return {onCreateRule}
 })
 
+// Animation name ref replacer.
+plugins.push({
```
HenriBeck: Not a fan of this. I think we should rename all of the Rules to StylePlugin etc. and move this functionality to the KeyframesPlugin.

packages/jss/src/plugins/rules.js
```diff
@@ -30,4 +31,27 @@ const plugins: $ReadOnlyArray<Plugin> = Object.keys(classes).map((key: string) =
...
+  onProcessStyle: (style, rule, sheet) => {
+    if (rule.type !== 'style' || !sheet) return style
+
+    // We need to support camel case here, because this plugin runs before the camelization plugin.
+    const prop = 'animationName' in style ? 'animationName' : 'animation-name'
```
HenriBeck: I think we also need to support `animation: { name: 'fadeIn' }`

HenriBeck: and also the `animation: $keyframeId 150ms` shortcut

packages/jss/src/plugins/rules.js
kof: yeah, that would be cleaner, but also more hooks and more code, not sure its worth it

packages/jss/src/plugins/rules.js
kof: thought about it an decided to limit it for now, because it goes into difficult directions, adds more code and slows down.

In the first case you would introduce the knowledge from jss-expand plugin about new syntax, in the second case, you would have to always run a regex or indexOf on the animation property.

We already had to introduce camel case support in the core, which is suboptimal. In theory we could make this refs replacement a separate plugin, but then its not very logical that one can't reference keyframes from the same sheet without plugins.

kof: Making it a separate plugin and running after jss-expand would allow to support "animationName", because jss-expand can generate it

kof: Another thought I had a while ago: what if we extract all core rules into separate packages and plugins? Most people use preset anyways. A few people can also setup by themselves all the plugins. This would give us the flexibility with above mentioned plugin.

kof: Additional benefit: someone who doesn't need all the rules, doesn't have to include them. For e.g. we got a request to include @page rule and there are some more rarely used rules. Why not make them all separate packages, but included in default preset.

HenriBeck: Yeah, let's maybe make all of the rules plugins for now and then move them out maybe.

We could do a `KeyframeRulePlugin` and a `KeyframeReferencePlugin` then. I think this would be the best solution. 

This should allow us to support all of the mentioned case above without needing to introduce anything.

packages/jss/src/plugins/rules.js
HenriBeck: I think it's worth it.

packages/jss/src/plugins/rules.js
kof: by "mak all of the rules plugins" you mean just jss/src/plugins or making them separate packages each?

kof: lets do this in  a separate PR after the other PRs are merged, too many conflicts otherwise

HenriBeck: Okay, I think we need to apply the rule plugins last so that the keyframes name replace works on every possibility.

packages/jss/src/rules/KeyframesRule.js
```diff
@@ -1,11 +1,15 @@
...
+import warning from 'warning'
 import RuleList from '../RuleList'
 import type {CSSKeyframesRule, RuleOptions, ToCssOptions, ContainerRule} from '../types'
 
 const defaultToStringOptions = {
   indent: 1,
   children: true
 }
+
+const nameRegExp = /@keyframes\s(\w*)/
```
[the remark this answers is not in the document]
kof: agree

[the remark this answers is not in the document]
kof: dash is already there `[\w-]`

packages/jss/src/plugins/keyframesRule.js
```diff
@@ -79,6 +79,23 @@ export class KeyframesRule implements ContainerRule {
 
 const keyRegExp = /@keyframes\s+/
 
+/**
+ * Replace the reference for a animation name.
+ */
+function replaceRef(style, prop, keyframes) {
+  if (style[prop]) {
+    const ref = /\$([\w-]+)/.exec(style[prop])
```
kof: value could be an array

packages/jss/src/plugins/keyframesRule.js
```diff
@@ -79,6 +79,23 @@ export class KeyframesRule implements ContainerRule {
 
 const keyRegExp = /@keyframes\s+/
 
+/**
+ * Replace the reference for a animation name.
+ */
+function replaceRef(style, prop, keyframes) {
```
kof: I recently stopped using function declarations, for the same reasons as we don't use "var" + hoisting which allows writing inconsistent code with function after usage.

kof: https://twitter.com/oleg008/status/1052133439980363777

packages/jss/src/plugins/keyframesRule.js
HenriBeck: grrrr.....

packages/jss/src/plugins/keyframesRule.js
HenriBeck: I don't know, eslint should cache such bugs and I think the function declaration syntax is a bit nicer than using `const` with arrow functions

kof: Yeah, bugs is not much of a problem with eslint, its more consistency to me.

kof: I did some changes, will push in a minute

kof:
Now with arrays support its quite a lot of code, thats why I wanted to support animationName property only

packages/jss/src/plugins/keyframesRule.js
```diff
@@ -0,0 +1,131 @@
...
+
+const refRegExp = /\$([\w-]+)/
+
+/**
+ * Replace the reference for a animation name.
+ */
+const replaceRef = (style: JssStyle, prop: string, keyframes: KeyframesMap) => {
+  const value = style[prop]
+
+  if (Array.isArray(value)) {
```
kof: I am not sure we should support arrays here at all, if we would only support animationName, but if we do we def need to test, I already see this has  a bug

kof: As far as I understand animation-name position in animation property is fixed: https://www.w3.org/TR/css-animations-1/#animation

Please try in a codesandbox to be sure, but if so we don't need a loop.

packages/jss/src/Jss.js
```diff
@@ -2,9 +2,8 @@
 import isInBrowser from 'is-in-browser'
 import StyleSheet from './StyleSheet'
 import PluginsRegistry from './PluginsRegistry'
-import rulesPlugins from './plugins/rules'
 import sheets from './sheets'
-import StyleRule from './rules/StyleRule'
+import {plugins as defaultPlugins, StyleRule, KeyframesRule} from './plugins/index'
```
HenriBeck: Can you export the plugins as the default export maybe?

HenriBeck:
```suggestion
import {plugins as defaultPlugins, StyleRule, KeyframesRule} from './plugins'
```

HenriBeck: This is awesome

packages/jss/src/Jss.js
```diff
@@ -95,24 +99,28 @@ export default class Jss {
...
     const rule = createRule(name, style, ruleOptions)
 
+    if (!rule) return null
+
     if (!ruleOptions.selector && rule instanceof StyleRule) {
       rule.selector = `.${ruleOptions.generateClassName(rule)}`
     }
 
+    if (rule instanceof KeyframesRule) {
+      rule.id = ruleOptions.generateClassName(rule)
```
HenriBeck: This doesn't feel in the right place, same with setting the selector. Can we maybe move this to the dedicated rules?

HenriBeck:
Still not really a fan of the hooks with the different queues though.

packages/jss/src/Jss.js
kof: for some reason without index rollup complains and can't build, try yarn build

packages/jss/src/Jss.js
kof: You mean move this logic to the rule itself?

kof:
> Still not really a fan of the hooks with the different queues though.

You mean the way it is implemented or in general? Other ideas?

packages/jss/src/Jss.js
HenriBeck: yeah, I think this would be the logical place

kof: its been there some time ago, was not nice too

packages/jss/src/Jss.js
HenriBeck: weird. will try it.

packages/jss/src/Jss.js
kof: The way I think about rules is that they have no knowledge of generated selectors, they only get them from the outside, no matter generated or not

packages/jss/src/Jss.js
HenriBeck: yeah, rollup has some kind of problem 🤔

packages/jss/src/Jss.js
HenriBeck: I think rules should generate everything that they need to work. In the case of StyleRule and KeyframesRule now that would be a selector

HenriBeck:
Yeah, I would prefer it the old but just apply the default plugins always last.

packages/jss/src/Jss.js
kof: They don't really care if selector was generated or not. In cases where rules are inside of `@global` selector needs to be global, in other cases  selector needs to be generated.

kof:
> Yeah, I would prefer it the old but just apply the default plugins always last.

Its an implementation problem, jss.use() can be called at any point of time, if you want to apply default plgins last, you will have to sort every time jss.use is called. This would be another way, but for this I would need  to differentiate default plugins in the sort alrgorthm. Actually I could just sort by the queue and get rid of additional complexity of 2 dimensional queue.

kof:
> Yeah, or a default plugins registry maybe? This registry would be private, and those hooks would be called by the public registry.

Separate registries? This could work now, but not sure if it won't be a problem later, if we want to expose the registry for some reason.

kof:
Also PluginsRegistry is called from many places, I would have to make 2 maps inside PluginsRegistry, but thats exactly what I did.

kof:
I can sort on .use() call though, this allows to have one registry. Sorting on .use() call is possible because its not a hot call.

PR #889, opened 2018-10-13
Add TS type defs

HenriBeck:
This adds ts typings to all of our currently supported packages.

Still missing
- [x] react-jss


appsforartists:
Is there a reason this is different than what we've written for [`@types/jss`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/eb8b3eff61e3606fe812b65384190e7aba925fce/types/jss)?

At first glance, the `@types` ones seem more thorough.  It checks the names of the keys in `classes`, asserts that the styles matched the published CSS spec, etc.

If the jss maintainers want to own its typings, we should start with what's already in DefinitelyTyped.

Issue #880, opened 2018-10-10
Use WeakMap


kof:
We are using in many cases a generated property directly on an object/function when we need to associate a value with that object. I think WeakMap support is hight enough now that we can start using it and assume people will include a WeakMap polyfill for older browsers.

I think React is also using WeakMap.

## Todo
- [ ] use it in plugin-syntax-rule-value-function
- [ ] use it in plugin-cache
- [ ] update docs which refer required polyfils
- [ ] search where else it could be used

HenriBeck:
👍

HenriBeck:
WeakMap can't be used inside plugin-syntax-rule-value-function because a WeakMap needs objects as keys.

kof:
Function is an object, can be used as a key

HenriBeck:
Ohh yeah, so the function would be the keys and the actual css prop name would be the value in the WeakMap

kof:
@HenriBeck I think we can close this one, no?

HenriBeck:
No, not yet, we haven't converted the function plugin to use it.

PR #890, opened 2018-10-13
[jss-plugin-cache] Use WeakMap for cache plugin

HenriBeck:
__What would you like to add/fix?__ Use WeakMap for the cache plugin.

__Corresponding issue (if exists):__ #880


kof:
Need to add the polyfill requirement, I think here in the plugin docs.

PR #893, opened 2018-10-16
Use CSSTOM in default unit plugin

HenriBeck:
#885


packages/jss-plugin-syntax-default-unit/src/defaultUnits.js
```diff
@@ -1,114 +1,164 @@
...
-  'transform-origin-x': '%',
-  'transform-origin-y': '%',
-  'transform-origin-z': '%',
-  'transition-delay': 'ms',
-  'transition-duration': 'ms',
-  'vertical-align': 'px',
-  width: 'px',
-  'word-spacing': 'px',
+  // Animation properties
+  'animation-delay': getDefaultUnit(window.CSS.ms, 'ms'),
```
kof: I would prefer to not access window.CSS all over the place, because it might be not implemented and not polyfilled

kof: you can rewrite getDefautUnit to something like this:

```js
const hasCSSTOMSupport = window.CSS && window.CSS.number
const getDefaultUnit = (unit) => (hasCSSTOMSupport ? window.CSS[unit === '%' ? 'percent' : unit]() : unit)

```

kof: and then use it `getDefaultUnit('ms')`

packages/jss-plugin-syntax-default-unit/src/defaultUnits.js
```diff
@@ -1,7 +1,10 @@
 // @flow
 
-const hasCSSTOMSupport = window.CSS && window.CSS.number
-const getDefaultUnit = (cssomFN, fallback) => (hasCSSTOMSupport ? cssomFN : fallback)
+import {hasCSSTOMSupport} from 'jss'
+
+const px = hasCSSTOMSupport ? window.CSS.px : 'px'
+const ms = hasCSSTOMSupport ? window.CSS.ms : 'ms'
+const percent = hasCSSTOMSupport ? window.CSS.percent : '%'
```
kof: nice

PR #897, opened 2018-10-17
Update readme.md

davidprae:
Fixes JS syntax issue in code block

JS syntax issue in code block

NA


kof:
merged, thanks!

PR #900, opened 2018-10-23
Fixed cannot find top level reference issue inside @media query

### LINKED ISSUE #848: [jss-nested] Can't find referenced top-level rules inside of @ rules
__Expected behavior:__

I was trying to reference `$foo`-style class name references media queries like the following and hoping it would work. ([codesandbox reproduction](https://codesandbox.io/s/wyplxrr7lw))

Here is the raw object I'm using with react-jss:

```js
{
  mapperLogo: {
    display: "block",
    position: "absolute",
    top: "50%",
    left: 0,
    transform: "translate(0,-50%)",
    "& img": {
      display: "block"
    },
    "@media (max-width: 69.686875em)": {
      paddingTop: "0.4375rem",
      marginLeft: "1.61875rem",
      "& img": {
        width: "12.5rem"
      }
    },
    "@media (min-width: 69.6875em)": {
      marginLeft: "2.3125rem",
      marginTop: "0.1875rem",
      zIndex: "1000",
      "& img": {
        height: "1.6875rem"
      }
    }
  },
  header: {
    position: "fixed",
    top: 0,
    left: 0,
    right: 0,
    width: "100%",
    backgroundColor: "#f1f1f1",
    zIndex: 100,
    "@media (max-width: 69.686875em)": {
      height: "6.25rem",
      "& $menuItem": {
        marginLeft: "1.25rem",
        marginRight: "1.25rem",
        float: "right"
      },
      "& $menu": {
        margin: 0,
        position: "absolute",
        top: "50%",
        right: 0,
        transform: "translate(0%,-50%)",
        "& $menuItem": {
          display: "none"
        },
        "& $ham": {
          display: "block",
          padding: [["0.875rem", "1.61875rem"]],
          "& img": {
            width: "3.5rem"
          }
        }
      }
    },
    "@media (min-width: 69.6875em)": {
      height: "3.125rem",
      textAlign: "left",
      "& $menu": {
        float: "right",
        margin: 0,
        position: "absolute",
        top: "50%",
        right: "1.25rem",
        transform: "translate(0%,-50%)",
        display: "flex",
        fontSize: "1.25rem",
        "& $menuItem": {
          marginLeft: "2.5rem",
          marginRight: "2.5rem",
          display: "block",
          color: "#898989",
          fontWeight: "bold",
          textDecoration: "none",
          fontFamily: "SF, sans-serif",
          textAlign: "center",
          letterSpacing: "0.0625rem",
          "&:hover, &:active, &:focus, &$currentPage": {
            fontWeight: "bold",
            fontFamily: "SFBold",
            letterSpacing: 0,
            color: "black"
          }
        },
        "& $ham": {
          display: "none"
        }
      }
    },
    opacity: 1,
    transition: "opacity 200ms",
    "@supports ( -webkit-backdrop-filter: blur(40px) ) or ( backdrop-filter: blur(40px) )": {
      opacity: 0.8
    },
    backdropFilter: "blur( 2.5rem )",
    "&.faded": {
      opacity: 0.8,
      transition: "opacity 200ms"
    }
  },
  menu: {},
  menuItem: {},
  currentPage: {},
  ham: {}
}
```

__Describe the bug:__

I get some errors like these:

screen shot 2018-09-10 at 3 29 48 pm

__Codesandbox link:__ 

None yet.

 - jss: 9.8.7

---

Any ideas? Do you see anything obviously wrong? Are top-level `@` rules with references supported?

sammi:
Fixed #848


packages/jss-plugin-syntax-nested/src/index.test.js
```diff
@@ -637,4 +637,32 @@ describe('jss-plugin-syntax-nested', () => {
       `)
     })
   })
+
+  describe('nest rules inside media query', () => {
+    let sheet
+
+    beforeEach(() => {
+      const localJss = create(settings).use(jssExtend(), nested())
```
HenriBeck: I don't think we need jss-extend in this test. Please remove

HenriBeck:
Please do not try to trigger anymore builds. Your build is never going to succeed for the simple reason that we don't export private env vars to forks.

packages/jss-plugin-syntax-nested/src/index.test.js
sammi: cool, it makes sense.

sammi:
I would like join this project if you guys are still looking for contributors.

HenriBeck:
I fixed the expectation of your test and adjusted the code to pass it.

packages/jss-plugin-syntax-nested/src/index.test.js
```diff
@@ -637,4 +637,32 @@ describe('jss-plugin-syntax-nested', () => {
...
     })
   })
+
+  describe('nest rules inside media query', () => {
+    let sheet
+
+    beforeEach(() => {
+      const localJss = create(settings).use(nested())
+      sheet = localJss.createStyleSheet({
+        card: {},
```
kof: please use simplified names like `"a": {}, b: {}` for consistency with the rest

packages/jss-plugin-syntax-nested/src/index.test.js
```diff
@@ -637,4 +637,32 @@ describe('jss-plugin-syntax-nested', () => {
       `)
     })
   })
+
+  describe('nest rules inside media query', () => {
+    let sheet
+
+    beforeEach(() => {
+      const localJss = create(settings).use(nested())
```
kof: it seems there is no ned to set up localJss instance, the one defined in the beginning has exact same setup, you can just use `jss` ...  we used a new one because it needed a different setup

packages/jss-plugin-syntax-nested/src/index.js
```diff
@@ -87,7 +88,7 @@ export default function jssNested(): Plugin {
         let selector = replaceParentRefs(prop, styleRule.selector)
         // Lazily create the ref replacer function just once for
         // all nested rules within the sheet.
-        if (!replaceRef) replaceRef = getReplaceRef(container)
+        if (!replaceRef) replaceRef = getReplaceRef(container, sheet)
```
kof: My comment before was overwritten here, so we can make sure above that container is either container rule or a sheet, without using a second argument, getRule method is on both.

HenriBeck: Should we maybe always use the sheet to only ever reference rules from the top level?

packages/jss-plugin-syntax-nested/src/index.js
```diff
@@ -67,7 +68,7 @@ export default function jssNested(): Plugin {
     }
   }
 
-  function onProcessStyle(style, rule) {
+  function onProcessStyle(style, rule, sheet?: StyleSheet) {
```
sammi: why we add third argument sheet here?

packages/jss-plugin-syntax-nested/src/index.js
sammi: I'm wondering why we need change onProcessStyle(style, rule) to onProcessStyle(style, rule, sheet?: StyleSheet)?  It seems sheet value is not an input, could get got from the first 2 argument.

HenriBeck: It is a valid argument. It's the top level sheet which I think we want here.

kof: actually I think you want always the sheet, we should only support referencing top level rules

kof: looking at this example, this would be unexpected I think, so it should b container or sheet ...

kof:
merged 🎉

emirotin:
I've encountered the same issue now with both v9 and the most recent v10 alpha
I was able to workaround it by moving the media query to the deepest level and having the nested selector in the higher-level key, like this (I'm using MUI and react-jss):

```
const styles = theme => {
    const floatingStyles = { ... };

    const floatingArrowStyles = { ... };

    return {
        error: {},
        errorArrow: {},
        floatingAuto: {},
        floatingNever: {},
        floatingAlways: {},
        root: {
            '&$floatingAlways': floatingStyles,
            '&$floatingAlways $errorArrow': floatingArrowStyles,
            '&$floatingAuto': {
                [theme.breakpoints.up('sm')]: floatingStyles
            },
            '&$floatingAuto $errorArrow': {
                [theme.breakpoints.up('sm')]: floatingArrowStyles
            }
        }
    };
};
```

But it didn't work like this

```
const styles = theme => {
    const floatingStyles = {
        ...,
        '& $errorArrow': { ... }
    };

    return {
        error: {},
        errorArrow: {},
        floatingAuto: {},
        floatingNever: {},
        floatingAlways: {},
        root: {
            '&$floatingAlways': floatingStyles,
            '&$floatingAuto': {
                [theme.breakpoints.up('sm')]: floatingStyles
            }
        }
    };
};
```

Strange enough I cannot repro it with REPL, there things just work.

kof:
@emirotin feel free create a separate issue with reproduction

emirotin:
Need to isolate it, will try

PR #905, opened 2018-10-26
Ignore undefine, null, or empty value rules when generate style.

### LINKED ISSUE #899: [jss-plugin-syntax-global] TypeError: Cannot read property '@global' of undefined
__Expected behavior:__

No errors. 

__Describe the bug:__

I can't create a stylesheet with an undefined rule. This can happen in Material-UI where we create CSS directly from the theme. 

__Codesandbox link:__ 

```jsx
import jss from "jss";
import presetDefault from "jss-preset-default";

jss.setup(presetDefault());
const s = jss.createStyleSheet({
  tutu: undefined
});
s.attach();
export default jss;
```

https://codesandbox.io/s/lyz1kryzl9

 - jss: 9.8.7

sammi:
Fix #899

### DIFF
--- a/packages/jss-plugin-syntax-global/src/index.test.js
+++ b/packages/jss-plugin-syntax-global/src/index.test.js
@@ -19,7 +19,17 @@ describe('jss-global', () => {
   describe('@global rule with null, undefined or empty value', () => {
     let sheet
 
-    beforeEach(() => {
+    it('should generate correct CSS for one rule', () => {
+      sheet = jss.createStyleSheet({
+        '@global a': {
+          color: 'red'
+        }
+      })
+      expect(sheet.getRule('a')).to.be(undefined)
+      expect(sheet.toString()).to.be('a {\n  color: red;\n}')
+    })
+
+    it('should generate correct CSS without empty values', () => {
       sheet = jss.createStyleSheet({
         '@global': {
           a: null,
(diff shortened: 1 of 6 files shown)


packages/jss-plugin-syntax-global/src/index.test.js
```diff
@@ -16,6 +16,25 @@ describe('jss-global', () => {
     jss = create(settings).use(global())
   })
 
+  describe('@global rule with null, undefined or empty value', () => {
```
HenriBeck: Can you also test `@global a` with `undefined` rule?

sammi: I guess b: undefined should cover it?

kof: What @HenriBeck describes is  a prefixed rule, what you are testing is a container rule:
```js
{
  '@global a': {color: 'red'}
}
```

sammi: Test it on sanbox, it's working.  

export default {
  '@global a': {color: 'red'}
};

Although not existing issue, I added it as a regression test in the PR.

sammi: oh, wait. @HenriBeck , what the "prefixed rule" test case looks like?

kof:
```js
'@global a': undefined
'@global b': null,
```

it might be already working, just to make sure we cover that case.

sammi: it is working. I added those test cases as well to protect more such empty value edge cases.

DurgaPammi:
how can we right the JSS for the below example :

<Collapsible trigger="Start here">
    <p>This is the collapsible content. It can be any element or React component you like.</p>
  </Collapsible>
CSS is:

.Collapsible__trigger {
display: block;
font-weight: 600;
text-decoration: none;
color: #ffffff;
position: relative;
border: 1px solid white;
padding: 10px;
background: #f3f4f4;
color: black;
}
.Collapsible__trigger:after {
font-family: 'FontAwesome';
content: '\f107';
position: absolute;
right: 10px;
top: 10px;
display: block;
transition: transform 300ms;
}
.Collapsible__trigger.is-open:after {
transform: rotateZ(180deg);
}
.Collapsible__trigger.is-disabled {
opacity: 0.5;
background-color: grey; }

Can You please any one help me about this

PR #906, opened 2018-10-27
Don't use undefined as a selector 

kof:

Currently when generateId is called, rule.selector is not set yet. If we try to stringify the rule, it will be somthing like `undefined {color: red;}`.

With this PR we are only getting the stringified styles without a selector if its not set.


HenriBeck:
Does it make sense to render a rule when it has no selector? Rendering only stringified styles might break some stylesheets then.

kof:
Its not for rendering. It is when you call toString on a rule in generateId
in case you want a hash based id.

PR #911, opened 2018-11-03
[react-jss] Update theming

HenriBeck:
Update to the new theming version.


packages/react-jss/tests/theming.js
```diff
@@ -244,7 +245,7 @@ describe('React-JSS: theming', () => {
       node
     )
 
-    expect(document.querySelectorAll('style').length).to.equal(1)
+    expect(document.querySelectorAll('style').length).to.equal(2)
```
HenriBeck: I think to make those changes here make sense as this is only a behavior when `ThemeProvider` aren't nested. As soon as you nest one of them, this will not work anymore.

kof: not sure what you mean, why is the amount of sheets changed here?

HenriBeck:
```js
const theme = { color: 'green' };
// This will provide only create sheet
const App1 = (
  <React.Fragment>
    <ThemeProvider theme={theme}>
      <ThemedComp />
    </ThemeProvider>
    <ThemeProvider theme={theme}>
      <ThemedComp />
    </ThemeProvider>
  </React.Fragment>
);

// This will create two sheets
const App2 = (
  <React.Fragment>
    <ThemeProvider theme={theme}>
      <ThemeProvider theme={theme}>
        <ThemedComp />
      </ThemeProvider>
    </ThemeProvider>
    <ThemeProvider theme={theme}>
      <ThemeProvider theme={theme}>
        <ThemedComp />
      </ThemeProvider>
    </ThemeProvider>
  </React.Fragment>
);
```

kof: I still don't understand, even the test name says if its the same theme and styles are static 2 components generate only one style tag

HenriBeck: Yeah, but the principle of the same sheet only applies for ThemeProviders which are on the Top-Level and have no nested ThemeProviders. See my second example, I nest to ThemeProvider both with the same theme but we will still create 2 sheets

packages/react-jss/src/createHoc.js
```diff
@@ -183,23 +183,21 @@ export default function createHOC<
...
     }
 
-    getDynamicSheet(staticSheet) {
+    getDynamicSheet(staticSheet: StyleSheet): StyleSheet | void {
       // $FlowFixMe Cannot access random fields on instance of class StyleSheet
       const dynamicStyles = staticSheet[dynamicStylesNs]
 
-      if (dynamicStyles) {
-        const contextSheetOptions = this.context[ns.sheetOptions]
+      if (!dynamicStyles) return undefined
```
HenriBeck: this could just be `return;` right?

kof: yeah, we need to change our eslint config for this, it requires explicite return value right now

HenriBeck: Ah ok, yeah makes sense to enforce explicit return value.

PR #914, opened 2018-11-07
remove nag message from postinstall script

### LINKED ISSUE #881: Please remove postinstall message about donating
While I understand the need to ask for donations, having it as the [postinstall message](https://github.com/cssinjs/jss/blob/master/packages/jss/package.json#L42), to me, isn't the most ideal location.

Downsides of having it there:
- It clutters up server logs (it does not respect the `--quiet` flag
- It is excessive (4 lines long)
- It sets a bad precedent/example. If every package did this, there would be hundreds of these messages.

This type of message is better suited to be at the top of the repo (hint hint), in the readme file and on the website.

slang800:
closes #881

I don't know what JSS is (it was installed as a dep of another dep in a project I'm maintaining), but this message shows up in the logs of my CI for every single run and it's getting annoying.

### DIFF
--- a/packages/jss/package.json
+++ b/packages/jss/package.json
@@ -36,8 +36,7 @@
   ],
   "scripts": {
     "build": "node ../../scripts/build.js",
-    "check-snapshot": "node ../../scripts/match-snapshot.js",
-    "postinstall": "node -e \"console.log('\\u001b[35m\\u001b[1mLove JSS? You can now support us on open collective:\\u001b[22m\\u001b[39m\\n > \\u001b[34mhttps://opencollective.com/jss/donate\\u001b[0m')\""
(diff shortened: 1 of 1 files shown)


kof:
The suggestion was to use https://github.com/cssinjs/jss/issues/881#issuecomment-428726121

slang800:
I think my solution is far more elegant. It requires zero environment variables and zero additional deps :wink:

kof:
Does it also has integrated reminder for the user to support software where some other people spend a lot of their private time on instead of spending time with their families?

kof:
TBH I don't care about it, because it simply doesn't bring anything. A few contributors came probably not over this channel, so I am just going to merge it.

Issue #917, opened 2018-11-14
Un-mount and re-mount JssProvider


scaredcat:
__Expected behavior:__
Expect dynamically generated style names and statically generated style names to be unique

__Describe the bug:__
I have a component wrapped by a JssProvider and  Theme provider eg.

```javasdcript
const jss = createJss()
jss.use(global(), nested(), extend(), compose(), camelCase(), vendorPrefixer())

      <JssProvider classNamePrefix={this.props.classNamePrefix} jss={jss}>
        <ThemeProvider theme={this.props.theme}>
          {this.props.children}
        </ThemeProvider>
      </JssProvider>
```

When I unmount this component and re-mount this component, any dynamically generated styles in the child components now have style names that clash with the original statically generated styles.

__Codesandbox link:__ 
Only happens in production. Will try to write an example soon

__Versions:__
 - jss: 
    "jss": "^9.8.1",
    "jss-camel-case": "^6.1.0",
    "jss-compose": "^5.0.0",
    "jss-global": "^3.0.0",
    "jss-nested": "^6.0.1",
    "jss-vendor-prefixer": "^8.0.1",
 - Browser [e.g. chrome, safari]: latest Chrome (Version 70.0.3538.77 (Official Build) (64-bit)
), latest Firefox (63.0 (64-bit))
 - OS [e.g. Windows, macOS]: Ubuntu 18.10

HenriBeck:
It most likely has to do something with your generateClassName function I would guess. But without seeing more code I can’t tell with 100% certainty.

Once you are able to reproduce the issue on codesandbox we will definetly have a look at it.

scaredcat:
Thanks for the quick reply. I've re-produced the issue here https://github.com/scaredcat/jss-class-generator-test
Since it requires a production build, I've deployed it using gh-pages:
https://scaredcat.github.io/jss-class-generator-test

pressing toggle unmounts the component, then remounts it when toggled back. Notice that the dynamic classnames applied to the third button now is the same as the static classname of the second button group.

The issue does not occur when running the code in development mode.

HenriBeck:
In general I would advise against remounting the JssProvider. I don't think we can fix this bug, because it also happens when you don't use a custom jss instance.

HenriBeck:
I think we should add a section in the docs to never remount the `JssProvider` so there are no class name conflicts.

kof:
Can't we warn instead? I feel like docs won't help there ...

HenriBeck:
[...]

PR #919, opened 2018-11-16
Ensure class name generator is re-used when mounting and unmounting jss providers

scaredcat:
JssProvider creates a new classsname generator each time it is mounted.
When a local JSS thing is available, use the already generated function.

Not sure if this is the right approach, as I think `localJss.options.createGenerateId` should be singleton where `ruleCounter` is consistent between ID generators
https://github.com/cssinjs/jss/blob/0abd98b65abb521b7b213cbc3218fbda76ca3db5/packages/jss/src/utils/createGenerateId.js#L17

#917

### DIFF
--- a/packages/react-jss/src/JssProvider.js
+++ b/packages/react-jss/src/JssProvider.js
@@ -67,13 +67,18 @@ export default class JssProvider extends Component<Props> {
       sheetOptions.generateId = generateId
     } else if (!sheetOptions.generateId) {
       if (!this.generateId) {
-        let createGenerateId = createGenerateIdDefault
-        if (localJss && localJss.options.createGenerateId) {
-          createGenerateId = localJss.options.createGenerateId
+        if (localJss && localJss.generateClassName) {
+          // re-use the same classname generator pinned to the local jss instance
+          this.generateClassName = localJss.generateClassName
+        } else {
+          let createGenerateId = createGenerateIdDefault
+          if (localJss && localJss.options.createGenerateId) {
+            createGenerateId = localJss.options.createGenerateId
+          }
+          // Make sure we don't loose the generator when JssProvider is used inside of a stateful
+          // component which decides to rerender.
+          this.generateId = createGenerateId()
         }
-        // Make sure we don't loose the generator when JssProvider is used inside of a stateful
-        // component which decides to rerender.
-        this.generateId = createGenerateId()
       }
 
       sheetOptions.generateId = this.generateId


packages/react-jss/src/JssProvider.js
```diff
@@ -67,13 +67,18 @@ export default class JssProvider extends Component<Props> {
       sheetOptions.generateId = generateId
     } else if (!sheetOptions.generateId) {
       if (!this.generateId) {
-        let createGenerateId = createGenerateIdDefault
-        if (localJss && localJss.options.createGenerateId) {
-          createGenerateId = localJss.options.createGenerateId
+        if (localJss && localJss.generateClassName) {
+          // re-use the same classname generator pinned to the local jss instance
+          this.generateClassName = localJss.generateClassName
```
HenriBeck: This should be `this.generateId = localJss.generateId`

HenriBeck:
Can you also please add a test for this?

packages/react-jss/src/JssProvider.js
```diff
@@ -67,13 +67,18 @@ export default class JssProvider extends Component<Props> {
...
+        } else {
+          let createGenerateId = createGenerateIdDefault
+          if (localJss && localJss.options.createGenerateId) {
+            createGenerateId = localJss.options.createGenerateId
```
HenriBeck: I don't think we need this anymore. When a local jss instance is passed, it will always have a `generateId` and this check should be redundant.

packages/react-jss/src/JssProvider.js
scaredcat: localJss doesn't seem to have a generateId property at this stage.

screenshot from 2018-11-20 09-40-13

HenriBeck: Yes, this is true for jss version below v10. We are currently in the process of polishing jss v10, and in this release, we renamed `generateClassName` to `generateId`

HenriBeck: If you run `yarn typecheck` you will some errors which correspond to that.

packages/react-jss/src/JssProvider.js
HenriBeck: @kof what was the reason for this? Can't seem to have a use case where it makes sense? I think if the user passes in a jss instance, we should use the `generateId` instead of calling the `createGenerateId` again, as this might lead to duplicate classes.

kof: I think we didn't have a cached generateId() on the jss instance back then, try to remove and see if any tests indicate a problem

HenriBeck:
This will be done inside the #924 PR

PR #924, opened 2018-11-27
[WIP] Update to new react context for the jss context

HenriBeck:
Move from the old context to the new context.

We might need to consider to export the consumer and provider so that other libaries can integrate into it.


HenriBeck:
@kof 

I checked again our options on the issue with the `generateId` from the `jss` instance.

In the case that we support it, we need to require the user to generate a new jss instance on every request when he wants to use a custom one.

We could also not support this and simply requiring if the user wants a custom `generateId` function, he will need to use the prop.


In my opinion, we should support it and require the user to create a new jss instance for every request. This is also a safer practice so that requests don't share anything.

packages/react-jss/src/JssProvider.js
```diff
@@ -18,65 +20,79 @@ type Props = {
...
-      // This way we identify a new request on the server, because user will create
-      // a new Registry instance for each.
-      if (registry !== this.registry) {
-        // We reset managers because we have to regenerate all sheets for the new request.
-        this.managers = {}
-        this.registry = registry
+    if (sheetOptions) {
+      warning(
+        'generateId' in sheetOptions,
+        '[JssProvider] Do not pass the generateId inside the sheet options! We have a special prop for that!'
```
kof: I think we always used [jss] so far, I am not sure what the convention should be, but we should decide on that first. For e.g. it could be package name.

A better wording would be "Pass generateId as a prop instead of sheet options."

kof: I think we should actually stick with [JSS] just to let user know what project this warning is related to, they might not know the package name anyways

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
   disableStylesGeneration?: boolean,
+  sheetOptions: StyleSheetFactoryOptions,
   children: Node
 }
 
 export default class JssProvider extends Component<Props> {
   static propTypes = {
-    ...propTypes,
+    registry: PropTypes.instanceOf(SheetsRegistry),
+    jss: PropTypes.shape({}),
```
kof: That should be Jss instance, right?

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
     generateId: PropTypes.func,
     classNamePrefix: PropTypes.string,
     disableStylesGeneration: PropTypes.bool,
-    children: PropTypes.node.isRequired
+    children: PropTypes.node.isRequired,
+    sheetOptions: PropTypes.shape({})
```
kof: isn't it StyleSheetFactoryOptions?

kof:
Can you please list all changes from user perspective? We need a changelog

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
   }
 
-  static childContextTypes = contextTypes
+  managers: {} = {}
 
-  static contextTypes = contextTypes
+  generateId = createGenerateId()
 
-  // JssProvider can be nested. We allow to overwrite any context prop at any level.
```
kof: So we don't support overriding a specific option any more from a particular provider, right? I think this is a breaking change. We need to document etc.

packages/react-jss/src/JssProvider.js
HenriBeck: yeah, but we don't export the Jss class

packages/react-jss/src/JssProvider.js
HenriBeck: Yes, should we copy the definition? I'm using StyleSheetFactoryOptions in the flow types

packages/react-jss/src/JssProvider.js
HenriBeck: We do.

packages/react-jss/src/JssProvider.js
HenriBeck: 👍

packages/react-jss/src/injectSheet.test.js
```diff
@@ -75,23 +75,6 @@ describe('React-JSS: injectSheet', () => {
...
-          <JssProvider>
-            <MyComponent />
-          </JssProvider>
-        </div>,
-        node
-      )
-      expect(document.querySelectorAll('style').length).to.be(1)
-      unmountComponentAtNode(node)
-      expect(document.querySelectorAll('style').length).to.be(0)
-    })
```
HenriBeck: I had to remove this test. In my opinion, it doesn't make sense, because if you add a `classNamePrefix` to the first provider, the second provider with the styled components will also end up using the `classNamePrefix`

Example: https://codesandbox.io/s/r5j1ynwnwm

packages/react-jss/src/injectSheet.test.js
```diff
@@ -338,85 +323,44 @@ describe('React-JSS: injectSheet', () => {
...
-
-      const style = document.querySelectorAll('style')[0]
-      expect(removeWhitespaces(style.innerText)).to.be(
-        removeWhitespaces(`
-        .a-id {
-          right: 2px;
-        }
-      `)
-      )
-    })
```
HenriBeck: This was a pretty unstable behavior because of context blocking.
We would need to track a few more things inside the HoC for this to be implemented properly.

packages/react-jss/src/JssProvider.js
kof: can we export Jss as a type only?

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
-    children: PropTypes.node.isRequired
+    children: PropTypes.node.isRequired,
+    sheetOptions: PropTypes.shape({})
   }
 
-  static childContextTypes = contextTypes
+  managers: {} = {}
 
-  static contextTypes = contextTypes
+  generateId = createGenerateId()
```
kof: Now you will create a new generator every time user renders a new provider, right? Is it what you wanted to achieve?

packages/react-jss/src/JssProvider.js
kof: ah, we are in proptypes, so we would need a class here ... not sure whats best, mb we should export it, like we did with some others but not document.

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
+
+    // Merge the classname prefix
+    if (classNamePrefix) {
+      context.sheetOptions.classNamePrefix =
+        classNamePrefix + (context.sheetOptions.classNamePrefix || '')
+    }
 
-      sheetOptions.generateId = this.generateId
+    if (jss) {
+      context.jss = jss
```
kof: what about jss on outerContextt?

packages/react-jss/src/injectSheet.test.js
kof: Yeah, reusing styles from different providers will only work if they have identic settings and in that case why there are different providers.

packages/react-jss/src/injectSheet.test.js
kof: I think the intention was to verify that we use jss instance that was passed to the provider, did we verify it somewhere else as well?

packages/react-jss/src/types.js
```diff
@@ -18,15 +18,16 @@ export type InnerProps = {
 }
 // Needs to be hard coded for stricter types
 export type Context = {
-  '64a55d578f856d258dc345b094a2a2b3'?: Jss,
-  d4bd0baacbc52bbd48bbb9eb24344ecd?: SheetsRegistry,
-  b768b78919504fba9de2c03545c5cd3a?: {[key: number]: SheetsManager},
-  '6fc570d6bd61383819d0f9e7407c452d': StyleSheetFactoryOptions & {disableStylesGeneration?: boolean}
+  jss?: Jss,
+  registry?: SheetsRegistry,
+  managers?: {[key: number]: SheetsManager},
```
kof: we should use the same type in other places where we use managers

packages/react-jss/tests/theming.js
```diff
@@ -43,15 +48,11 @@ describe('React-JSS: theming', () => {
   const ThemedDynamicComponent = injectSheet(themedDynamicStyles)()
 
   let localJss
+  let generateId
 
   beforeEach(() => {
```
kof: somehow we ended up having 2 beforeEach in one describe, I think it wasn't intentional

packages/react-jss/src/createHoc.js
```diff
@@ -84,39 +83,36 @@ export default function createHOC<
...
     static propTypes = {
       innerRef: PropTypes.func
     }
 
-    static defaultProps = {
-      ...defaultProps,
-      theme: noTheme
+    static defaultProps = defaultProps
+
+    static getTheme(props) {
```
kof: do you need this one to be static property of the component? or can this be  just a function scoped to the module

packages/react-jss/src/JssProvider.js
kof: Sorry, didn't realize we are in proptypes, yea we should reuse proptypes for sheetOptions

packages/react-jss/src/JssProvider.js
HenriBeck: It should be possible to overwrite jss correct?

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
-      sheetOptions.disableStylesGeneration = disableStylesGeneration
+      context.disableStylesGeneration = disableStylesGeneration
     }
 
     return context
   }
 
-  registry: SheetsRegistry
+  renderProvider = (outerContext: Context) => {
+    const {children} = this.props
```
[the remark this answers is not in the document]
HenriBeck: It works because of the arrow function

packages/react-jss/src/createHoc.js
HenriBeck: No, it could be moved outside

packages/react-jss/src/JssProvider.js
kof: I thought it was possible before? At least it would make sense to me right now

packages/react-jss/src/JssProvider.js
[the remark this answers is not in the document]
kof: sorry, overlooked it

packages/react-jss/src/createHoc.js
kof: I would do it to avoid making it public

packages/react-jss/src/injectSheet.test.js
HenriBeck: Yeah we do that in a different test

packages/react-jss/src/JssProvider.js
HenriBeck: I will create a generator for every rendered instance.

packages/react-jss/src/JssProvider.js
HenriBeck: Now I'm confused. It should be possible to overwrite the jss instance, correct?

packages/react-jss/src/types.js
HenriBeck: We do.

packages/react-jss/src/JssProvider.js
kof: I get it now, we override in the inner provider the instance on the context if one was passed higher up. Its the same beehavior as it was before

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,88 @@ type Props = {
...
     generateId: PropTypes.func,
     classNamePrefix: PropTypes.string,
     disableStylesGeneration: PropTypes.bool,
-    children: PropTypes.node.isRequired
+    children: PropTypes.node.isRequired,
+    sheetOptions: PropTypes.shape({})
   }
 
-  static childContextTypes = contextTypes
+  managers: {} = {}
```
kof: I ment we could use the same here {[key: number]: SheetsManager} or even better reference a type Managers

packages/react-jss/src/JssProvider.js
kof: yes but that will mean that 2 instances will generate conflicting ids

packages/react-jss/src/JssProvider.js
kof: we could use jss instance to access the constructor as well, I think this should work: PropTypes.instanceOf(jss.constructor)

packages/react-jss/src/JssProvider.js
kof: We should have a correct shape still

packages/react-jss/tests/theming.js
HenriBeck: I checked, there are not two generateId's

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +20,92 @@ type Props = {
...
-        if (localJss && localJss.options.createGenerateId) {
-          createGenerateId = localJss.options.createGenerateId
-        }
-        // Make sure we don't loose the generator when JssProvider is used inside of a stateful
-        // component which decides to rerender.
-        this.generateId = createGenerateId()
-      }
+      context.sheetOptions.generateId = generateId
+    } else if (!context.sheetOptions.generateId) {
+      this.generateId = createGenerateId()
```
kof: why do we assign generateId to the instance? we don't seem to be using it from the instance

packages/react-jss/src/JssProvider.js
HenriBeck: Done.

kof:
We completely forgot the changelog and docs. Since docs are in refactoring, lets at least add to the changelog here, so we can't forget update the docs later

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +13,80 @@ type Props = {
   generateId?: GenerateId,
   classNamePrefix?: string,
   disableStylesGeneration?: boolean,
+  media?: string,
```
kof: changelog

packages/react-jss/src/JssProvider.js
HenriBeck: Fixed.

packages/react-jss/src/JssProvider.js
```diff
@@ -20,85 +13,80 @@ type Props = {
...
-  registry: SheetsRegistry
+  generateId: GenerateId
 
-  managers: {}
+  renderProvider = (outerContext: Context) => {
+    const {children} = this.props
 
-  generateId: GenerateId
+    return (
+      <JssContext.Provider value={this.createContext(outerContext)}>{children}</JssContext.Provider>
```
kof: what should user should know from changelog in this regard?

HenriBeck: That's just an internal change. Also mentioned the switch to the new React context in the changelog

packages/react-jss/src/JssProvider.js
HenriBeck: Done.

packages/react-jss/src/createHoc.js
```diff
@@ -137,16 +128,12 @@ export default function createHOC<
...
-    }
-
     get jss() {
-      return this.context[ns.jss] || optionsJss || jss
+      return this.props.jss || optionsJss || jss
     }
 
     get manager(): SheetsManager {
-      const managers = this.context[ns.managers]
+      const {managers} = this.props
```
kof: why are we using props now instead of context?

kof: oh we pass everything from context as props ... hmm this would mean props can override our context.

HenriBeck: Yeah, but we could just simply put the context after the props.

kof: In that case user wouldn't be able to pass certain props to his own component

HenriBeck: we could pass the context as an own prop, something like `jssContext` and not spread it

kof: something is wrong with the way we use context consumer, we should not have that namespace problem

kof: we can do something like __jssContext__, but it seems weird

HenriBeck: I think we can just use `jssContext`

kof: Ah actually if you pass it after user props, there is no danger user can override it, so in case yes

packages/react-jss/src/createHoc.js
```diff
@@ -274,18 +269,21 @@ export default function createHOC<
...
-
-    // eslint-disable-next-line no-inner-declarations
-    function ContextSubscribers(props) {
-      return <ThemeConsumer>{theme => <Jss theme={theme} {...props} />}</ThemeConsumer>
-    }
-
-    ContextSubscribers.InnerComponent = InnerComponent
-
-    return ContextSubscribers
+  return function ContextSubscriber(props) {
```
kof: This is basically a component user will see in the devtools, right?

HenriBeck: Yes.

kof: Can we avoid having this additional component in each HOC?

HenriBeck: Nope, not possible.

kof: hmm, what about having the Consumer inside of Jss Component?

kof: You will still have to render Consumer, but you won't have ContextSubscriber component

HenriBeck: Won't work because we need the context in the lifecycle methods

packages/react-jss/src/injectSheet.js
```diff
@@ -29,6 +29,9 @@ export default function injectSheet(styles: StylesOrCreator, options?: Options =
   return (InnerComponent: ComponentType<InnerProps> = NoRenderer) => {
     const Jss = createHoc(styles, InnerComponent, {...options, index})
 
+    // $FlowFixMe
+    Jss.InnerComponent = InnerComponent
```
kof: Mb if we place it as a static property in Jss component it won't complain?

packages/react-jss/src/createHoc.js
kof: Ok, lets rename to JssContextSubscriber

kof: or JssContextConsumer

packages/react-jss/src/createHoc.js
HenriBeck: Done.

packages/react-jss/src/JssContext.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import createReactContext, {type Context} from 'create-react-context'
+import type {Context as JssContextValue} from './types'
+
+const JssContext: Context<JssContextValue> = createReactContext({
+  sheetOptions: {},
```
kof: didn't we remove those?

HenriBeck: No only the prop on the `JssProvider`

HenriBeck: We still create it for the context

packages/react-jss/src/injectSheet.js
HenriBeck: That won't fix it, a React Component type simply doesn't have a InnerComponent prop

PR #928, opened 2018-12-06
[jss] Add a benchmark for creating jss instances

HenriBeck:
Add a benchmark for creating new jss instances. Once without any plugins and once with the preset.

Results:
Chrome 70.0.3538 (Mac OS X 10.14.2)  create a new instance with and without preset: without preset at 205906 ops/sec
Chrome 70.0.3538 (Mac OS X 10.14.2)  create a new instance with and without preset: with preset at 51269 ops/sec

### DIFF
--- a/packages/jss/benchmark/tests/create-instance.js
+++ b/packages/jss/benchmark/tests/create-instance.js
@@ -1,0 +1,18 @@
+import {create} from 'jss'
+import preset from 'jss-preset-default'
+
+// Avoid memory leak with registry.
+const options = {virtual: true}
+
+suite('create a new instance with and without preset', () => {
+  benchmark('without preset', () => {
+    create(options)
+  })
+
+  benchmark('with preset', () => {
+    create({
+      ...options,
+      ...preset()
+    })
+  })
+})
--- a/packages/jss/benchmark/tests/create-instance.js
+++ b/packages/jss/benchmark/tests/create-instance.js
@@ -4,7 +4,7 @@ import preset from 'jss-preset-default'
 // Avoid memory leak with registry.
 const options = {virtual: true}
 
-suite('create a new instance with and without preset', () => {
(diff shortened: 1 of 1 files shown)


kof:
so the difference is 4x

packages/jss/benchmark/tests/create-instance.js
```diff
@@ -0,0 +1,18 @@
...
+const options = {virtual: true}
+
+suite('create a new instance', () => {
+  benchmark('without preset', () => {
+    create(options)
+  })
+
+  benchmark('with preset', () => {
+    create({
+      ...options,
```
kof: mb try to create an options before the bench?

HenriBeck: Moving the options outside the bench only gives you about 2000 ops/sec more.

HenriBeck:
We can still create 51 jss instance per ms which is quite enough in my opinion for creating a new jss instance on every SSR render.

HenriBeck:
We can still create 51 jss instance per ms which is quite enough in my opinion for creating a new jss instance on every SSR render.

kof:
what we wanted to know actually is this: does jss instance creation if done often create relative much overhead compared to having just one instance. Right now the number is 4x, though its not very representative, sinve we actually need to know this number relative to styles creation.

I would additionally add a bootstrap toString case and compare it with and without new jss instance

HenriBeck:
No, simply creating a new jss instance without plugins is 4x faster.

The `toString` benchmark shouldn't be affected by creating a new jss instance on every request.

kof:
I didn't mean it should be affected, I ment I would want to see an impact on perf when creating a new instance in comparison to something that is generating styles, just to see the relative value additionally, toString is pretty much what happens on the server.

PR #934, opened 2018-12-11
[react-jss] Remove inject option

HenriBeck:
This removes the `inject` option for the HOC and replaces it with only one option for injecting the theme.

Injecting the `sheet` shouldn't be necessary and unstable because it would be sometimes the dynamic sheet if it exists or the static sheet.
And there is no real benefit to not injecting the `classes` prop.

Injecting the theme is now handled by the option `injectTheme` which is a boolean.


packages/react-jss/src/createHoc.js
```diff
@@ -245,31 +228,31 @@ export default function createHOC<
...
         theme,
         jssContext,
         // $FlowFixMe: Flow complains for no reason...
         ...props
       } = this.props
-      const sheet = dynamicSheet || staticSheet
 
-      if (injectMap.sheet && !props.sheet && sheet) props.sheet = sheet
-      if (injectMap.theme) props.theme = theme
+      if (injectTheme && props.theme) props.theme = theme
```
kof: I am not sure we even need injectTheme prop

HenriBeck: Always inject theme?

HenriBeck: Actually always injecting theme shouldn't be a big problem, the theme shouldn't change often that it would cause a lot of rerenders

kof: I would say yes, btw that line doesn't make sense, bc you have no theme in props, props is the rest ... so we seem to lack a test for this

HenriBeck: We have a test for that, the tests are failing anyway right now

packages/react-jss/src/createHoc.js
```diff
@@ -76,8 +60,7 @@ export default function createHOC<
   options: Options
 ): ComponentType<OuterPropsType> {
   const isThemingEnabled = typeof stylesOrCreator === 'function'
```
kof: I think we can also get rid of isThemingEnabled, because if a theme is on the context, its weird it would be implicitely depending on styles being a function

packages/react-jss/src/index.d.ts
```diff
@@ -25,7 +25,7 @@ export const JssProvider: React.ComponentType<{
 type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
 type Options = {
   index?: number
-  inject?: Array<'classes' | 'theme' | 'sheet'>
+  injectTheme?: boolean
```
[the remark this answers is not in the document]
HenriBeck: No, someone might still want to subscribe to the theme even though he has a styles object

packages/react-jss/src/createHoc.js
HenriBeck: No, we still need to do some checks based on `isThemingEnabled`

kof: Lets revisit them, I think we can get rid of it.

HenriBeck: We still need this to decide which theme we are using. 

When someone passed a function as the styles, we use the theme prop. 

Because the user still might pass `injectTheme: true` which would subscribe to the theme, even though the styles are not theme dependent, we would create additional stylesheets.

kof: I think isThemingEnabled logic needs to be different,  e.g.  `shouldSubscribeTheme = options.injectTheme || stylesIsfunction`  and rendering of styles just based on stylesIsFunction

PR #937, opened 2018-12-13
Use document.querySelector

oliviertassinari:
Pros:
- smaller bundle size
- better performance? https://twitter.com/matthewcp/status/1073302509186363394

Cons:
- ?

*oops, I'm sorry for the new branch. I should use my fork!*

### DIFF
--- a/packages/jss/src/renderers/DomRenderer.js
+++ b/packages/jss/src/renderers/DomRenderer.js
@@ -120,9 +120,7 @@ function setSelector(cssRule: CSSStyleRule, selectorText: string): boolean {
 /**
  * Gets the `head` element upon the first call and caches it.
  */
-const getHead = memoize(
-  (): HTMLElement => document.head || document.getElementsByTagName('head')[0]
-)
+const getHead = memoize((): HTMLElement => document.querySelector('head))
 
 /**
  * Find attached sheet with an index higher than the passed one.


packages/jss/src/renderers/DomRenderer.js
```diff
@@ -120,9 +120,7 @@ function setSelector(cssRule: CSSStyleRule, selectorText: string): boolean {
 /**
  * Gets the `head` element upon the first call and caches it.
  */
-const getHead = memoize(
-  (): HTMLElement => document.head || document.getElementsByTagName('head')[0]
-)
+const getHead = memoize((): HTMLElement => document.querySelector('head'))
```
oliviertassinari: Regarding the memoize function. We might be able to use it with some of the jss plugins.

kof: I am not sure which browsers needed the geteelementsByTagName, I can run your change  to see if everthing we support works

kof: what do you mean by using it in the plugins?

kof: btw. its not actually faster https://twitter.com/matthewcp/status/1073306680300105734 and it really doesn't matter, in our case I would even go for slower document.head if it works in all browsers we support because this micro performance doesn't matter if its done once and cached

kof: lets focus on important stuff, releasing v10, fixing bugs, refactoring the docs, especially docs, they suck.

kof: and most important static extraction

oliviertassinari: Yes, it's definitely not faster. I was just curious why it wasn't using `document.querySeletor()`.

kof: according to https://caniuse.com/#search=queryselector it should work everrywhere we need

PR #941, opened 2018-12-21
Removed -syntax from plugin package names.

kof:
A merge of unpkg pr by @appsforartists + adding react-jss to jss-starter-kit + renaming all packages to be without -syntax suffix/prefix.


kof:
Something is wrong on the CI, I don't see those errors locally

appsforartists:
Does it make sense to remove `value-` too, if you're removing `-syntax`?

kof:
You mean in jss-plugin-rule-value-function and jss-plugin-rule-value-observables? it was ment like css rule and css value function/observable. Better ideas?

HenriBeck:
I don't think we should drop the `rule-value`

kof:
I just have no other idea how to call them, just jss-plugin-functions & jss-plugin-observables?

appsforartists:
You know your naming conventions better than I do.  😃 

I just saw you dropped the category for one set, and didn't know if you wanted to drop it from the others.

kof:
Ah no its not a category, its literally what they do ...

kof:
We started to prefix with -syntax like babel but then realized some have syntax + additional effects, so it wasn't consistent.

HenriBeck:
I think we need to split this PR into three PRs:
- Rename packages (this has nothing to do with unpkg)
- Add unpkg support for the currently existing packages
- Add the jss-starter-kit package

This will make it easier to find out what is wrong with the tests

appsforartists:
I believe the tests were passing here, modulo the missing `react-jss` bundle:

https://github.com/cssinjs/jss/pull/936

We could rebase in the `react-jss` fix, merge it, and then have this one just be the renaming of the packages.

kof:
react-jss is added to starter kit and merged, will send a separate PR for renaming

Issue #800, opened 2018-08-03
[react-jss] Readd forwardRef support


HenriBeck:
_From @kof on June 12, 2018 8:52_

It was removed due to issues with redux-connect and the fact that it needs a major release, see #255 

Now we need to understand how redux-connect and similar can be supported.

_Copied from original issue: cssinjs/react-jss#256_

HenriBeck:
I don't think we can support react-redux in any way, rather we need to wait for react-redux to not check against a function and rather against a valid react element described [here](https://github.com/reduxjs/react-redux/issues/914#issuecomment-377406014)

HenriBeck:
_From @kof on June 12, 2018 9:6_

Yeah, but how can we release react-jss with forwardRef support without making it incompatible with so  many react-redux versions?

HenriBeck:
I don't think we can.

HenriBeck:
_From @kof on June 12, 2018 10:13_

Lets release this change in a major version as soon as https://github.com/reduxjs/react-redux/issues/914 is fixed

acatcalledfrank:
Looks like [react-redux/issues/914](https://github.com/reduxjs/react-redux/issues/914) has been resolved so hopefully that's unblocked this?

HenriBeck:
This was implemented in #943

PR #943, opened 2018-12-22
[react-jss] Add forward ref support

HenriBeck:
Because we now only support react 16.3 or higher, we can safely add forward ref support again as also react redux has fixed their issue with the API.

__Corresponding issue:__ #800

### DIFF
--- a/packages/react-jss/src/createHoc.js
+++ b/packages/react-jss/src/createHoc.js
@@ -247,21 +247,24 @@ export default function createHOC<
     }
   }
 
-  return function JssContextSubscriber(props) {
-    return (
-      <JssContext.Consumer>
-        {context => {
-          if (isThemingEnabled || injectTheme) {
-            return (
-              <ThemeConsumer>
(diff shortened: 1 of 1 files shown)


packages/react-jss/src/createHoc.js
```diff
@@ -247,21 +247,24 @@ export default function createHOC<
...
-    )
-  }
+  // $FlowFixMe: Sadly there is no support for forwardRef yet
+  const JssContextSubscriber = React.forwardRef((props, ref) => (
+    <JssContext.Consumer>
+      {context => {
+        if (isThemingEnabled || injectTheme) {
+          return (
+            <ThemeConsumer>
+              {theme => <Jss innerRef={ref} theme={theme} {...props} jssContext={context} />}
```
kof: do we still need innerRef here? or can it be dirrectly ref?

HenriBeck: We still need `innerRef` because we don't want the user to have the instance of the HOC.

PR #945, opened 2018-12-23
global.CSS might not exist in the browser and def. doesn't in node

kof:


packages/jss/src/utils/escape.js
```diff
@@ -2,14 +2,12 @@ const env = process.env.NODE_ENV
 
 const escapeRegex = /([[\].#*$><+~=|^:(),"'`])/g
 
+const nativeEscape = typeof CSS === 'undefined' ? CSS.escape : undefined
```
HenriBeck: Should be `typeof CSS !== 'undefined'`

HenriBeck: Also could be `const escape = typeof CSS !== 'undefined' ? CSS.escape : (str) => str.replace(escapeRegex, '\\$1')`

That way we would just call `escape(str)` then.

kof: already fixed the first one, we need an additional wrapper around this escape function to avoid escaping in production

packages/jss/src/utils/escape.js
```diff
@@ -2,14 +2,12 @@ const env = process.env.NODE_ENV
 
 const escapeRegex = /([[\].#*$><+~=|^:(),"'`])/g
 
+const nativeEscape = typeof CSS !== 'undefined' && CSS.escape
+
 export default str => {
   // We don't need to escape it in production, because we are not using user's
   // input for selectors, we are generating a valid selector.
   if (env === 'production') return str
```
HenriBeck: Quick question about that: What happens if the user has a custom generate-id and passed some non-valid css selector.

packages/jss/src/utils/escape.js
kof: actually I need to shift as much as possible inside of that wrapping function, because it will be removed from the bundle

HenriBeck: yeah, but instead of doing the check if nativeEscape exists we would just call a function then

packages/jss/src/utils/escape.js
kof: in that case its up to the user, because we want to have max perf, escaping has a cost

packages/jss/src/utils/escape.js
HenriBeck: Why will it be removed from the bundle?

kof: because of the if production check

packages/jss/src/utils/escape.js
HenriBeck: We should maybe note that in the docs about a custom generate-id if it's not already there.

kof: yeah, the resulting id should be a selector conform string

packages/jss/src/utils/escape.js
HenriBeck:
```js
const escape = typeof CSS !== 'undefined' ? CSS.escape : str => str.replace(escapeRegex, '\\$1')

export default str => {
  // We don't need to escape it in production, because we are not using user's
  // input for selectors, we are generating a valid selector.
  if (env === 'production') return str
  
  return escape(str)
}
```

kof:
updated again, @HenriBeck check . out min bundle, the entire fn body is gone

HenriBeck:
Do we actually need to inline all of the code to get it removed? The function should be in production `str => str`. Then the variables outside aren't used anymore so they should also be stripped away correct?

kof:
not sure, will check

kof:
if its on top: `"undefined"!=typeof CSS&&CSS.escape;var k=function(e){return e}`

HenriBeck:
weird, alright.

kof:
it probably has to stay because minifier doesn't know if property access can be a getter, so that code might actually do something

kof:
I don't get it, CI fails on bundle size checks, I cloned the repo fresh and everything passes locally, I am gonna try merging it and hope its some weird CI problem that won't be there on master

kof:
nope, still fails, I need it on master though for @oliviertassinari , lets figure why it fails on master retrospectively

packages/jss/src/utils/escape.js
```diff
@@ -1,15 +1,10 @@
-const env = process.env.NODE_ENV
-
-const escapeRegex = /([[\].#*$><+~=|^:(),"'`])/g
-
 export default str => {
   // We don't need to escape it in production, because we are not using user's
```
oliviertassinari: On Material-UI, we still escape in production as people can use a global CSS approach, where they provide a name to each style sheet.

packages/jss/src/utils/escape.js
```diff
@@ -1,15 +1,10 @@
...
   // input for selectors, we are generating a valid selector.
-  if (env === 'production') return str
+  if (process.env.NODE_ENV === 'production') return str
 
-  if (CSS && CSS.escape) {
-    return CSS.escape(str)
-  }
+  const escapeRegex = /([[\].#*$><+~=|^:(),"'`])/g
```
oliviertassinari: Should the regexp be moved outside of the function to provide better pers in dev?

oliviertassinari: Also, we uses `/([[\].#*$><+~=|^:(),"'`\s])/g;` on Material-UI side to solve some other edge cases.

HenriBeck:
I will check it this evening or tomorrow

packages/jss/src/utils/escape.js
oliviertassinari: It's what we use at https://www.onepixel.com/en/ to get some decent caching performance.

packages/jss-preset-default/.size-snapshot.json
```diff
@@ -1,13 +1,13 @@
 {
   "dist/jss-preset-default.js": {
-    "bundled": 67203,
-    "minified": 23226,
-    "gzipped": 6969
+    "bundled": 30500,
```
oliviertassinari: I would encourage the usage if [size-limit](https://github.com/ai/size-limit) over having to commit the new values every time a pull request is done. it's wasteful.

packages/jss/src/utils/escape.js
kof: If they do this, they also use valid strings for class names, why do you still need escaping there?

packages/jss/src/utils/escape.js
kof: I usually do that, but actually engine already does optimizations there, its not really necessarily and it won't be removed from the bundle afaik, if it does get removed we can move it to the top

kof: will update the regex with yours, it fixes the issue with spaces we already got reported and backticks

packages/jss-preset-default/.size-snapshot.json
kof: size limit has also a fixed limit, which you have to fix if you increase, if you make it too flexible - than you don't control the size. I think the point of it is to make you aware of every size change on   for every change

kof: also we have different numbers here, which might behave differently, its interesting to see how minified/gziped size changes compared to raw, those numbers are not linear

packages/jss/src/utils/escape.js
kof: oh just realized you don't have the backtick in the regex, you should add it too

kof:
fixed on master, also moved the regex to the top and added whitespace handling, regex is getting removed in production, since it makes no side effects

packages/jss/src/utils/escape.js
kof: yes, react-jss does too, exactly the same story

packages/jss/src/utils/escape.js
kof: just looked again yes, I already added it too!

kof: our regexes are unified now, does mui has tests for all cases that regex covers? If yes I would copy them to make sure we don't break them by accident

packages/jss-preset-default/.size-snapshot.json
kof: well, they only need to run yarn build, thats it, if not, we can as well do it, its not too bad so far, but if it gets annoying we can add a git hook cc @TrySound

oliviertassinari:
🎉

Issue #933, opened 2018-12-10
[react-jss] Merging classes instead of an override and optionally override


kof:
People ask for a way to merge the classes when using the HOC interface instead of overriding them. If thats what everyone wants to do, we could implement it by default.

https://stackoverflow.com/questions/53674340/merge-classes-with-jss-instead-of-overriding/53674874

## Optional override

We could additionally support an override like this: 

```js
classes = {
   a: 'class-a', // extends
   b: ['class-b' , true], // overrides
   c: ['class-c', 'class-d'  , true] // overrides and bonus - one can use array of classes
}
```

pswg:
Just brainstorming a little. If `classes` could accept a function, you could use a syntax like this:

```
import _mapKeys from 'lodash/mapKeys';

const mergeClasses =
  (extends, overrides) =>
    (baseClasses) =>
      _mapKeys((cls, key) =>
        key in overrides ? overrides[key] : key in extends ? cls + ' ' + extends[key] : cls);

...

classes = {
  mergeClasses({
    a: 'class-a'
  }, {
    b: 'class-b',
    c: 'class-c class-d'
  })
}
```

Not sure if this makes anything easier or much more difficult, though.

kof:
Yes, but would you want to do it as a consumer every time?

pswg:
Probably not, just proposing an alternate syntax. If it *were* implemented this way, I would propose that a comparable `mergeClasses` function be added to the `react-jss` library so the only extra burden to users would be an import then calling it in `classes`.

oliviertassinari:
This sounds like a good tradeoff, for reference, the `mergeClasses` method on Material-UI side:
https://github.com/mui-org/material-ui/blob/5be50c86e9ba5ca08659e4bef14aea2b8bc2d902/packages/material-ui-styles/src/mergeClasses.js

kof:
The question is what is the mainstream behaviour. If everyone wants mergeClasses behaviour often, asking to provide a function and then import mergeClasses and merge it in consumer component is a lot of work.

HenriBeck:
If we want to support both merging and overriding, I think the syntax from @kof is the cleanest and easiest.

But I'm not quite sure about allowing multiple classes inside the array.

HenriBeck:
Implemented in #946

PR #946, opened 2018-12-23
[react-jss] Merge classes instead of overwriting

HenriBeck:
This adds support for merging classes instead of overriding them.
Currently, it is not supported to override classes by the syntax in #993.
We would need to do more work to not compute unless there is a change.

__Corresponding issue (if exists):__ #933


packages/react-jss/src/createHoc.js
```diff
@@ -3,9 +3,10 @@
 import React, {Component, type ComponentType} from 'react'
 import PropTypes from 'prop-types'
 import {ThemeContext} from 'theming'
-import {getDynamicStyles, SheetsManager, type StyleSheet, type Classes} from 'jss'
+import {getDynamicStyles, SheetsManager, type StyleSheet} from 'jss'
+import memoize from 'memoize-one'
```
kof: do we need a lib for this? I have a feeling that function is 2-3LOC and easier to inline than to have a dependency

packages/react-jss/src/createHoc.js
```diff
@@ -223,22 +214,22 @@ export default function createHOC<
       return {
         staticSheet,
         dynamicSheet,
-        classes: this.computeClasses(staticSheet, dynamicSheet)
+        classes: mergeClasses(staticSheet.classes, dynamicSheet ? dynamicSheet.classes : {})
```
kof: could this benefit from memoized mergeClassesProp? its also called on each update

packages/react-jss/src/merge-classes.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import type {Classes} from 'jss'
+
+/**
+ * This merges to classes object.
+ * When a property exists in both classes, it will not overwrite but merge the two classes.
+ *
+ * @param {Object} baseClasses - The base classes.
```
kof: we should drop jsdoc anotations since we use flow already, maintaining both is bad, because it will get out of sync without validation

packages/react-jss/src/createHoc.js
HenriBeck: No this is only called when creating a new stylesheet, so only on theme changes and on mount when no sheet exists

kof:
You know what? Lets do as you wanted - only shallow merge the defaultProps.classes. If component author really wants a deep merge, they can still write a utility function and do it, like we do in react-jss hoc. If we don't deeply merge it, they have at least a choice to decide how they want it.

Deeply merging defaultProps.classes will have as a result those classes always added to the "class" attribute, which will increase SSR payload. Evtl. component author wants to control that behaviour.

packages/react-jss/src/memoize-one.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+function memoize<Arg, Return>(fn: (arg: Arg) => Return) {
```
kof: btw why do you like named functions so much? by the latest spec a function expression assigned to a variable has a function name as well, derived from the variable name. So there is no problem with debugging.

I don't like named functions because of the implicit hoisting, its a generally leaky concept.

kof: its like var and even worse, no one is using var any more and named functions are even worse, because they can be called before they are defined and once people start relying on this, code gets messy.

HenriBeck: Personally, it makes it easier to differentiate between variables and I'm always using eslint or tslint where I disallow to not use something before it's not defined.

Issue #913, opened 2018-11-05
Replace warning with smaller version


TrySound:
`warning` package is published as a strong mirror of fbjs thing. Maintainer even do not want to support esm. This requires to transpile more packages from commonjs when bundling app with rollup.

There is a smaller alternative https://github.com/alexreardon/tiny-warning. It doesn't allow passing arguments and interpolation should be used instead. It's distributed as esm and has flow types out of the box.

`react-router` already migrated to it.
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/package.json#L50-L51

kof:
We still do support IE 9 same as react does, IE9 doesn't implement console unless devtools are open and this package has no fallback. So to do this we need to officially deprecate IE9 or add a fallback to tiny-warning.

I am curious if react-router maintainers know it.

HenriBeck:
Do we need to support ie9? IE 9 has 0.13% usage globally

HenriBeck:
I think we could configure our babel setup to transpile for browsers who have a usage bigger than 0.5%

kof:
I personally don't, but I am not sure what the majority does.

TrySound:
Hey, guys. ie9 is dead browser. I think major release is a good reason to stop supporting it.
https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support

kof:
That announcement from microsoft just means they don't ship any updates. Developers community will support or not support a browser based on usage statistics and the question is where to make a cut. 

I am happy to drop IE9, not sure regarding 10.

TrySound:
tiny-warning now has IE9 support

kof:
cool, thanks for the heads up

kof:
I wonder how it is generating an error, because the error is thrown inside of a catch block, which does nothing ... I guess I am missing something

TrySound:
Here's the reason of swallowed error. But it's dropped for production, so don't worry.
https://github.com/facebook/react/issues/4216

kof:
> Chrome allows you to break on caught errors. That's why that try/catch is there. So that you can optionally toggle the breakpoint on and off.

That was the thing I missed

HenriBeck:
Let’s migrate then?

HenriBeck:
PR is merged

oliviertassinari:
The tiny wording is misleading 😆

[...]

PR #953, opened 2018-12-27
[core] Use tiny warning

HenriBeck:
Use `tiny-warning` instead of `warning`.

__Corresponding issue:__ #913


packages/jss/src/renderers/DomRenderer.js
```diff
@@ -214,7 +212,7 @@ function findPrevNode(options: PriorityOptions): PrevNode | false {
 
     // If user specifies an insertion point and it can't be found in the document -
     // bad specificity issues may appear.
-    warning(insertionPoint === 'jss', '[JSS] Insertion point "%s" not found.', insertionPoint)
+    warning(false, `[JSS] Insertion point "${insertionPoint}" not found.`)
```
HenriBeck: Not sure why we are checking for `jss` here, when the user wants a custom insertion point, it should exist.

HenriBeck:
I fixed the tests by removing the rewires for the warning libary and replaced it with spying on the console.

HenriBeck:
We could also drop the rewire for the env stuff if we didn't cache them.

packages/jss-plugin-template/src/parse.js
```diff
@@ -18,7 +18,7 @@ export default (cssText: string) => {
     if (!decl) continue
     const colonIndex = decl.indexOf(':')
     if (colonIndex === -1) {
-      warning(false, 'Malformed CSS string "%s"', decl)
```
kof: we should use JSS prefix everywhere

HenriBeck: Good catch 👍

packages/jss/src/utils/createGenerateId.js
```diff
@@ -1,27 +1,26 @@
...
-const env = process.env.NODE_ENV
-
 /**
  * Returns a function which generates unique class names based on counters.
  * When new generator function is created, rule counter is reseted.
  * We need to reset the rule counter for SSR for each request.
  */
 export default (): GenerateId => {
   let ruleCounter = 0
+  const env = process.env.NODE_ENV
```
HenriBeck: We cache it here so that we can still manipulate it in tests and don't have a negative impact on performance on when doing SSR.

packages/jss/src/utils/escape.js
```diff
@@ -1,10 +1,9 @@
-const env = process.env.NODE_ENV
 const escapeRegex = /([[\].#*$><+~=|^:(),"'`\s])/g
 
 export default str => {
   // We don't need to escape it in production, because we are not using user's
   // input for selectors, we are generating a valid selector.
-  if (env === 'production') return str
+  if (process.env.NODE_ENV === 'production') return str
```
HenriBeck: We might need to investigate if this has performance problems on the server because accessing environment variables is expensive.

packages/jss/tests/unit/escape.js
```diff
@@ -1,35 +1,16 @@
...
 
     it('should not escape in production', () => {
-      escape.__Rewire__('env', 'production')
+      process.env.NODE_ENV = 'production'
       expect(escape('test()')).to.be('test()')
-      escape.__ResetDependency__('env')
-    })
-  })
-
-  describe('without CSS.escape', () => {
```
HenriBeck: I don't think we need to test this here. We just need to make sure that it works in our supported browsers.

packages/react-jss/src/createHoc.js
```diff
@@ -109,6 +107,8 @@ export default function createHOC<
     constructor(props: OuterPropsType, context: Context) {
       super(props, context)
 
+      this.context = context
```
HenriBeck: @kof This fixes the weird issue with `react-jss` in the alpha.

packages/jss/src/utils/escape.js
kof: its not expensive, it will be replaced by a string and in prod after uglify removed completely

packages/jss/tests/unit/escape.js
kof: I think this test is there to ensure the regex fallback works.

kof: since in most env it will use CSS.escape

packages/react-jss/src/createHoc.js
kof: strange, parent class should have done this

kof: oh mb we are using context before super does the job

packages/jss/src/renderers/DomRenderer.js
kof: I think it was there because we used to have a default insertion point 'jss', which should not warn if its not found.

packages/jss/src/utils/escape.js
kof: Yeah, I don't think its expensive, its a regular property.

PR #958, opened 2018-12-30
Docs

kof:


docs/faq.md
```diff
@@ -0,0 +1,39 @@
+# FAQ
```
kof: What else can we add here?

docs/intro.md
```diff
@@ -16,6 +16,11 @@ If you’re interested in playing around with JSS, you can use an online code pl
 
 https://cssinjs.org/repl
 
+## Third-party API adapters
```
HenriBeck: We should also add styled-jss here

kof: true

PR #961, opened 2018-12-31
Create .npmignore

AleshaOleg:


TrySound:
This is monorepo. npmignore in the root does not do anything. There are files fields which fix this for v10.
https://github.com/cssinjs/jss/blob/master/packages/jss/package.json#L22

Issue #964, opened 2018-12-31
typescript is not tested


rosskevin:
__Expected behavior:__
Any typescript definitions exported from this library should be tested, otherwise it can cause real pain and doubt for ts users.

__Describe the bug:__
https://github.com/cssinjs/jss/pull/889#issuecomment-447220080 was integrated without tests.  It did not start with @types/jss package and we have no guarantee compatibility.

__Versions (please complete the following information):__
 - jss: 9 to 10.alpha.3

__Context__
Changelog from 9->10 should address user's removal of @types/jss as a requirement otherwise conflicts will arise.  As-is, I am seeing app errors downstream of material-ui 3.8.1 because of a mix of type use between 9.x and 10.x.  It is not easy or apparent there, but is within our app. Investigating in material-ui, different packages require different versions of jss, but ultimately the root of the yarn workspace has 10.alpha.3 **and** @types/jss.  It just so happens that the tests/usage there passes, but does not in the downstream app.

For samples of typescript testing, check out material-ui https://github.com/mui-org/material-ui and usage tests such as https://github.com/mui-org/material-ui/tree/master/packages/material-ui/test/typescript

/cc @appsforartists @HenriBeck

rosskevin:
I'm not sure, I don't have an answer yet - I think olivier is probably celebrating new years.  Probably unintended but I did find it in the monorepo root "resolutions".  I think it was intended to be performance testing for another alpha mui package and...unintentionally...since it was in the root resolutions runtime types (which were scoped to 9.x and @types/jss) ended up getting updated to 10.alpha instead of staying on @types/jss.

Actual details here: https://github.com/mui-org/material-ui/issues/14040#issuecomment-450684391

Intentions yet to be determined.

rosskevin:
@HenriBeck it looks like alpha dependency was unintentional for `@material-ui/core`, and intended only for new public package (also alpha) of `@material-ui/styles`.  Bug/error in package.json(s).

HenriBeck:
Breaking changes are anyways required as we changed the name of `createGenerateClassname` to `createGenerateId`.

I think we can update the current definitions to include such features. I was hesitant to just copy them over as the `Styles` type might be wrong/incomplete.

glenveegee:
FWIW: following a recent upgrade to the bleeding edge I'm now getting this error and I suspect it is directly related to some of the points raised in the discussion above 
[...]

Issue #806, opened 2018-08-03
styles typing (auto-completion, validation)


HenriBeck:
_From @caub on March 29, 2018 9:25_

I think many of us would like to have syntax highlighting, validation, .. for the styles

I'll talk about VSCode, sorry about other editors, styled-components have https://github.com/styled-components/vscode-styled-components

It should be possible to do it, but more using React definitions:
![2018-03-29-112240_1305x489_scrot](https://user-images.githubusercontent.com/3134574/38081267-e469068c-3343-11e8-9297-1606459c512d.png)

And trigger it for a 
```js
const styles = {
  foo: {
    alignItems: '...'
  }
};
```

@visioncan did something interesting: https://github.com/visioncan/vscode-jss-snippets, but I think I'd prefer that it behaves like React inline styles

_Copied from original issue: cssinjs/react-jss#218_

HenriBeck:
_From @kof on March 29, 2018 9:46_

Also having flow and/or typescript typings for styles would allow vscode and other editors to show the hints https://github.com/cssinjs/jss/issues/361 https://github.com/cssinjs/jss/issues/375

Types could be generated out of [mdn database](https://www.npmjs.com/package/mdn-data), we started a repo but 
 I still had to time  https://github.com/cssinjs/style-type-generator

HenriBeck:
I would also prefer adding types rather than specific plugins as types can work in all editor which support flow/typescript.

Also tracking down which object is a Jss stylesheet might be hard, especially when they are in different files.

HenriBeck:
_From @caub on April 19, 2018 17:57_

The ones from [Glamorous](https://github.com/paypal/glamorous/tree/master/typings) might be really close to what is needed ([talk](https://www.youtube.com/watch?v=r5UtdJNxN94))

HenriBeck:
_From @jeffal on April 23, 2018 21:37_

Would this idea include class name validation? For example, a no unused class name lint rule.

HenriBeck:
_From @kof on April 24, 2018 11:6_

@jeffal I think its  a separate issue, created one here https://github.com/cssinjs/jss/issues/705

HenriBeck:
I have typed my styles like this, though that doesn't save you from unused rules.

Example:
```js
const styles = {
  root: {
    display: 'flex',
  },
};

type Props = { classes: { [keys: $Keys<typeof styles>]: string } };

// or with a little helper type:
[...]

PR #973, opened 2019-01-05
[jss] Update TS types

HenriBeck:
This aligns the types more with `@types/jss` and adds a few features which were missing before.

__Corresponding issue (if exists):__ #964


kof:
I can't really review it, lets wait for @appsforartists

appsforartists:
I'm on vacation until February, and probably won't have much time for reviewing in the mean time.

"This aligns the types more with @types/jss and adds a few features which were missing before."

Can you split this into two PRs - one that imports `@types/jss` and one that makes whatever changes you are proposing?

kof:
Lets add changelog and move forward with this, we can always fix things retrospectively since we are in alpha

appsforartists:
I'm back from vacation now.

I still don't understand why this isn't just an importing of `@types/jss`.  A bunch of work has gone into that package, and 600k downloads per week have tested it.

Can we please import that package, and then you can make whatever changes you see fit?  There's no good reason to start from scratch.

HenriBeck:
I have a commit now which replaces the styles, and then I made the required changes on top of that.
They were mostly about incomplete/outdated typings.

kof:
@HenriBeck did you port all types from the `@types/jss` package?

HenriBeck:
Yeah, most of them.

packages/jss/src/index.d.ts
```diff
@@ -1,35 +1,156 @@
...
-export interface Rule {
-  className: string;
-  selector: string;
-  applyTo(element: HTMLElement): void;
-  prop(key: string): string;
-  prop(key: string, value: any): this;
-  toJSON(): string;
+export type Rule = BaseRule | ContainerRule
+
+export declare class Renderer {
```
appsforartists: You've added a whole bunch of types here that I don't see exported:

https://github.com/cssinjs/jss/blob/master/packages/jss/src/index.js

Since JSS doesn't use TypeScript internally, this typings file should only describe the public API.  Please remove the types that aren't used publicly.

For objects that are used in the public API but not exported from `index.js`, prefer `interface` over `class`.

packages/jss/src/index.d.ts
```diff
@@ -1,35 +1,156 @@
...
+declare class SheetsManager {
+  readonly size: number
+  get(key: object): StyleSheet | null
+  add(key: object, sheet: StyleSheet): void
+  manage(key: object): StyleSheet | null
+  unmanage(key: object): void
 }
 
-export interface StyleSheet<RuleName extends string = any> {
+export class StyleSheet<RuleName extends string = string> {
```
appsforartists: I don't see a publicly exported class named `StyleSheet`, so changing this to `class` is incorrect.

packages/jss/src/index.d.ts
```diff
@@ -1,35 +1,156 @@
...
+  sheet?: StyleSheet
+  index?: number
+  parent?: ContainerRule | StyleSheet
+  classes: Classes
+  jss: Jss
+  generateId: GenerateId
+  Renderer: Renderer
+}
+
+declare class BaseRule {
```
appsforartists: Why is this shape so different from `Rule`?  For instance, `applyTo` and `prop` are missing.

packages/jss/src/index.d.ts
```diff
@@ -1,35 +1,156 @@
...
+
+interface ContainerRule extends BaseRule {
+  rules: RuleList
+}
+
+export interface Plugin {
+  onCreateRule?(name: string, decl: Style, options: RuleOptions): Rule
+  onProcessRule?(rule: Rule, sheet?: StyleSheet): void
+  onProcessStyle?(style: Style, rule: Rule, sheet?: StyleSheet): Style
+  onProcessSheet?(sheet?: StyleSheet): void
```
appsforartists: Do `onProcessSheet` or `onProcessRule` ever get called without `sheet`?

packages/jss/src/index.d.ts
```diff
@@ -67,80 +191,47 @@ export interface StyleSheet<RuleName extends string = any> {
...
+  update(name: string, data: object, options?: UpdateOptions): this
+  update(data: object, options?: UpdateOptions): this
   /**
    * Convert rules to a CSS string.
    */
   toString(options?: ToCssOptions): string;
 }
-export type GenerateClassName<Name extends string = any> = (rule: Rule, sheet?: StyleSheet<Name>) => string;
-
-export interface JSSPlugin {
```
appsforartists: It looks like you've changed both the order and the shape of some of these types.  That makes it really hard to review.

packages/jss/src/index.d.ts
HenriBeck: BaseRule is the most common rule, `prop` and `applyTo` are only defined on a `StyleRule`

packages/jss/src/index.d.ts
HenriBeck: Yeah, because the one was renamed and the other typing was wrong.

packages/jss/src/index.d.ts
HenriBeck: `onProcessSheet` shouldn't, but `onProcessRule` can be called without a sheet, so this is very rare.

packages/jss/src/index.d.ts
HenriBeck: The types are exported. Not sure if a class would be the correct type here as they are only types and can't be initiated.

packages/jss/src/index.d.ts
```diff
@@ -1,35 +1,156 @@
-// Type definitions for jss 9.5
-// Project: https://github.com/cssinjs/jss#readme
-// Definitions by: Brenton Simpson <https://github.com/appsforartists>
-//                 Oleg Slobodskoi <https://github.com/kof>
-//                 Thomas Crockett <https://github.com/pelotom>
-//                 Sebastian Silbermann <https://github.com/eps1lon>
-// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+export type Style = object;
```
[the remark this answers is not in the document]
HenriBeck: Well, the file only guaranteed partial type safety on `Style` and a few things are missing. Also with our `plugin` API currently, it is really hard to type those things. So, for now, I would like to hold off on typing `Style`.

There is an open issue about adding types to `Style`: #806

kof:
@HenriBeck no changelog entry?

PR #980, opened 2019-01-13
Remove warning from production

TrySound:


packages/jss/.size-snapshot.json
```diff
@@ -1,30 +1,30 @@
...
-    "gzipped": 6624
+    "bundled": 54651,
+    "minified": 24090,
+    "gzipped": 6647
   },
   "dist/jss.esm.js": {
-    "bundled": 53619,
-    "minified": 23302,
-    "gzipped": 6536,
+    "bundled": 54135,
```
kof: q: why did esm bundle grow?

TrySound: Because it has now all warnings wrapped with condition
```js
if ("production" !== process.env.NODE_ENV) {
  warning(condition, argument, argument);
}
```

Issue #976, opened 2019-01-08
question: why "types/react" is included in deps?


iamstarkov:
I'm not an experienced Typescript, but I dont think it is a best practice to directly depend on `@types/react`. Correct me if im wrong here and if it should be there please explain why.

Another aspect of this direct dependency is that there are 4 types of react consumers: raw js, compile2js lang, flowtype and typescript consumers. Even If it makes sense for typescript users that react-css directly depends on "types/react", it introduces download, bandwidth and build time tax for everybody. I dont think it is fair

TrySound:
@iamstarkov Tell this to some toxic typescript guys.

TrySound:
They are sure their ecosystem solves their problem and "who cares about others". "TS everywhere!"

HenriBeck:
So what do you suggest? Move it to optional dependencies? As moving it to peer dependencies doesn't make sense IMHO.

TrySound:
Anyway we also include flow types in package so it increases load time a bit too.

HenriBeck:
True, and flow types are the complete source code 🤣

iamstarkov:
shouldn't it be in devDeps?

iamstarkov:
> @iamstarkov Tell this to some toxic typescript guys.

we can't we tell them to heck off

kof:
In general I started using flow for internal purposes, to typecheck the library itself. The fact that it has been exported to the enduser is sort of an accident. I would love to not have to mix those things up with the actual library, just don't know how to do that. Currently if we apply the bus factor to @HenriBeck I will have to remove ts types lol, because I won't be able to make changes.

iamstarkov:
people on twitter tell me it should be in devDeps

iamstarkov:
1: https://twitter.com/stereobooster/status/1082632639549460480
2: https://twitter.com/_nitive/status/1082619613559574528

HenriBeck:
Putting the types into peer deps doesn't make sense because everyone would need to install them, also non-TS users. 

We could add them to dev and optional deps maybe?

iamstarkov:
maybe it is the better option than the current one

kof:
What do other libs do with the same problem?

HenriBeck:
I think the best option is to have in optional dependencies and in dev deps.

PR #990, opened 2019-01-19
Move react-jss types to dev dependencies

HenriBeck:
Installing external types can have a problem when the end user also installs the same types package.

As we shouldn't force someone to install the @types/react package, I just moved to the dev dependencies. I think we can expect someone to have the appropriate package installed when using `react-jss` and using typescript

__Corresponding issue (if exists):__ #976

### DIFF
--- a/packages/react-jss/package.json
+++ b/packages/react-jss/package.json
@@ -42,11 +42,13 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@types/react": "^16.4.18",
     "hoist-non-react-statics": "^3.2.0",
     "jss": "^10.0.0-alpha.8",
     "jss-preset-default": "^10.0.0-alpha.8",
     "prop-types": "^15.6.0",
     "theming": "^3.0.2"
+  },
+  "devDependencies": {
+    "@types/react": "^16.7.20"
   }
 }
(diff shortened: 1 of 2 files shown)


packages/react-jss/package.json
```diff
@@ -42,11 +42,13 @@
...
   "dependencies": {
     "@babel/runtime": "^7.0.0",
-    "@types/react": "^16.4.18",
     "hoist-non-react-statics": "^3.2.0",
     "jss": "^10.0.0-alpha.8",
     "jss-preset-default": "^10.0.0-alpha.8",
     "prop-types": "^15.6.0",
     "theming": "^3.0.2"
+  },
+  "devDependencies": {
```
kof: what about additionally having them in peerDependencies?

HenriBeck: This would require everyone to install them 🤔 
Or if they don't decide on installing them, they will constantly have a warning for a missing peer dependency which isn't nice either

kof: yeah, how do we ensure that userr gets the right version of typings with a version of JSS?

HenriBeck: I don't think we can.

HenriBeck: The things which we are using from `react` are also not to special/new so I don't think we will have a problem with that.

HenriBeck: The definetly typed version also has no dependency on react types package

Issue #984, opened 2019-01-17
Improve publishing experience for JSS maintainers


kof:
## What want the release to do

- [ ] Manually set the version (or mb take it from changelog.md)
- [ ] Build with the new version, commit the size-snapshots with changes (or avoid them somehow), we use the verson in the JS build and this increases bundle sizes
- [ ] Update peer dependencies (#991)
- [ ] Make sure npm tag is set, e.g. an alpha release should not be published with dist tag "latest" otherwise all people get the update.
- [ ] Make it use the changelog.md and add its contents for the release to the release tag on github. This will send out an email to people who subscribed releases and they will get the changees.

HenriBeck:
I think we can use the `postversion` npm hook for building, that hook should be run after bumping the version and before `lerna` commits the changes

kof:
I think this is done now.

PR #991, opened 2019-01-19
[all] Fix peer deps

HenriBeck:
Upgrades all of the peer dependencies to 10.0.0-alpha.8 and fixes the range of some of the peer dependencies

### DIFF
--- a/packages/jss-plugin-cache/package.json
+++ b/packages/jss-plugin-cache/package.json
@@ -41,7 +41,7 @@
     "jss": "^10.0.0-alpha.8"
   },
   "peerDependencies": {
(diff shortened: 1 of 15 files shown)


kof:
damn, we need it to be automated, added a todo in #984

HenriBeck:
We could easily automate them if we would use normal dependencies instead of peer + dev dependencies.

The reason why lerna doesn't automate it is that peer dependencies should be as wide as possibles.

kof:
Then we might need a custom script or something....

HenriBeck:
If we want to operate in "locked" mode, so have the same version across all packages when can use normal dependencies instead of peer + dev dependencies.

PR #992, opened 2019-01-19
Update lerna

HenriBeck:
- Uses `postversion` hook to build packages now
- Renamed `prepare` script back to build again
- Use lerna to execute scripts inside packages (lerna has a nicer output)
- Fix `check:all` command

__Corresponding issue (if exists):__ #984

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "packages/*"
   ],
   "scripts": {
-    "build": "lerna run prepare",
+    "build": "lerna run build",
     "check:all": "yarn check:flow && yarn check:ts && yarn check:snapshots && yarn check:lint",
     "check:flow": "flow check --max-warnings=0",
     "check:ts": "tsc",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -5,13 +5,13 @@
...
   ],
   "scripts": {
-    "build": "yarn workspaces run prepare",
-    "check:all": "yarn check:flow && yarn check:ts && yarn check:snapshots yarn check:lint",
+    "build": "lerna run build",
+    "check:all": "yarn check:flow && yarn check:ts && yarn check:snapshots && yarn check:lint",
     "check:flow": "flow check --max-warnings=0",
     "check:ts": "tsc",
-    "check:snapshots": "yarn workspaces run check-snapshot",
+    "check:snapshots": "lerna run check-snapshot",
```
TrySound: Why not leave `yarn workspaces`? It's a kind of "native" solution.

Issue #932, opened 2018-12-09
New docs


kof:
- [ ] Make sure all docs regarding v10 changes are up-to-date
- [ ] Restructure the site
- [ ] Move all packages docs to /docs
- [ ] Separate syntax part from the plugins into separate md files
- [ ]  /docs/plugins/plugin-x/syntax.md, /docs/plugins/plugin-x/api.md
- [ ] Default Preset should fetch all syntax files and render on one page
- [ ] Each Plugins/plugin page should combine all md files for that plugin in one page

## Structure (inspired by react docs)

Installation
- Getting Started
- Setup
- CDN Links

Main Concepts
- Introduction
- Hello World
- Benefits

Advanced Guides
- Bundling
- SSR
- Performance

API Reference
- Styles Syntax
  - Core
  - Default Preset
- JSS
- React-JSS
- Styled-JSS
- Plugins
  - jss-plugin-x
  - jss-plugin-y

Contributing

FAQ

eps1lon:
Removing the readmes from the packages does have the downside that you won't have any displayed on npmjs.org: https://www.npmjs.com/package/jss-plugin-vendor-prefixer

The individual `package.json` also don't have a repo listed at the moment which again causes trouble with npmjs.org. Packages usually have a link to their repository on their page on npmjs.org.

Just wanted to let you know in case this was not intended.

HenriBeck:
Good catch. I think we should just copy the docs when releasing but ignore them in git.

kof:
Regarding readme.md, lets do the same thing as babel: https://www.npmjs.com/package/@babel/plugin-syntax-export-namespace-from

Regarding docs: its broken right now. My original plan was to publish the 9th version as a subdomain and when dropdown selects v9, switch to that subdomain. I am not sure though its worth the effort and maybe we should just remove v9 from the docs.

HenriBeck:
👍 
I will create a script for creating the readme then

kof:
@HenriBeck its not a full readme, its just title, link to the docs on the website and install steps

PR #996, opened 2019-01-21
[all] Add build readme script

HenriBeck:
This adds a script to create the readmes for the repositories automatically.

Not sure if this should be an own script or be executed when we build all of the packages.

__Corresponding issue (if exists):__ #932


scripts/build-readme.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const contents = `
+# ${packageJSON.name}
+
+> ${packageJSON.description}
+
+See our website [${
+  packageJSON.name
+}](https://cssinjs.org/${getDocsPath()}?v=v10.0.0-alpha.8) for more information.
```
kof: hardcoded version)))

packages/jss-starter-kit/readme.md
```diff
@@ -1,67 +1,19 @@
...
-The `jss-starter-kit` combines all of the other [packages](https://github.com/cssinjs/jss/tree/master/packages) into a single bundle, to make it easy to import directly into a playground like Codepen, or a local HTML file.
+See our website [jss-starter-kit](https://cssinjs.org/jss-starter-kit?v=v10.0.0-alpha.8) for more information.
 
-:warning: This bundle includes every JSS package, unminified. It's great for experimenting, but not suitable for production deployment.
+## Install
 
-## Quick start
+Using npm:
 
-There's an instance of JSS already set up with all of the default plugins, exported as `jss`.
```
kof: Just realized we have no good place where all exports from the starter kit are documented. We should probably create a docs/jss-starter-kit.md and list them all there

scripts/build-readme.js
kof: can we keep them as a separate helper script outside of rollup config to keep the config more readable and declarative?

PR #997, opened 2019-01-23
Fix broken demo links

rollercodester:
The demo links are broken for the extend plugin.

### DIFF
--- a/docs/jss-plugin-extend.md
+++ b/docs/jss-plugin-extend.md
@@ -76,6 +76,5 @@ const styles = {
 
 ### Demo
 
-[Simple demo](http://cssinjs.github.io/examples/plugins/jss-plugin-extend/simple/)
-
-[Multi objects demo](http://cssinjs.github.io/examples/plugins/jss-plugin-extend/multi/)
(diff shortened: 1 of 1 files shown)


HenriBeck:
Thanks 👍

docs/jss-plugin-extend.md
```diff
@@ -76,6 +76,5 @@ const styles = {
 
 ### Demo
 
-[Simple demo](http://cssinjs.github.io/examples/plugins/jss-plugin-extend/simple/)
-
-[Multi objects demo](http://cssinjs.github.io/examples/plugins/jss-plugin-extend/multi/)
+[Simple demo](http://cssinjs.github.io/examples/plugins/jss-extend/simple/)
```
HenriBeck: @kof We could maybe move all of the examples into an own codesandbox? What do you think?

kof: Long term I think we should make them live in the repo and runnable over codesandbox at the same time, from one source, like I did here https://github.com/cssinjs/egghead/tree/master/from-sass-to-cssinjs

PR #998, opened 2019-01-23
[WIP] Improve publishing experience

HenriBeck:
Improving the publishing experience. It should be as easy as running on yarn command without even providing the version.

__Corresponding issue (if exists):__ #984


scripts/update-changelog.js
```diff
@@ -0,0 +1,23 @@
...
+
+const lerna = require('../lerna')
+const {CHANGELOG_FILENAME} = require('./constants')
+
+const content = fs.readFileSync(`../${CHANGELOG_FILENAME}`, 'utf-8')
+
+const lines = content
+  .split('\n')
+  .map(line => {
+    if (line.startsWith(`## ${lerna.version} (unreleased)`)) {
```
HenriBeck: This script is expecting the changelog to be like this for the unreleased version:

```md
## v10.0.0-alpha.9 (unreleased)
```

Is this fine with you @kof or should we use instead of the version `Next` maybe?

kof: I think unreleased is better because, because next is also used in dist-tag and basically means alpha for us

HenriBeck: No, unreleased is the place holder for the date. I was asking about replacing the version with `Next`

HenriBeck: We could have it like this:

```
## Next
```

And then the script could replace it with the version and the date.

Issue #995, opened 2019-01-20
[react-jss] Observable rules/values


HenriBeck:
__Expected behavior:__
Observables should work without explicitly passing `link: true` as options and should be part of the static stylesheet. And no dynamic stylesheet should be generated when only observables are used.

__Describe the bug:__
Currently, observables are part of the static sheet but the user needs to pass `link: true` and a dynamic sheet will also be generated.

__Codesandbox link:__ 
https://codesandbox.io/s/01ro5xn32p

__Versions (please complete the following information):__
 - jss: 10.0.0-alpha.8

kof:
Why that expectation? The plan was to treat observables same  way as fn values reg. link option.

kof:
Ah in react-jss we should be able to detect observables like we do with fn values and set link: true automatically

HenriBeck:
Yeah, also observables can go into the static sheet and are actually currently only in the static sheet

kof:
Oh noo, observables should go into a dynamic sheet!!!!!

HenriBeck:
But observables are shared between all mounted components.

kof:
Right, I forget that we currently have only 2 ways to create different observables:
1. Use a function that returns styles with component together and a accepts/creates Observables
1. Use theming as a mechanism to create a new observable

kof:
Mb to support an Observable per element, we should implement function value that can return an observable, so that Observable instance can come from the props/data.

kof:
Since I forget this every time myself, we should document this current behaviour

PR #1002, opened 2019-01-27
[react-jss] Add proper support for observables in react-jss

HenriBeck:
Add better support for observables in react-jss.

Depends on #1000 

__Corresponding issue (if exists):__ #995

### DIFF
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -3,6 +3,7 @@ import React, {Component, type ComponentType, type Node} from 'react'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import {SheetsManager, type StyleSheet} from 'jss'
 import {getFunctionStyles} from 'jss-plugin-rule-value-function'
+import {getObservableStyles} from 'jss-plugin-rule-value-observable'
 import {ThemeContext} from 'theming'
 
 import type {HOCProps, Options, Styles, InnerProps} from './types'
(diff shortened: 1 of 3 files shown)


HenriBeck:
I opened the PR against the `move-get-dynamic-styles` branch, so the diff is not cluttered.

packages/react-jss/src/withStyles.js
```diff
@@ -142,7 +143,8 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
           ...contextSheetOptions,
           index,
           meta: `${displayName}, ${isThemingEnabled ? 'Themed' : 'Unthemed'}, Static`,
-          classNamePrefix: this.classNamePrefix
+          classNamePrefix: this.classNamePrefix,
+          link: getObservableStyles(themedStyles) !== null
```
kof: Now I think because of the separation in observable and function styles, we have double overhead for those 2 function calls.

kof: We could have one function and do it in one go.

kof: Also we don't need getObservableStyles, we need hasObservables() from the logic above. It could be potentially one `getDynamicStyles() : {styles, hasObservables, hasFunctions}`

HenriBeck: I think this way it's a lot cleaner and the core doesn't need to know about observables or functions in that matter.

kof: It is, but performance is the primary priority here. I think iterating over styles is quite a bit of overhead.

kof: I think we can't evaluate that, because most people don't know this API exist. We need to promote it more, but in the end, we don't know how and were it is used.

Observables are part of the default preset, so it would be not logical if they didn't work with react-jss, which states using default preset.

Also even if we would remove observables from a default preset, it wouldn't be logical if react-jss doesn't work when user defined a new jss instance with an observable plugin.

kof: I think we just need to figure out a good way to use observables with react-jss and get some use cases we can target for.

PR #1006, opened 2019-01-29
warn consumers if themed styles are misused. fix #1005

### LINKED ISSUE #1005: warn users when allegedly "themed" styles dont use themes.
__Is your feature request related to a problem? Please describe.__

Sometimes developers define styles as themed once purely out of habit without real necessity to depend on a theme.

```
export const styles = () => ({
  root: {},
});
```

Apparently people do it a lot.

__Describe the solution you'd like__
When component pulls this trick please warn developer _this component_ doesn't need to use themed styles.

__Are you willing to implement it?__
Yes

iamstarkov:

I would like to get feedback on initial implementation and get clarification

Also I didn't get how can I achieve the same `(NODE_ENV !== 'production')` functionality with `tiny-warning` module. I didnt grasp how that is achievable from it's documentation.

If implementation is ok, I'll add tests.

Also I ran `yarn` as contributing document demands but eslint complains:

```
   ✖ eslint
🚫 eslint found some errors. Please fix them and try committing again.

/Users/vlasta/projects/oss/jss/packages/react-jss/src/withStyles.js
  4:64  error  Unable to resolve path to module 'jss'  import/no-unresolved

✖ 1 problem (1 error, 0 warnings)
```

system info for eslint bug:
```
~/projects/oss/jss feat/themed-styles-misuse-warning
☯ node -v && yarn -v
v10.14.2
1.13.0
```

#1005


packages/react-jss/src/withStyles.js
```diff
@@ -39,6 +39,29 @@ const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme) => {
   if (typeof styles !== 'function') {
     return styles
   }
+  if (process.env.NODE_ENV !== 'production') {
+    if (styles.length < 1) {
+      console.warn(
+        `
+        Warning: [JSS]: This Component uses themed styles notation (function) while not relying on a theme (0 arguments).
+        It slows your app down and you should rewrite these styles to plain object notation.
```
kof: I would remove the slows down part, because one might interpret it as something that is really really slow, instead of "it could be a bit better, because styles are removed across the element". I think we can skip the explanation. Just give a hint what to do.

packages/react-jss/src/withStyles.js
```diff
@@ -39,6 +39,29 @@ const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme) => {
   if (typeof styles !== 'function') {
     return styles
   }
+  if (process.env.NODE_ENV !== 'production') {
+    if (styles.length < 1) {
```
kof: I would check for === 1, so that it includes also more than one arguments, just in case

packages/react-jss/src/withStyles.js
```diff
@@ -39,6 +39,29 @@ const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme) => {
...
+      console.warn(
```
kof: warning(false, 'error')

iamstarkov: why `false`?

packages/react-jss/src/withStyles.js
```diff
@@ -39,6 +39,29 @@ const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme) => {
...
+        `
+        Warning: [JSS]: This Component uses themed styles notation (function) while not relying on a theme (0 arguments).
+        It slows your app down and you should rewrite these styles to plain object notation.
+      `
+          .replace(/[\s]{2,}/gim, ' ')
```
kof: Why do we need that regexp?

packages/react-jss/src/withStyles.js
iamstarkov: I thought it would be something `warning(process.env.NODE_ENV !== 'production' && styles.length < 1, 'ERROR')

packages/react-jss/src/withStyles.js
kof: One generic message that covers too both cases is good enough, more than one argument is not something that many will have

packages/react-jss/src/withStyles.js
iamstarkov: it is multiline message so it will keep all indentation when being printed without adjusting

packages/react-jss/src/withStyles.js
kof: no, it is designed to use like this `warning(isValueCorrect, 'error')` so if value is correct no warning happens

kof: like an assertion

packages/react-jss/src/withStyles.js
iamstarkov: that's the second check. I separated these two cases, because 1st one is perf concern for sure, the 2nd one is misuse concern of unknown origin

packages/react-jss/src/withStyles.js
kof: warning module does the check for env, but we want to avoid the .length check in production here, so we should have the production cheeck like you did

packages/react-jss/src/withStyles.js
kof: I think this is something that should be evtl done in warning module

kof: or keep errors short

packages/react-jss/src/withStyles.js
kof: Yeah, I would make it one and write a message that covers both cases. Its not a huge problem we need to be particularly explicit about.

kof: "Styles function expects one `theme` argument." or something

packages/react-jss/src/withStyles.js
HenriBeck: I think the warning calls are wrapped automatically with the env check now.

HenriBeck: Also `!== 1` please

kof: yes, but we don't need the .length check in runtime either, I am not sure its cheap

kof: actually we can check, lets see quickly a bench

kof: https://esbench.com/bench/5c5078f94cd7e6009ef6209d

ok I think its fine to use fn.length at runtime, so we can just write warning(styles.length === 1, 'error') without adding env checks

kof:
regarding the lint error, does it go away if you run `yarn`?

packages/react-jss/src/withStyles.js
iamstarkov: I disagree.

> it could be a bit better

This part will give anyone free pass to not do any optimisations and people will complain to remove the warning.

> styles are removed across the element

I am not sure I understand what does it mean.

packages/react-jss/src/withStyles.js
iamstarkov: I would argue that warning messages should not only warn but also provide suggestions for better dx

iamstarkov:
> regarding the lint error, does it go away if you run yarn?

no

iamstarkov:
also should we pass the component name to the getStyles function to provide better contextual feedback?

iamstarkov:
here is a complete log of "import/no-unresolved" violations https://pastebin.com/HfBhv3q1

HenriBeck:
Also please update the changelog

iamstarkov:
> Run yarn build

@HenriBeck, this helped, thanks

packages/react-jss/src/withStyles.js
iamstarkov: this warning will be stripped away from prod build anyway, right?

HenriBeck: Yeah, all of the warning calls are automatically wrapped with `process.env.NODE_ENV === 'development'`.

kof: Just checked, yes it gets removed in the minified version, even if it isn't wrapped with evn conditional. Not sure who removes it though :)

HenriBeck: All of the warning come this in the bundled code:
```
if (process.env.NODE_ENV !== 'production') {
  warning();
}
```

In the minified umd version this will be removed because we replace the node env. In the commonjs and ESM, it will exist until the user bundles the code.

kof: true,  something like this

```
        warning(false, "[JSS] Bad keyframes name " + key);
```

becomes in the ESM bundle

```
      process.env.NODE_ENV !== "production" ? warning(false, "[JSS] Bad keyframes name " + key) : void 0;

```

so later uglify probably removes it once conditional becomes 'production' !== 'production'

packages/react-jss/src/withStyles.js
iamstarkov: I don't see why cant we provide specific warning explanations for each case, because it will be removed from prod build anyway

kof: I think we should start adding links to the warnings with a really complete full explanation. An explanation that can't be complete is not much better than a really short one without an explanation at all in my opinion.

kof: For example your explanation was confusing for me.

packages/react-jss/src/withStyles.js
```diff
@@ -35,32 +36,20 @@ let managersCounter = 0
...
-        React-jss provides only one argument.
-      `
-          .replace(/[\s]{2,}/gim, ' ')
-          .trim()
-      )
-    }
-  }
+  warning(
+    styles.length < 1,
+    `Warning: [JSS]: <${displayName} />'s styles function doesn't rely on a theme. We recommend to rewrite it to plain object.\nRead more: https://github.com/cssinjs/jss/blob/master/docs/react-jss.md#basic`
```
kof: what if the styles function is used with multiple components?

kof: We have to discuss the format of the links we use and the content. I was actually thinking of separate documents, which explain exactly the error, not just generic documents. Basically, the way react does that. But its a bit out of scope for this issue.

iamstarkov: that doesn't change the misuse scenario in one or all of the components.

in my experience (few thousands components) people do it only when dont have time to split big stylesheet into separate ones and keep adding styles in one stylesheet for several components.

kof: yeah, could be good enough, I don't see a good other pointer except of the stack trace, fn.name won't be useful since its usually `styles`

packages/react-jss/src/withStyles.js
```diff
@@ -35,10 +36,21 @@ let managersCounter = 0
 
 const NoRenderer = (props: {children?: Node}) => props.children || null
 
-const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme) => {
+const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme, displayName: string) => {
   if (typeof styles !== 'function') {
     return styles
   }
+  warning(
+    styles.length < 1,
```
HenriBeck: Let's check for `=== 0` here?

packages/react-jss/src/withStyles.js
HenriBeck: Why would it be a problem if it's used multiple times?

kof: Its just a bit of indirection, it might be used in multiple but will mention one. But its ok I guess.

packages/react-jss/src/withStyles.js
iamstarkov: lets do one

packages/react-jss/src/withStyles.js
kof: We need to get rid of the links, I don't want to have issues complaining about broken links. We need permalinks and full issue descriptions.

iamstarkov:
I think eslint on travis has the same problem as I did https://travis-ci.org/cssinjs/jss/builds/486923274?utm_source=github_status&utm_medium=notification
screenshot 2019-01-31 at 2 12 36 pm

HenriBeck:
Merge master in this branch again

iamstarkov:
@HenriBeck didn't help

iamstarkov:
also I have problems with the following test case. I forgot to add it from the start, and now am struggling with karma. I suspect karma isn't configured to propagate process variable into the karma env.

```js
    it.only('doesnt warn if themed styles dont use theme in _prod_', () => {
      debugger;
      process.env.NODE_ENV = 'production'
      function DisplayNameTest() {
        return null
      }
      const MyComponent = withStyles(() => ({}))(DisplayNameTest) // eslint-disable-line no-unused-vars

      TestRenderer.create(<MyComponent />)

      expect(console.warn.called).to.be(false)
      process.env.NODE_ENV = 'development'
    })
```

screenshot 2019-01-31 at 3 30 31 pm

HenriBeck:
@iamstarkov try now again, forgot that I hadn't merged the PR which fixes it yet.

packages/react-jss/src/withStyles.js
kof: If you don't mind I will change the error message, we have a separate issue for linking errors to some permlink with good deescriptions

iamstarkov:
it works now, but aforementioned test case isn't passing. Can someone else take a look at it?

HenriBeck:
Will have a look tomorrow.

HenriBeck:
@iamstarkov the test is passing for me locally, is the test still a problem?

iamstarkov:
@HenriBeck it doesn't pass for me locally. But I committed it now, lets see what travis thinks

HenriBeck:
I fixed the test, though a different test is now failing, not sure why though.

HenriBeck:
Is this good to merge?

iamstarkov:
@HenriBeck  from the implementation point of view it is good to merge

HenriBeck:
Alright, merging it then. I don't think we need to add any kind of documentation for it.

iamstarkov:
yay

kof:
I am removing the link though.

iamstarkov:
ok

Issue #969, opened 2019-01-04
TypeScript guide


NicholasLYang:
__Is your feature request related to a problem? Please describe.__
I've been scratching my head over typing JSS, specifically React JSS. I've read [this post](https://github.com/cssinjs/react-jss/issues/194) and tried to implement what they suggested, but it kind of falls apart with theming. In addition, I've noticed that there's already typings in the repo, but I can't make heads or tails of them without some usages. I've also tried looking at Material UI, but it's a little overwhelming.

__Describe the solution you'd like__
I'd really appreciate some examples of typing styles and how to implement theming, dynamic values, etc. Also a quick runthrough of the rational behind typings would be fantastic. One of the important, yet frustrating parts of TS is that typings can be written at several levels of granularity and choosing the level is extremely important.

__Are you willing to implement it?__
Yes! However, I'd need some guidance on the right way to use the typings.

HenriBeck:
Sound great. Feel free to create a PR.

HenriBeck:
Yes, should be in the docs under `Usage with typescript`.

@NicholasLYang No, we are in the process of refining the types right now and adding tests for them.

Saturate:
I've used React-JSS and TypeScript for some time, feel free to ask any questions - I might be able to answer.

NicholasLYang:
Hey @Saturate, I'm just struggling with the basics here. Let's say I'm trying to inject styles into a simple component, no theming, no dynamic values. Any chance you could provide a code snippet?

Saturate:
@NicholasLYang 

Sure the following should be fine for you:

```
import * as React from 'react';
import injectSheet from 'react-jss';

const styles = {
	root: {
		color: 'red',
		fontWeight: 600
	}
};

export interface IHeadingProps {
	children: React.ReactNode;
}

class Heading extends React.Component<IHeadingProps> {
	public render() {
		return (
			<h1 className={this.props.classes.root}>{this.props.children}</h1>
		);
	}
}

export default injectSheet(styles)(Heading);
```

HenriBeck:
@Saturate is this for the new typings or the old ones in definitely typed?

NicholasLYang:
@Saturate, I'm a little confused by your example. `classes` isn't included as a prop, which gives an error for me in create-react-app (with TS). Also, styles is implicitly typed as a literal (due to const), which probably won't play nice with the injectSheet typing unless you widen the type, no?

Saturate:
[...]

PR #1009, opened 2019-01-31
Added TypeScript Usage documentation for React JSS

NicholasLYang:
Just a simple example of typing with React JSS. Happy to add more details if necessary.
#969

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -569,3 +569,36 @@ const Button = withStyles(buttonStyles)(() => (
   </button>
 ))
 ```
+
+### Usage with TypeScript
+
+React JSS includes first class support for TypeScript. React JSS provides
+ a `WithStyles` type which adds types for all of the injected props.
+  To use it, simply extend your existing props interface with
+   `WithStyles<typeof styles>`, where `styles` is your styles object.
+
+ Example
+
+ ```typescript
+ import * as React from 'react'
+ import withStyles, { WithStyles } from 'react-jss'
+
+ const styles = {
+   button: {
+       backgroundColor: 'yellow'
+     },
+   label: {
+     fontWeight: 'bold'
+   }
+ };
+
+ interface IProps extends WithStyles<typeof styles> {
+   children: React.ReactNode
+ }
+
+ const Button: React.FunctionComponent = ({classes, children}) => (
+   <button className={classes.button}>
+     <span className={classes.label}>{children}</span>
+   </button>
+ )
+ ```


docs/react-jss.md
```diff
@@ -569,3 +569,36 @@ const Button = withStyles(buttonStyles)(() => (
...
+   label: {
+     fontWeight: 'bold'
+   }
+ };
+
+ interface IProps extends WithStyles<typeof styles> {
+   children: React.ReactNode
+ }
+
+ const Button: React.FunctionComponent = ({classes, children}) => (
```
HenriBeck: Shouldn't it be `React.FunctionComponent<IProps>`?

docs/react-jss.md
```diff
@@ -569,3 +569,36 @@ const Button = withStyles(buttonStyles)(() => (
...
+   <button className={classes.button}>
+     <span className={classes.label}>{children}</span>
+   </button>
+ )
```
HenriBeck: Also please add:
```js
export default withStyles(styles)(Button);
```

HenriBeck:
@kof Not sure if the docs for typescript should be in the individual package documentation or we have a global doc `Usage with Typescript`. What do you think?

docs/react-jss.md
NicholasLYang: Whoops! Can't believe I forgot that. Fixed

docs/react-jss.md
```diff
@@ -569,3 +569,38 @@ const Button = withStyles(buttonStyles)(() => (
   </button>
 ))
 ```
+
+### Usage with TypeScript
+
+React JSS includes first class support for TypeScript. React JSS provides
+ a `WithStyles` type which adds types for all of the injected props.
+  To use it, simply extend your existing props interface with
+   `WithStyles<typeof styles>`, where `styles` is your styles object.
```
HenriBeck: Not sure where this indentation comes from, but if possible remove it.

HenriBeck:
@NicholasLYang we should also add `Usage with Typescript` to the tables of content on top of the readme

NicholasLYang:
Added it to the table. I noticed that we don't have `Injection order` in there either. I can open a different PR for that if you'd like

sheenm:
But it's unclear how to get intellisense when writing styles object 
That would be perfect, if I could write 
```
const styles = {
    button: {
        backgrndColor: 'yellow'
    }
}
```
and get some help from editor, and get error from typescript compiler since there's no css property backgrndColor. Or I could get autocomplection.

Can somebody help with it and add this to documentation?

HenriBeck:
@sheenm There is an open issue for adding types to the styles object but it's not as trivial as it seems.

#806

PR #1014, opened 2019-02-02
[react-jss] Export the context as __Context from react-jss

HenriBeck:
I need to export the context to build the new hooks package with it.
We shouldn't document it because it is a really advanced feature and should only be used in very rare cases.


packages/react-jss/src/index.js
```diff
@@ -3,4 +3,5 @@ export {ThemeProvider, withTheme, createTheming} from 'theming'
 export {default as JssProvider} from './JssProvider'
 export {default as jss} from './jss'
 export {SheetsRegistry, createGenerateId} from 'jss'
+export {default as __Context} from './JssContext'
```
kof: why not export as  JssContext?

HenriBeck: It shouldn't be used by the public. `react-router` also exports their context as `__Context`. 

I just looked at `react-redux` and they also just export their context.

kof: I think we should not document it, thats it.

Issue #1012, opened 2019-02-01
CssVars should not be camel-cased


texels:
__Expected behavior:__
I tried to put a css variable into the jss object on the jss playground:
 "--helloWorld" : "red",
I expected this to remain untouched. Instead it was converted to:
  --hello-world: red;

__Describe the bug:__
css variables should not be camel-cased

__Codesandbox link:__ 
Enter this into the jss playground:

export default {
  button: {
    "--helloWorld" : "red",
    border: '1px solid var( --helloWorld )',
  }
};

__Versions (please complete the following information):__
 - jss: The one on https://cssinjs.org/repl/
 - Browser: chrome
 -OS: macOS

Feel free to add any additional versions which you may think are relevant to the bug.

kof:
I don't know if this is a bug, seems consistent to me, camel cased props always converted to hyphenated.

kof:
Oh I get it, you wrote camel cased custom prop, but because of the conversion you have to use a hyphenized version of it in (var()), so it becomes inconsistent from the  consuming perspective, unless you write the first one also hyphenised.

PR #1017, opened 2019-02-02
[jss-plugin-camel-case] Support css vars better

HenriBeck:
Currently, css variables are also being hyphenated.

__Corresponding issue (if exists):__ #1012

### DIFF
--- a/packages/jss-plugin-camel-case/src/index.js
+++ b/packages/jss-plugin-camel-case/src/index.js
@@ -12,7 +12,9 @@ function convertCase(style) {
   const converted = {}
 
   for (const prop in style) {
-    converted[hyphenate(prop)] = style[prop]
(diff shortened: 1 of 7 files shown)


packages/jss-plugin-camel-case/src/index.js
```diff
@@ -12,7 +12,9 @@ function convertCase(style) {
   const converted = {}
 
   for (const prop in style) {
-    converted[hyphenate(prop)] = style[prop]
+    const key = prop.startsWith('--') ? prop : hyphenate(prop)
```
kof: created a bench, https://esbench.com/bench/5c55f3b54cd7e6009ef620ec 
`prop[0] === '-' && prop[1] === '-'` seems to be much faster, its on a level where it probably doesn't matter but still kinda nice to have

PR #1018, opened 2019-02-03
[docs] Update SSR setup with react-jss

HenriBeck:
This fixes resetting the `generateId` for SSR in v10.
This was reported in gitter.

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -321,14 +321,15 @@ After the application is mounted, you should remove the style tag used by critic
 ```javascript
 import React from 'react'
 import {renderToString} from 'react-dom/server'
-import {JssProvider, SheetsRegistry} from 'react-jss'
+import {JssProvider, SheetsRegistry, createGenerateId} from 'react-jss'
 import MyApp from './MyApp'
 
 export default function render(req, res) {
   const sheets = new SheetsRegistry()
+  const generateId = createGenerateId()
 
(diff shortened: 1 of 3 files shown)


kof:
Looks good, but without a test I don't know what we are fixing

HenriBeck:
The problem is that we are not resetting the generateId when someone tries to do SSR.
Lucien PESLIER reported this on gitter.

kof:
I get the description, a bug needs a test though, otherwise its gonna be broken again.

HenriBeck:
Yeah, so I tested quite a bit now, and I wasn't able to fix the bug. I think we need to require the user to regenerate it on every request.

kof:
Did we change any behaviour from 9 to 10 here?

HenriBeck:
Yeah, the context. This is the problem.

kof:
Not sure how.

HenriBeck:
When you call `renderToString` with the same root node, the context is being persisted. This means you second `renderToString` will still have the same context as the first one. That's why the counter is increasing. 

Not sure if this is a bug or a feature because if you call `render` with the same root node, it updates the node instead of re-rendering it which makes sense in the browser.

kof:
I think its just a thing we need to document. Its how the new context works.

HenriBeck:
Yeah, that's why I now changed the docs in this pr

PR #1024, opened 2019-02-08
[react-jss] Add react stylesheet defintion for ts autocomplete

dleavitt:

Right now there doesn't seem to be a way to get typechecking/autocompletion when defining style rules.

This adds a type definition for style rule definitions, so you can go like this:
```typescript
const styles: StyleSheet = {
  myButton: {
    color: 'green',
  }
}
```

and valid CSS properties will typecheck/autocomplete. I think the [definitely typed defs](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jss/css.d.ts) do some fancier version of this. This isn't ready to merge or anything, b/c:

- [ ] Only done for react-jss
- [ ] I'm not sure the type def is right - are style rules actually arbitrarily recursive?
- [ ] `StyleSheet` is probably a bad name since it's a browser global
- [ ] The `StyleSheet` type should probably be used in place of the `Style` type at use sites.
- [ ] Not tested - not sure there's a good way to do it with the test suite.
- [ ] Etc! Probably other things.

Happy to take a stab at figuring the above out if this seems like a useful approach.

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -46,11 +46,11 @@ Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
 ```javascript
 import React from 'react'
 import {render} from 'react-dom'
-import withStyles from 'react-jss'
+import withStyles, {StyleSheet} from 'react-jss'
 
 // Create your Styles. Remember, since React-JSS uses the default preset,
 // most plugins are available without further configuration needed.
-const styles = {
+const styles: StyleSheet = {
   myButton: {
     color: 'green',
     margin: {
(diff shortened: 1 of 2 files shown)


HenriBeck:
There is an open ticket for adding types to the stylesheet itself. The problem with that is that we have a lot of ways on how to write styles currently, so for example nested rules, expanded values, function rules/values, default values, etc.

And because the user in the runtime applies them, this makes it very hard to type the stylesheet currently correctly. 

So in your types, nested rules would still be allowed even if the user doesn't use the nested plugin. Also, this includes only a partial type-safety. It will not warn against typos or wrong literal values for specific properties.

HenriBeck:
I will close this PR for now, as the types should be added inside `jss` and the typings are incomplete.

dleavitt:
oops, sorry for the late reply. makes sense that the stylesheets are hard to type properly. I sort of think a permissive + optional approach like this might make sense given the constraints you described?

omasback:
I tried out this fork and it was a huge improvement to have css property autocomplete. I think that's gonna be a dealbreaker for a lot of people. Hope this functionality makes it into the lib soon!

Issue #1026, opened 2019-02-09
[react-jss] 10.0.0-alpha.9 does not remove previously applied styles


bhupinderbola:
__Expected behavior:__
passing {borderColor: null} should remove the previously applied borderColor.

__Describe the bug:__
Unable to remove the previously applied style.

__Codesandbox link:__ 
https://codesandbox.io/s/j3nqovqozy
vs
https://codesandbox.io/s/z3mromo1y3

__Versions (please complete the following information):__
 - jss: 10.0.0-alpha.9
 - Browser [e.g. chrome, safari]: Chrome
 - OS [e.g. Windows, macOS]: Windows

I would like to help with fixing the bugs, but I am unable to build it locally. I am not sure if it is my local issue or a code issue.
Running "yarn build" was used to give me sizeSnapshot related error. After I commented out rollup-plugin-size-snapshot plugin, I started getting following error:
Error: Could not resolve './SheetsRegistry.js' from ../jss/dist/jss.esm.js

HenriBeck:
This seems to be an issue with `react-jss` or `jss-plugin-rule-value-function` itself, because when I try to change values (remove them from a rule) with `.prop` everything works.

https://codesandbox.io/s/wqmk2w5z98

HenriBeck:
![screenshot 2019-02-13 at 13 20 00](https://user-images.githubusercontent.com/8514706/52711448-1bd22b80-2f92-11e9-83a7-82cda8307371.png)

HenriBeck:
This seems to be an issue with updating a rule: https://codesandbox.io/s/qlvxmpqwjq

kof:
I have an assumption that we remove null values somewhere in react-jss, so the property doesn't get unset.

kof:
Ok it turned out we never removed the props when update comes from a fn rule specifically.

kof:
fixed in 10.0.0-alpha.12

PR #1028, opened 2019-02-09
[jss] Remove prop when null is returned from a function rule/value

HenriBeck:
Currently, a prop is not removed when a function value returns undefined.

I got two tests to fail. I narrowed down to the problem that `renderable` is not defined in the `StyleRule`: https://github.com/cssinjs/jss/blob/master/packages/jss/src/plugins/styleRule.js#L69

@kof can you also please have a look why it's not defined?

__Corresponding issue (if exists):__ #1026

### DIFF
--- a/packages/jss-plugin-rule-value-function/src/function-values.test.js
+++ b/packages/jss-plugin-rule-value-function/src/function-values.test.js
@@ -284,6 +284,7 @@ describe('jss-plugin-rule-value-function: Function values', () => {
       })
 
       it('should remove declarations when value is null', () => {
+        sheet.update({color: 'green'})
         sheet.update({color: null})
 
         expect(sheet.toString()).to.be(stripIndent`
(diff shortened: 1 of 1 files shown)


kof:
it seeems with your changnes only this tets fails "should remove declarations when value is undefined"

HenriBeck:
Yeah exactly, the problem with the test was that we updated the values, but the values didn't have a value before, so there were no properties to remove.

HenriBeck:
Oh, wait yeah, this test is failing because when we pass undefined, we treat `prop` as a getter instead of a setter, so returning `undefined` from a function value doesn't have any effect.

kof:
yes and that makes sense, I don't even know why we have that test there ... on the other hand the other tests with null don't fail

HenriBeck:
They don't fail because we test `toString` and this works, but it will not remove the property from the rule.

kof:
in that case we should be able to reproduce it it in a functional case in the core

PR #1029, opened 2019-02-10
Upgrade all esm ready packages

TrySound:
In this diff I upgraded

- hyphenate-style-name
- css-initials
- css-vendor

which were converted to esm.

This means, that all jss and all plugins may be consumed by unpkg with
module flag like https://unpkg.com/jss-preset-default?module

We probably may get rid from unpkg prop in favour of `?module` flag for
non react packages.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -20,8 +20,6 @@ const input = path.join(rootPath, './src/index.js')
 
 const name = camelCase(pkg.name)
 
-const reactPkg = pkg.name === 'react-jss' || pkg.name === 'jss-starter-kit'
-
 const globals = {
   jss: 'jss',
   'react-jss': 'reactJss',
@@ -50,6 +48,11 @@ const getBabelOptions = ({useESModules}) => ({
 })
 
 const commonjsOptions = {
+  include: [
+    /\/node_modules\/react\//,
+    /\/node_modules\/prop-types\//,
+    /\/node_modules\/hoist-non-react-statics\//
+  ],
   ignoreGlobal: true,
   // The CommonJS plugin can't resolve the exports in `react` automatically.
   // https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports
@@ -88,7 +91,7 @@ export default [
     plugins: [
       nodeResolve(),
       babel(getBabelOptions({useESModules: true})),
-      reactPkg && commonjs(commonjsOptions),
+      commonjs(commonjsOptions),
       nodeGlobals({process: false}),
       replace({
         'process.env.NODE_ENV': JSON.stringify('development'),
@@ -111,7 +114,7 @@ export default [
     plugins: [
       nodeResolve(),
       babel(getBabelOptions({useESModules: true})),
-      reactPkg && commonjs(commonjsOptions),
+      commonjs(commonjsOptions),
       nodeGlobals({process: false}),
       replace({
         'process.env.NODE_ENV': JSON.stringify('production'),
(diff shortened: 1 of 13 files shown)


rollup.config.js
```diff
@@ -86,7 +88,7 @@ export default [
     plugins: [
       nodeResolve(),
       babel(getBabelOptions({useESModules: true})),
-      commonjs(commonjsOptions),
+      reactPkg && commonjs(commonjsOptions),
```
kof: Why do we need it?

TrySound: To make sure we will never ever add another commonjs package in our dependencies.

kof: but why this conditional and not for everything?

TrySound: Because we dropped commonjs for everything except react based packages.

kof: it just seems like we are addng there another list to maintain

TrySound: I don't think it will be changed before react will accept esm.

kof:
@AleshaOleg is looking into css-vendor problem

yarn.lock
```diff
@@ -5,21 +5,18 @@
 "@babel/code-frame@^7.0.0":
   version "7.0.0"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0.tgz#06e2ab19bdb535385559aabb5ba59729482800f8"
-  integrity sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==
```
TrySound: @kof Update your yarn please

kof: damn, what happened here? we have 1.9 in engines

kof: should we have 1.14?

kof: no 1.13, 1.14 is rc

TrySound: Yes, I have 1.13. integrity hash was added after 1.9

AleshaOleg:
Tests are passed, but red cross still there. Who knows what's wrong?

yarn.lock
kof: strange, upgraded, but it doesn't add themm back

kof:
there is another issue https://travis-ci.org/cssinjs/jss/builds/491780740#L612 with flow, it can't find react-jss dependency, I have the same locally

yarn.lock
TrySound: Revert yarn.lock and run yarn again

AleshaOleg:
Ah, ok. I had it too today. I did `yarn run build` before running tests, and then this error went away.

kof:
for me run build doesn't fix this, neither it does for the ci

HenriBeck:
Will also take a look at the issue

rollup.config.js
```diff
@@ -48,6 +48,11 @@ const getBabelOptions = ({useESModules}) => ({
 })
 
 const commonjsOptions = {
+  include: [
+    /\/node_modules\/react\//,
```
HenriBeck: This here creates the issue with flow.

HenriBeck: @TrySound @kof

TrySound: I get it. We use ancient rollup version which silently swallowed commonjs errors without emitting files.

TrySound: Upgrading it in separate branch

HenriBeck: So upgrading to rollup v1 would solve the issue?

TrySound: It will show the real error which is swallowed now.

TrySound: Seems like the error was in missed `react-display-name` package which is a transitive dependency of `theming`.

HenriBeck:
Looks good now 👍

PR #1030, opened 2019-02-10
[react-jss] Fix TS type definitions

HenriBeck:
This adds the JssContext to the typescript definitions.


kof:
Changelog?

PR #1031, opened 2019-02-11
Upgrade babel and rollup

TrySound:
- upgraded babel including runtime (to dedupe it later)
- upgraded rollup to fix the problem with swallowed errors
- replaced uglify with less buggy terser


TrySound:
@HenriBeck I don't get why the build is failed? I don't see errors.

AleshaOleg:
@TrySound try to re-run built.

Probably tests didn't pass because of this line: https://travis-ci.org/cssinjs/jss/builds/491830017#L1302

Sometimes it's happening for IE instances, it just disconnects

HenriBeck:
I rerun the build if it still has problems with karma and browserstack we can just merge it.
We really need to switch to jest for more reliable tests and to run tests of external contributors.

HenriBeck:
This is good to merge, the tests only fail because of browserstack timeouts

PR #1040, opened 2019-02-21
Fix dynamic values inside of @global #664

kof:
#664

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+## Next
+
+- [jss] Fix dynamic values inside of @global plugin ([#664](https://github.com/cssinjs/jss/pull/664))
+
 ## 10.0.0-alpha.10 (2019-2-9)
 
 ### Improvements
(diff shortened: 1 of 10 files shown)


packages/jss/src/renderers/DomRenderer.js
```diff
@@ -255,10 +256,10 @@ const getNonce = memoize(
 )
 
 const insertRule = (
-  container: CSSStyleSheet | CSSKeyframesRule | CSSMediaRule,
+  container: CSSStyleSheet | CSSMediaRule | CSSKeyframesRule,
   rule: string,
   index?: number = container.cssRules.length
-): false | Object => {
+): false | any => {
```
HenriBeck: Is there no way around `any` here?

kof: Ah, no this was temp change, but it doesn't matter when we update to the latest flow version, Object will equal any

kof: I think we can actually do much better types now, many anys are just there because of initial lack of experience with flow

changelog.md
```diff
@@ -1,3 +1,7 @@
+## Next
+
+- [jss] Fix dynamic values inside of @global plugin ([#664](https://github.com/cssinjs/jss/pull/664))
```
HenriBeck: This should point to the PR and not the issue

PR #1043, opened 2019-02-23
Add support for dynamic keyframes

HenriBeck:
This fixes having a dynamic animation name.

I think we can have a better system to process the update style changes and not require this duplicate code.

Still missing some tests to make sure we don't break anything.

### DIFF
--- a/packages/jss/src/plugins/keyframesRule.js
+++ b/packages/jss/src/plugins/keyframesRule.js
@@ -85,16 +85,16 @@ const findReferencedKeyframe = (val, keyframes) => {
   if (typeof val === 'string') {
     const ref = refRegExp.exec(val)
 
-    if (!ref) return null
+    if (!ref) return val
 
     if (ref[1] in keyframes) {
-      return val.replace(val, keyframes[ref[1]])
+      return val.replace(ref[0], keyframes[ref[1]])
     }
 
     warning(false, `[JSS] Referenced keyframes rule "${ref[1]}" is not defined.`)
   }
 
-  return null
+  return val
 }
 
 /**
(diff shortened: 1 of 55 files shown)


HenriBeck:
@kof please review.

packages/jss/src/plugins/keyframesRule.js
```diff
@@ -82,36 +81,66 @@ const keyRegExp = /@keyframes\s+/
...
+  if (refKeyframe !== value) {
+    style[prop] = refKeyframe
   }
 }
 
-export default {
-  onCreateRule(key: string, frames: JssStyle, options: RuleOptions): KeyframesRule | null {
-    return keyRegExp.test(key) ? new KeyframesRule(key, frames, options) : null
+const plugin: Plugin = {
+  onCreateRule(key, frames, options) {
```
kof: why removing the types?

kof: ah gotcha

packages/jss/src/plugins/keyframesRule.js
```diff
@@ -82,36 +81,66 @@ const keyRegExp = /@keyframes\s+/
...
 
     if ('animation-name' in style) replaceRef(style, 'animation-name', sheet.keyframes)
     if ('animation' in style) replaceRef(style, 'animation', sheet.keyframes)
     return style
+  },
+
+  onChangeValue(val, prop, rule) {
+    const {sheet} = rule.options
+
+    if (rule.type !== 'style' || !sheet) {
```
kof: afaik only style rule value can be changed, so the first check can be removed

PR #1046, opened 2019-02-25
Fix removing styles from function rules 

kof:
#1026

### DIFF
--- a/packages/jss/tests/functional/sheet.js
+++ b/packages/jss/tests/functional/sheet.js
@@ -515,6 +515,22 @@ describe('Functional: sheet', () => {
     })
   })
 
+  describe('sheet.update()', () => {
+    let sheet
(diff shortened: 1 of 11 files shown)


packages/jss/tests/functional/sheet.js
```diff
@@ -515,6 +515,22 @@ describe('Functional: sheet', () => {
     })
   })
 
+  describe('sheet.update()', () => {
```
HenriBeck: Please also add a test where the property becomes `null` and `false`

kof: will do

HenriBeck:
A changelog is missing as well.

kof:
> A changelog is missing as well.

Forgot to push, fixed

Issue #1037, opened 2019-02-17
[react-jss] Use static sheet for dynamic rules


HenriBeck:
__Is your feature request related to a problem? Please describe.__
Currently, we are rendering for each mounted component a dynamic stylesheet which contains the dynamic rules.

__Describe the solution you'd like__
Similar to `styled-jss`, we insert the dynamic rules dynamically into the static sheet of the component.

This would substantially remove the count of stylesheets inside a single page.

__Are you willing to implement it?__
Yes

HenriBeck:
This was implemented in #1048

PR #1048, opened 2019-03-02
[react-jss] Add dynamic rules to static sheet

HenriBeck:
This adds the dynamic rules to the static sheet instead of creating a new sheet for every component instance.

I haven't updated the tests, but this should also fix the issues with having to different sheets in `react-jss` (`compose` behaves a little bit different, dynamic keyframe names aren't possible).

__Corresponding issue (if exists):__ #1037

### DIFF
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -232,7 +232,7 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
         const rules = {};
 
         for (const key in sheet.dynamicStyles) {
-          rules[key] = sheet.addRule(`${key}-${this.instanceId}`, rules[key])
+          rules[key] = sheet.addRule(`${key}-${this.instanceId}`, sheet.dynamicStyles[key])
         }
 
         return rules;
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -28,11 +28,13 @@ interface State {
  * So our solution is to render sheets them in the reverse order child->sheet, so
  * that parent has a higher specificity.
  */
-let indexCounter = -100000
+let indexCounter = -Number.MAX_SAFE_INTEGER
 
 let managersCounter = 0
 
 const NoRenderer = (props: {children?: Node}) => props.children || null
+
+const sheetsMeta = new WeakMap()
 
 const getStyles = <Theme: {}>(styles: Styles<Theme>, theme: Theme, displayName: string) => {
   if (typeof styles !== 'function') {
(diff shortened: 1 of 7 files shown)


HenriBeck:
Benchmarks with shared sheet between static rules and dynamic rules:
screenshot 2019-03-03 at 19 46 11

Benchmarks from v10.0.0-alpha.12
screenshot 2019-03-03 at 19 49 00

kof:
did you try to compare the latest v10 vs this changes?

HenriBeck:
> did you try to compare the latest v10 vs this changes?

Yeah, I did that. The `react-jss@8.6.1` are the changes from this PR

kof:
oh so we have twice better performance for dynamic styles with this PR and the other cases are the same?

HenriBeck:
Yeah, with this changes we do more in the scripting part as we need to figure out which rules to update, which leads to a bit worse performance in mounting.

HenriBeck:
So for users who don't use dynamic styles, there is no downside/performance loss with this code.

For users who have dynamic styles, there is a slightly higher mounting time as we need to do more calculations instead of just creating a stylesheet.

@kof do you think it's worth it?

packages/react-jss/src/withStyles.js
```diff
@@ -79,6 +77,61 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
...
+
+        return classes
+      }
+
+      static updateDynamicRules(props: HOCProps<Theme, Props>, {dynamicRules, sheet}: State) {
+        if (!sheet) {
+          return
+        }
+
+        for (const key in dynamicRules) {
```
kof: not sure why you need this loop? if you don't pass the first argument as a key, JSS will already do the same it seems

packages/react-jss/src/withStyles.js
```diff
@@ -79,6 +77,61 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
...
+          // $FlowFixMe: Not sure why it throws an error here
+          sheet.update(dynamicRules[key].key, props, {})
```
kof: why is the last argument empty object? You want to avoid plugins execution on the return values?

packages/react-jss/src/withStyles.js
kof: Ah got it, you need to only update rules from a specific component instance, since that sheet will contain all dynamic values for all instances. We should add a comment, this isn't obvious.

packages/react-jss/src/withStyles.js
```diff
@@ -79,6 +77,61 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
...
+        }
+
+        for (const key in dynamicRules) {
+          sheet.deleteRule(dynamicRules[key].key)
+        }
+      }
+
+      static addDynamicStyles(sheet: StyleSheet): ?DynamicRules {
+        // $FlowFixMe Cannot access random fields on instance of class StyleSheet
+        if (!sheet.dynamicStyles) return undefined
```
kof: we could save dynamicStyles on a WeakMap, since we started using it anyways

kof: actually why not on state?

packages/react-jss/src/withStyles.js
HenriBeck: I think flow also complained here because it couldn't figure which overload to use I believe.

packages/react-jss/src/withStyles.js
HenriBeck: Will add a comment

packages/react-jss/src/withStyles.js
HenriBeck: Because we only need it once per sheet to be extracted

kof: right, then weakmap?

packages/react-jss/src/withStyles.js
```diff
@@ -79,6 +77,61 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
       // $FlowFixMe
       static defaultProps = {...InnerComponent.defaultProps}
 
+      static getSheetClasses(sheet, dynamicRules: ?DynamicRules) {
```
kof: actually we don't need this to be a static method, it can be a regular function outside of class

HenriBeck: Yeah, so for now I would like them to be `static`, once I start writing the hook I will most likely move them out and make most functions implementation independent for reuse most of the code.

kof: Why do you want them to be static?

packages/react-jss/src/withStyles.js
HenriBeck: Why would we need a WeakMap? There is only one dynamic styles object per sheet and we should save it on the sheet, we actually are already doing this.

kof: I did this because I didn't want to rely on WeakMap polyfill back then.  We could have a single WeakMap and use sheet as a key.

kof: Adding a property to a class instance is always a very dirty appoach. A less dirty would be to use a random property or a Symbol.

HenriBeck: But the sheet is already in a WeakMap. I don't think it makes sense to have another WeakMap.

packages/react-jss/src/withStyles.js
HenriBeck: I think right now they are part of the component, but I want to abstract the whole creating/updating/removing of the sheets for the hook.

packages/react-jss/src/withStyles.js
HenriBeck: We are only using the sheet internally and the can't access it, so I don't think it's a big deal if we don't use a symbol.

kof: where?

HenriBeck: The SheetsManager is using a WeakMap now. Also, the dynamic styles shouldn't be GC collected as long as the sheet is still in memory.

kof: It won't be collected as long as the sheet is referenced.  I still think we should find a way to not attach properties to the sheet instance.

HenriBeck: Well yeah, it won’t be collected until the sheet is received and I think this is exactly what we want. Not attaching it would be nice, but it would also come with a downside. Mainly the extra time to find the dynamic styles in another WeakMap

kof: Are you talking about map.get(ref) perf cost? This is so fast we don't even need to talk about it.

HenriBeck: I think this would add an extra layer of complexity to it

kof: basically dynamicStyles = new WeakMap and dynamicStyles.get(sheet) or am I missing somethign?

HenriBeck: Yeah, we would need three WeakMaps because we also now need to track the static styles and the number of dynamic rules, so I don't think it would make sense to have these right now.

packages/react-jss/src/withStyles.js
```diff
@@ -80,6 +78,66 @@ export default function withStyles<Theme: {}, S: Styles<Theme>>(
...
```
kof: We could do this, now?

`if (dynamicRules == null) return sheet.classes`

HenriBeck: Ohh yeah, if there are no dynamic rules, we can safely return the sheet classes

PR #1052, opened 2019-03-07
Correct npm Links

TatiTheFreaK:
The links to the npm website were wrong (they had packages instead of package)


kof:
Thank you!

Issue #1062, opened 2019-03-15
[react-jss] No critical css generated on server


bungu:
__Expected behavior:__
```javascript
const sheets = new SheetsRegistry();
const html = ReactDOM.renderToString(
  <JssProvider registry={sheets}>
    <App />
  </JssProvider>
);
const css = sheets.toString();
```

`css` should not be empty.

__Describe the bug:__
It seems that SheetsRegistry return empty string in `toString` method because no stylesheet is attached. Example repo: https://github.com/bungu/jss-no-critical-css

It works in v9

__Versions (please complete the following information):__

- JSS: v10.0.0-alpha.13
- React-JSS: v10.0.0-alpha.6

```
$ npm list jss react-jss
jss-no-critical-css@0.1.0 ~/sandbox/jss-no-critical-css
└─┬ react-jss@10.0.0-alpha.6
  └── UNMET PEER DEPENDENCY jss@10.0.0-alpha.13

npm ERR! peer dep missing: jss@10.0.0-alpha.3, required by jss-preset-default@10.0.0-alpha.7
```

HenriBeck:
Please update to the latest react-jss alpha version as well.

dandrei:
Could this be an issue in `9.8.7` as well? Over night, my `sheetsRegistry.toString()` started returning an empty string in my Next.js app on Linux (but not on Windows).

kof:
Most likely 2 versions are used in parallel, so registration happens in one, but .toString() call with another.

oliviertassinari:
The fix should be about:
```diff
  /**
   * Attach renderable to the render tree.
   */
  attach(): this {
-   if (this.attached || !this.renderer) return this
+   if (this.attached) return this
+   this.attached = true 
+   if (!this.renderer) return this
    this.renderer.attach()
-   this.attached = true
    // Order is important, because we can't use insertRule API if style element is not attached.
    if (!this.deployed) this.deploy()
    return this
  }
```

HenriBeck:
> Could this be an issue in `9.8.7` as well? Over night, my `sheetsRegistry.toString()` started returning an empty string in my Next.js app on Linux (but not on Windows). Maybe it's a dependency?

No, unless you explicitly changed some version, SSR shouldn't just break overnight.

HenriBeck:
This was introduced in the latest alpha. If you need SSR, please downgrade, for now, I will prepare a fix and a regression test for it.

HenriBeck:
[...]

PR #1064, opened 2019-03-16
[jss] Fix SSR rendering

HenriBeck:
This fixes generating the styles when no `Renderer` is present.

__Corresponding issue (if exists):__ #1062


HenriBeck:
@kof After this fix, some tests inside the priority test suite are failing. Not sure why though.

kof:
fixed

Issue #1021, opened 2019-02-06
Basic minification for SSR


kof:
It could be quite expensive to do a complex minification for SSR on each request, so we should provide a  n option to at least remove whitespaces in toString methods.

kof:
On the second thought. Removing new lines is trivial and can be done in the end after .toString() even in user land. Removing all spaces which can be omited less so, because some spaces are required, for example `@media all`. To remove omiitable spaces, we would have to add logic to all toString methods. Also it won't be consistent if user used a string value, since we won't parse it.

TriPSs:
I'm using this currently in production and did not have any problems with it so far (+- 1 year):
```
export default str => (
  str && str.replace(/[ \n\r\t\f\xA0]+/g, function (spaces) {
    return spaces === '\t' ? '\t' : spaces.replace(/(^|\xA0+)[^\xA0]+/g, '$1 ')
  })
)
```

cdnbacon:
@kof Is there a reason the default serialization is roughly a pretty print instead of a minimal, correct representation?

What is the current approach optimizing for?

kof:
@cdnbacon its been done this way for reading in dev tools html view when styles are static (no function values)

kof:
This one was actually already released in https://github.com/cssinjs/jss/releases/tag/v10.8.0

PR #1066, opened 2019-03-18
Add minification for SSR

perevezencev:
#1021

### DIFF
--- a/packages/jss/benchmark/tests/bootstrap.js
+++ b/packages/jss/benchmark/tests/bootstrap.js
@@ -1,4 +1,4 @@
-import {create} from 'jss'
+import {create, SheetsRegistry} from 'jss'
 import global from 'jss-plugin-global'
 
 import styles from '../fixtures/bootstrap.json'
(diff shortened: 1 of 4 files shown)


kof:
I wish you would have discussed this first, before investing time, because can't merge it like this.

1. This minification has cost that we have to measure
2. We will not increase bundle size like this
3. We need a clear use case presented. For e.g. would just removing whitespaces be sufficient for 90% of use cases?

HenriBeck:
I think we can easily make this ourselves in the `toCSS` function by just not adding line breaks and removing the indentation.

kof:
@HenriBeck yes, this is what original issue is about. The package he is using here, will do many more optimmizations, for e.g. converting to shorthands etc.

perevezencev:
Yes, csso minify css completely, but this entails a number of degradations. 
I rewrote the bootstrap test to check how minification affects performance during runtime.
image
This result is not clear to me, I expected to see a decrease, so it is necessary to consider this in more detail.
The size of the package certainly increases, but `SheetsRegistry` is used only on the server, so it does not matter (correct me if I'm wrong).
As a summary, can add an option for complete minification and do basic minification (removing spaces) by default.

kof:
> The size of the package certainly increases, but SheetsRegistry is used only on the server, so it does not matter (correct me if I'm wrong).

1. It is intended for SSR, but we don't know for sure how it is used, since it can be used at runtime as well.
2. We export it from index, so it will automatically get into the bundle. 
3. I think if you really want csso optimizations you will have to use it in your user code. This is not hard.

kof:
> As a summary, can add an option for complete minification and do basic minification (removing spaces) by default.

Yes, I think this is what we can do in the core. The rest can be done outside. It's just calling `minify(extractedCSS)` and perfectly doable in userland.

HenriBeck:
Closing this PR as this kind of optimization should be done in the user's code, and we can do a faster minification in our code.

justsml:
Are the examples of this implemented in userland?

Can't seem to find it.

PR #1074, opened 2019-03-27
rewrites typescript section

cancerberoSgx:
* react-jss doesn't include typings, you need to install them
 * the type WithStyles doesn't exists. It seems that its name is WithSheet 
 * working example of component using classes and theme and a app entry point using ThemeProvider

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -574,35 +574,64 @@ const Button = withStyles(buttonStyles)(() => (
 
 ### Usage with TypeScript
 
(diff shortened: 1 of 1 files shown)


HenriBeck:
react-jss includes typings in version 10

HenriBeck:
What we could add is a small section about react-jss prior to version 10

klequis:
Could a note be added to the doc about needing version 10.0.0 for withStyles/WithStyles? My first thought was to dig into the typings to see why this didn't work. Took some time before I came across this issue.

Issue #1072, opened 2019-03-25
Turn off class name compression by default and allow to opt-in


kof:
Majority of users seem to value more the easer debuggability in production mode with readable class names than the download size optimization.

Should we turn it off by default and allow to opt-in, like [MUI](https://github.com/mui-org/material-ui/pull/15044) did?

HenriBeck:
We could do it, but we should then include a prop on the jss provider to automatically do this

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
Yes, react-jss provider and jss instance both need to have a new option for this.

HenriBeck:
I think we could do that, but I would rather have 2 different class name generators in the core instead of one being configured by an option

kof:
How would a user turn on compression?

HenriBeck:
In react-jss I would provide a prop, but if the user decides to use compression he can just pass the other class name generator

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
They would have to reimplement our generateId function by themselves, I fear this is too much for the most.

HenriBeck:
No, people who use jss directly would just need to import a different createGenerateClassName

kof:
So JSS would provide 2 generators, one with compression, another one without? I dunno...

kof:
Or do you mean option for compression would exist in createGenerateId and user can create a generator with that option and set it in JSS?

HenriBeck:
We could also add an options object to createGenerateClassName function which has a property called minify

kof:
in that case what's the difference to having that option in Jss options?

HenriBeck:
```js
const jss = create({
  classNameOptions: { minify: true },
});
```

```js
const createGenerateClassName = (options: { minify: boolean }) => {
  // code
};
```

This way, MUI would also only need one class name generator and could just use the options

kof:
yeah, thats the original thought and an equivalent option can be provided through JssProvider as a prop. So, back to the origins.

kof:
Since we renamed the generator to id, I would call it `{idOptions: {minify: true}` or even `{id: {minify: true}` since we are kinda in options object, we can consider everything being an option semantically.

HenriBeck:
Merged in #1075

PR #1075, opened 2019-03-28
Added minify option to createGenerateId

HenriBeck:
This adds an option to the `createGenerateId` to allow to opt-in into minification of class names.

__Corresponding issue (if exists):__ #1072

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 
 - [jss-plugin-expand] Fix attributes spread for `border-bottom`, `border-top`, `border-left` and `border-right` ([#1083](https://github.com/cssinjs/jss/pull/1083))
 - [jss-plugin-props-sort] Fix sorting in Node 11 ([#1084](https://github.com/cssinjs/jss/pull/1083))
+- [jss] Fix escaping keyframes names ([#1100](https://github.com/cssinjs/jss/pull/1100))
 
 ## 10.0.0-alpha.16 (2019-3-24)
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+## Next
+
+### Improvements
+
+- [jss] Add option for opt-in minification of class names. ([#1075](https://github.com/cssinjs/jss/pull/1075))
+
 ## 10.0.0-alpha.16 (2019-3-24)
 
 ### Bug fixes
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,11 @@
 
 - [jss] Add option for opt-in minification of class names. ([#1075](https://github.com/cssinjs/jss/pull/1075))
 
+### Bug fixes
+
+- [jss-plugin-expand] Fix attributes spread for `border-bottom`, `border-top`, `border-left` and `border-right` ([#1083](https://github.com/cssinjs/jss/pull/1083))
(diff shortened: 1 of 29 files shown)


packages/react-jss/src/JssProvider.js
```diff
@@ -1,21 +1,28 @@
...
 
-type Props = {
-  jss?: Jss,
-  registry?: SheetsRegistry,
-  generateId?: GenerateId,
-  classNamePrefix?: string,
-  disableStylesGeneration?: boolean,
-  media?: string,
-  children: Node
+interface Props {
```
kof: why do you use in this case and in case of CreateGenerateIdOptions an interface instead of type?

packages/react-jss/src/JssProvider.js
```diff
@@ -26,9 +33,12 @@ export default class JssProvider extends Component<Props> {
     classNamePrefix: PropTypes.string,
     disableStylesGeneration: PropTypes.bool,
     children: PropTypes.node.isRequired,
-    media: PropTypes.string
+    media: PropTypes.string,
+    idOptions: PropTypes.shape({minify: PropTypes.bool})
```
kof: mb we should stick with one name in both jss and react-jss, because this kind of mapping often leads to confusion

packages/react-jss/src/JssProvider.js
HenriBeck: I'm just being used to use interfaces because of TS

kof: I think even in ts interfaces have a bit different purpose than types., we should use type when possible and interface when needed.

packages/react-jss/src/JssProvider.js
kof: so let's do idOptions in both cases?

HenriBeck: I would instead use `id` to not have duplicate `options` inside the jss options

changelog.md
```diff
@@ -1,3 +1,9 @@
+## Next
+
+### Improvements
+
+- [jss] Add option for opt-in minification of class names. ([#1075](https://github.com/cssinjs/jss/pull/1075))
```
kof: I think it's a feature and a breaking change

packages/react-jss/src/JssProvider.js
kof: alright, I was not sure if this will be somehow ambiguous but I like short version more too

kof:
we also need to update the docs

kof:
Just noticed we still didn't merge this one

HenriBeck:
I will finish this tomorrow evening

packages/jss/src/utils/escape.js
```diff
@@ -1,11 +1,4 @@
...
-export default str => {
-  // We don't need to escape it in production, because we are not using user's
-  // input for selectors, we are generating a valid selector.
-  if (process.env.NODE_ENV === 'production') return str
-
-  const nativeEscape = typeof CSS !== 'undefined' && CSS.escape
-
-  return nativeEscape ? nativeEscape(str) : str.replace(escapeRegex, '\\$1')
-}
+export default str => (nativeEscape ? nativeEscape(str) : str.replace(escapeRegex, '\\$1'))
```
HenriBeck: We will always need to escape the string now because the user might not minify the string

We could also pass here the sheet options and check if the user wants to minify the selector

PR #1077, opened 2019-04-03
Fix typspec for default unit plugin

yordis:
It seems that this plugin does not require me to pass the options.

https://github.com/cssinjs/jss/blob/27e2428297fa227d72d45d5757a74169edf70850/packages/jss-plugin-default-unit/src/index.js#L63

But the typespec is forcing me to pass a value.


kof:
thanks

yordis:
@kof I was questioning myself if you were a bot, that was quick 😆

kof:
not a bot, but a bot would say the same

PR #1078, opened 2019-04-03
Update react-jss.md

klequis:
Add a mention to the documentation that using WithStyles requires version 10 or higher. Seems a few people like me didn't notice the doc was for version 10.


HenriBeck:
Thanks 👍

PR #1085, opened 2019-04-09
Fix #1084

### LINKED ISSUE #1084: Css properties order are different between nodejs versions
__Expected behavior:__
I've set up snapshot-testing for CSS that is generated by cssinjs. When there is no change made in the source, I expect to have the same css-snapshot output on different nodejs environments.

__Describe the bug:__
Sadly the snapshots do differ on different nodejs version. The order of the properties in a stylesheet are different between node 10 and node 11.

__Codesandbox link:__ 
I don't know how to reproduce it as a codesandbox. But I have created a project on github. The project has one simple react component and a snapshot test for the component. The snapshot is also committed and is made with nodejs 10. Then I created two Pull-Request one with node9 and one with node11. You can see that nodejs 11 has a failed snapshot-test with some properties on different positions compared with the snapshot of node 10.

[Snapshot testcase on GitHub](https://github.com/justbrody/jss-snapshot-problem)

 - jss: react-jss@8.6.1
 - Browser [e.g. chrome, safari]:Not in the browser, testing in node
 - OS [e.g. Windows, macOS]:Linux/Windows/iMac

HenriBeck:
When the props have the same length while sorting in jss-plugin-props-sort, we check if the first prop is bigger than the second one and return 1 then. That way we should have a consistent output of orders even across different node versions

__Corresponding issue (if exists):__ #1084

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,6 +3,7 @@
 ### Bug fixes
 
 - [jss-plugin-expand] Fix attributes spread for `border-bottom`, `border-top`, `border-left` and `border-right` ([#1083](https://github.com/cssinjs/jss/pull/1083))
+- [jss-plugin-props-sort] Fix sorting in Node 11 ([#1084](https://github.com/cssinjs/jss/pull/1083))
 
 ## 10.0.0-alpha.16 (2019-3-24)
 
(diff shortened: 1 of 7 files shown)


HenriBeck:
I will validate this fix tomorrow morning

HenriBeck:
@kof, please review

Issue #907, opened 2018-10-27
[react-jss] Investigate on how to support the new React hooks


HenriBeck:
Supporting hooks when React 16.7 is officially being released might be good as they are cleaner than a HoC, though we should find a solution where we can share most of the code between a HoC and a Hook.

I created a prototype on how it might be implemented. Note this prototype isn't stable and there are quite a few things still missing from it. No guarantee that it works. I only tested it with a relatively small personal project.

 https://github.com/cssinjs/react-jss-hook

kof:
here are also some ideas https://gist.github.com/kof/43ac135686c71c536deb0db61b94e6e3

HenriBeck:
We will create a prototype here: https://github.com/cssinjs/react-jss-hook and merge this into `react-jss` once the hook is stable enough and when they are released in 16.7.

andywer:
Cool that you are already working on the hook prototype!

Had a quick look at the source code. Would like to share some early feedback if you don't mind:

Are you sure about the `createUseStyles()` factory?

Having a provider component to configure the hook via context seems like the cleaner approach, esp. users need to have one for SSR anyway already.

HenriBeck:
Yes, we need the factory to insert the stylesheets in the correct order properly.

kof:
For the context: if you define insertion order by render, it will be depending on when react renders a component, which might be different for the same component depending on the state and leads to specificity issues.

If we use a function call in the root of the module scope to set the order, we always rely on when module has been imported, which is a more stable approach.

andywer:
[FYI] I know you have given the whole injection order issue a lot of thought, but I couldn't help, but keep thinking that there might be another way.

Styled-components, for instance, they seemed to conclude in its early days that injection order based on time of first rendering of a component instance works well enough for them, as long as you don't use tag selectors: <https://github.com/styled-components/styled-components/issues/1>

I understand the appeal of an injection-order-perfect solution, but I wanted to see the other side, too. So I wrote that to see how well it works and "feels like": <https://github.com/andywer/react-usestyles>

Short summary: The styling hook that doesn't need to be instantiated works quite well, uses the same approach as styled-components, but suffers from the same edge-casy limitations, of course. Here is a [demo in combination with material-ui](https://codesandbox.io/s/myo47mvjw9).

Main benefit is that I was able to abstract the CSS-in-JS implementation away. Just wanted to share the results of my tests. Esp. abstracting the styling library away (see its readme) could be an interesting path to persue :)

[...]

PR #1089, opened 2019-04-13
[react-jss] Hooks

HenriBeck:
Hooks support 🎉 🎉 

__Corresponding issue (if exists):__ #907


packages/react-jss/src/hook/index.js
```diff
@@ -0,0 +1,97 @@
+// @flow
+
+import React, {useEffect, useLayoutEffect} from 'react'
+import {ThemeContext as DefaultThemeContext} from 'theming'
```
TrySound: Why not just hook.js?

packages/react-jss/src/JssContext.js
```diff
@@ -2,9 +2,7 @@
 import React, {type Context} from 'react'
 import type {Context as JssContextValue} from './types'
 
-const JssContext: Context<JssContextValue> = React.createContext({
-  sheetOptions: {},
+export const JssContext: Context<JssContextValue> = React.createContext({
```
kof: Why do we change the default export to a named one?

packages/react-jss/src/JssProvider.js
```diff
@@ -18,7 +18,7 @@ type Props = {
   children: Node
 }
 
-export default class JssProvider extends Component<Props> {
```
kof: same question here ... the naming convention for file names was this:

If a default export is a class use Capital case file name.  Also use a default export if it's the only export.

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
+// @flow
```
kof: File name convention so far was camel case.

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+    const theme: Theme = typeof styles === 'function' ? React.useContext(ThemeContext) : noTheme
+
+    // When the theme or the context changes we create a new sheet
+    const sheet = React.useMemo(
+      () => {
+        if (jssContext.disableStylesGeneration) {
+          return undefined
+        }
+
+        const stylesheet = createStaticSheet({
```
kof: we could call it `sheet`

kof: ah it's the linter shadow linter issue, I would disable the rule locally here, because it's exactly the same object and it is confusing to have different names for it

kof:
CC @oliviertassinari @eps1lon wanna review this one as well? Are there use cases that mui hooks interface support and this implementation isn't? I would love to align them

packages/react-jss/src/withStyles.js
```diff
@@ -1,316 +0,0 @@
-// @flow
```
kof: I am wondering if we should keep both interfaces until v11, it might make v10 adoption very slow, we should give people time to migrate to hooks. Especially considered React official statement is not to migrate classes to hooks, just for the sake of hooks, so people might be hesitative rewriting their entire code base and we end up everyone using v9.

I would not support  any maintenance of v9 though, because they can start using hooks API any time, it seems fully interoperable with a HOC and we can announce the HOC removal somewhere next year.

WDYT?

packages/react-jss/src/with-styles.js
```diff
@@ -0,0 +1,168 @@
+// @flow
```
kof: Just realized you only renamed withStyles, not removed it. Please avoid making  changes which could be done in a separate step when needed, it makes it harder to review and discussion gets bloated with unrelated stuff. I would prefer to keep current conventions.

kof: For the sake of consistency with the rest.

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+  const {index = getIndex(), theming, name = 'Hook', ...sheetOptions} = options
+  const ThemeContext = (theming && theming.context) || DefaultThemeContext
+
+  return (data: any) => {
+    const jssContext = React.useContext(JssContext)
+    // $FlowFixMe
+    const theme: Theme = typeof styles === 'function' ? React.useContext(ThemeContext) : noTheme
+
+    // When the theme or the context changes we create a new sheet
+    const sheet = React.useMemo(
```
oliviertassinari: React doesn't provide any guarantee about the underlying caching of useMemo. The cache might be clear at any time. You probably don't want to use it for anything involving `manager.add(options.theme, sheet)`. It "might" leak.

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+
+        updateDynamicRules(data, sheet, rules)
+
+        return rules
+      },
+      [sheet]
+    )
+
+    // Update the dynamic rules before the actual render if the data has changed
+    React.useLayoutEffect(
```
oliviertassinari: It needs to be changed for SSR.

packages/react-jss/src/create-use-styles.js
oliviertassinari: Also, the sheet attach should be performed in an effect. React can trigger a render to later ditch the work started. If the render has inserted a sheet into the page, no teardown will be triggered, the sheet will never be removed. It leaks. That's 💯 guarantee, we have this problem on MUI side. We didn't refactor the code yet.

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+      [data]
+    )
+
+    // Remove the old sheet when the sheet has changed after the render
+    React.useEffect(
+      () => () => {
+        removeDynamicRules(sheet, dynamicRules)
+
+        unmanageSheet({
+          context: jssContext,
```
kof: we could avoid mapping here

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+  updateDynamicRules
+} from './utils/sheets'
+import {getIndex} from './utils/index-counter'
+import type {HookOptions, Styles} from './types'
+import {unmanageSheet, manageSheet} from './utils/managers'
+import {getSheetClasses} from './utils/get-sheet-classes'
+
+const noTheme = {}
+
+function createUseStyles<Theme: {}>(styles: Styles<Theme>, options?: HookOptions<Theme> = {}) {
```
oliviertassinari: `makeStyles` would be shorter?

packages/react-jss/src/utils/sheets-meta.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+
+import type {StyleSheet} from 'jss'
+import type {StaticStyles} from '../types'
+
+interface SheetMeta {
```
kof: I would use an exact type here `type SheetMeta {| ... |}`

packages/react-jss/src/utils/sheets-meta.js
```diff
@@ -0,0 +1,22 @@
...
+
+function getMetaForSheet(sheet: StyleSheet) {
+  return sheetsMeta.get(sheet)
+}
+
+function addMetaForSheet(sheet: StyleSheet, meta: SheetMeta) {
+  sheetsMeta.set(sheet, meta)
+}
+
+export {getMetaForSheet, addMetaForSheet}
```
kof: if you use  named functions you could as well export them in place. Our convention though still is to not use named functions at all. It avoids a number of issues.

packages/react-jss/src/utils/sheets.js
```diff
@@ -0,0 +1,124 @@
...
+function createStaticSheet<Theme>(options: Options<Theme>) {
+  const manager = getManager(options.context, options.index)
+  const existingSheet = manager.get(options.theme)
+
+  if (existingSheet) {
+    return existingSheet
+  }
+
+  const jss = options.context.jss || defaultJss
+  const styles = getStyles(options)
```
kof: `staticStyles`? for consistency with dynamicStyles

packages/react-jss/src/utils/sheets-meta.js
```diff
@@ -0,0 +1,22 @@
...
+  dynamicRuleCounter: number;
+}
+
+const sheetsMeta = new WeakMap<StyleSheet, SheetMeta>()
+
+function getMetaForSheet(sheet: StyleSheet) {
+  return sheetsMeta.get(sheet)
+}
+
+function addMetaForSheet(sheet: StyleSheet, meta: SheetMeta) {
```
kof: `setMeta` and `getMeta` makes more sense to me

packages/react-jss/src/utils/sheets-meta.js
```diff
@@ -0,0 +1,22 @@
...
+
+interface SheetMeta {
+  styles: StaticStyles;
+  dynamicStyles: StaticStyles;
+  dynamicRuleCounter: number;
+}
+
+const sheetsMeta = new WeakMap<StyleSheet, SheetMeta>()
+
+function getMetaForSheet(sheet: StyleSheet) {
```
kof: I would add return types always

packages/react-jss/src/utils/sheets.js
```diff
@@ -0,0 +1,124 @@
...
+      options.name
+    } />'s styles function doesn't rely on the "theme" argument. We recommend declaring styles as an object instead.`
+  )
+
+  return styles(options.theme)
+}
+
+function getSheetOptions<Theme>(options: Options<Theme>, link: boolean) {
+  const classNamePrefix =
+    process.env.NODE_ENV === 'production' ? '' : `${options.name.replace(/\s/g, '-')}-`
```
kof: Since we started using a non-minified version in the other PR by default, it would make sense to do the same here, no?

packages/react-jss/src/utils/sheets.js
```diff
@@ -0,0 +1,124 @@
...
+    styles,
+    dynamicRuleCounter: 0
+  })
+
+  manager.add(options.theme, sheet)
+
+  return sheet
+}
+
+const removeDynamicRules = (sheet: ?StyleSheet, rules: ?DynamicRules) => {
```
kof: can StyleSheet be `null`? I would try to avoid maybe types as much as possible.

packages/react-jss/src/create-use-styles.js
eps1lon: I like this name more TBH. It makes it obvious that a hook is returned. ´makeStyles(styles)` doesn't make much sense. Why should a function make X if I already have X and pass it into `make`?

packages/react-jss/src/JssContext.js
HenriBeck: Fixed

packages/react-jss/src/create-use-styles.js
HenriBeck: Fixed

packages/react-jss/src/JssProvider.js
HenriBeck: Fixed

packages/react-jss/src/withStyles.js
HenriBeck: Yeah, that's why I also now exported the `withStyles` as a named export. I think we should switch our docs to use the named export and deprecate the default export at some time later this year.

packages/react-jss/src/with-styles.js
HenriBeck: Done

packages/react-jss/src/create-use-styles.js
HenriBeck: @eps1lon I thought the same.

I also thought about the following syntax:

```js
const {useStyles, withStyles} = makeStyles(styles);
```

You would get hoc with that as well, but this would make it a bigger change once `withStyles` is potentially completely removed.

packages/react-jss/src/withStyles.js
HenriBeck: I don't think we can remove `withStyles` any time soon, because unless react deprecates classes, we should support the API

packages/react-jss/src/utils/sheets-meta.js
HenriBeck: Done

packages/react-jss/src/utils/sheets-meta.js
HenriBeck: Done.

packages/react-jss/src/utils/sheets-meta.js
HenriBeck: Flow will infer them correctly. I don't think we necessarily need them.

packages/react-jss/src/utils/sheets.js
HenriBeck: Yeah, but the change shouldn't be in this PR.

packages/react-jss/src/utils/sheets.js
HenriBeck: Yes, it can be `undefined`

packages/react-jss/src/create-use-styles.js
```diff
@@ -0,0 +1,101 @@
...
+      },
+      [theme, jssContext]
+    )
+
+    // When the sheet changes, we readd the dynamic rules and update them
+    const dynamicRules = React.useMemo(
+      () => {
+        const rules = addDynamicRules(sheet)
+
+        updateDynamicRules(data, sheet, rules)
```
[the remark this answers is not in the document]
HenriBeck: Yes, this will do some DOM manipulations

packages/react-jss/src/create-use-styles.js
HenriBeck: So we should use a combination of `useState` with an initial state function and `useLayoutEffect` to flush changes to the DOM before an update happens.

Actually, on SSR we only need to add it to the registry and not manage the sheet, so we could move the managing of the sheet into a `useLayoutEffect` where we could also clean it up. We need to make sure that there are no flashes of unstyled content then.

packages/react-jss/src/create-use-styles.js
HenriBeck: I'm not sure if we can get around that. I think we need `useLayoutEffect` to flush the changes before the actual render happens, so we don't cause an additional render.

I'm not sure why react warns in this case as this is a valid use case.

HenriBeck:
I have been playing around with this hooks implementation and it works really well.
HMR is fully supported and works perfectly.

I think the layout effect is the right choice here to update the dynamic rules. We are actually not depending on it and updating the dynamic rules twice on the first render on the client. What we could do is check if we are in a server environment and then just call a function which will not be executed.

```js
const createUseStyles = () => {
  const useLayoutEffect = isInBrowser ? React.useLayoutEffect : (fn) => fn;

  return () => {
    useLayoutEffect(() => {
      updateDynamicRules();
    });
  };
};
```

packages/react-jss/src/create-use-styles.js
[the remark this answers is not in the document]
HenriBeck: I think we can remove it because we are currently updating the dynamic rules twice on the first render.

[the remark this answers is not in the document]
kof: we should not need to update rules twice, right?

kof:
> I have been playing around with this hooks implementation and it works really well.
HMR is fully supported and works perfectly.

Nice, let's update the HRM guide in this case and add it as a feature in the changelog

kof:
>  What we could do is check if we are in a server environment and then just call a function which will not be executed.

Sure, though I don't quite understand why we need it since useLayoutEffect is not running during SSR anyways.

packages/react-jss/src/create-use-styles.js
kof: yeah, makeStyles semantically isn't very accurate

kof: things which make sense semantically create lower mental overhead, because you don't have to remember what they do

packages/react-jss/src/create-use-styles.js
kof: useLayoutEffect will not cause a FOUC

packages/react-jss/src/create-use-styles.js
kof: > You would get hoc with that as well, 

I would try to clearly separate the HOC interface, because I think HOC interface is going away completely at some point.

packages/react-jss/src/create-use-styles.js
eps1lon: React doesn't know that the server rendered output will include the styling so they assume this causes flashes. 

I nasty workaround is to use something like `const useEnhancedEffect = onServer ? useEffect : useLayoutEffect` which gets rid of the warning but looses the linter.

packages/react-jss/src/createUseStyles.js
```diff
@@ -0,0 +1,106 @@
...
+    const theme = useTheme()
+
+    // When the theme or the context changes we create a new sheet
+    const sheet = React.useMemo(
+      () => {
+        if (context.disableStylesGeneration) {
+          return undefined
+        }
+
+        const staticSheet = createStaticSheet({
```
eps1lon: To reiterate on the render vs. commit phase leak: This can potentially be called multiple times after/before effect cleanups are executed.

It should ensure that unmanage and manage happens during render and only DOM manipulations happen in useLayoutEffect. It seems like sheet managing is coupled to the DOM currently in Jss?

packages/react-jss/src/create-use-styles.js
kof: This workaround seems to fit accurately to what's going on, linter can be disabled for that line specifically.

packages/react-jss/src/createUseStyles.js
kof: Sheet manager https://github.com/cssinjs/jss/blob/master/packages/jss/src/SheetsManager.js
has 2 functions
1. count the refs - the amount of component instances using a sheet
2. calling attach/detach once first component is using a sheet or none is using

Since .attach/.detach bascially calls the DomRenderer functions which do nothing if we are not on the client, we don't have a strong coupling to the DOM.

eps1lon: > Since .attach/.detach bascially calls the DomRenderer functions

If there is no manage/unmanage functionality that does not also use the DOM then it is tightly coupled. Just because something is a no-op in some environment doesn't mean it doesn't create a coupling. 

If you want to be concurrent ready then this functionality needs to be split up. There will be a lot of issues in concurrent react if you have side-effects in render that are not cleaned up in the same phase.

kof: cleaning up can be done if we need it, but I don't know for sure the exact requirements there, so why bother now? It seems like something we can solve later, no?

eps1lon: Maybe. I think pre-release would be a good chance to get this right since we already know that |render| >= |commit| and we have ways to test that our solutions work with that assumptions (stable: StrictMode, unstable: unstable_ConcurrentMode).

I'm currently creating some sandboxes that compare different styling solutions in different modes. I would hope that a styling solution using hooks does not introduce regressions.

HenriBeck: As we have a lot of confusion around what needs to happen, here is a breakdown:

On SSR:
Create the sheet and add it to the registry. We would also need to add the dynamic rules and update them.

On the client:
Create sheet, add the dynamic rules and update them. Here we also need to call so `manage` the sheets which will attach them to the DOM.

We are currently adding the sheet to the registry in the managing phase. So by splitting them up, we could solve the SSR problem very quickly.

My question would be: What happens if we call `setState` inside a layout effect? Will the state update be reflected in the first render?
Because on the client we also need to think about theme updates aka recreating the sheet. I would assume this should go into the a layout effect, but this would mean we would also need to move the initial creation to a layout effect

kof: > Create the sheet and add it to the registry. We would also need to add the dynamic rules and update them.

Are you sure we need this on SSR? It is supposed to be a single pass, rules should only need the initial state, without component updates.

kof: > Create sheet, add the dynamic rules and update them. Here we also need to call so manage the sheets which will attach them to the DOM.

Currently SheetsRegistry.toString() only returns attached nodes AFAIK. That's why we also call a .attach() during SSR. This basically marks sheets that we need to output during SSR, because there will be more created than mounted. Seems like we need a change in this logic.

kof: > My question would be: What happens if we call setState inside a layout effect? Will the state update be reflected in the first render?

Please verify this, but afaik  - no, SSR is a single pass.

kof:
Btw we should move the Manager to react-jss, it was in the core for sharing purposes between the react integrations.

packages/react-jss/src/createUseStyles.js
```diff
@@ -0,0 +1,106 @@
...
+        (): Theme => React.useContext(ThemeContext)
+      : // $FlowFixMe
+        (): Theme => noTheme
+
+  return (data: any) => {
+    const context = React.useContext(JssContext)
+    const theme = useTheme()
+
+    // When the theme or the context changes we create a new sheet
+    const sheet = React.useMemo(
```
kof: Do I understand correctly that we rely on memo to reuse the static sheet between component instances?

kof:
Also I think we need to change the logic in SheetsRegistry. It uses `attached` property to decide what gets returned from toString(), though there is no DOM attach on server, so it is semantically not very accurate, like @eps1lon said, we need to separate this into 2 stages:

1. Execution/Init - component function was executed, no matter SSR or client
2. Render/Mount - component is about to render to DOM - only client.

Our manager with methods manage/unmanage exists because we are reusing static style sheets between component instances and so we need to keep count of instances to know when to attach and when to detach.
During SSR we don't need this, because every component that was executed/initialized yelds html to renderToString, so SheetsRegistry should return that CSS from toString().


What I don't quite understand is why I added [this check](https://github.com/cssinjs/jss/blob/master/packages/jss/src/SheetsRegistry.js#L61) back then. I know I had a case for sure, is it possible that class based components were constructed without ever mounting, so that you could have CSS that is not needed for SSR?

HenriBeck:
@kof @oliviertassinari @eps1lon I update the hook to use states and also fixed the issue with the layout effect

packages/react-jss/src/createUseStyles.js
```diff
@@ -27,79 +26,60 @@ const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions
       : // $FlowFixMe
         (): Theme => noTheme
 
+  const useLayoutEffect = typeof window === 'undefined' ? React.useEffect : React.useLayoutEffect
```
kof: we depend on https://github.com/cssinjs/jss/blob/master/packages/jss/package.json#L44 anyways

HenriBeck:
The only question is when do we need to recreate the sheets when the theme changes for example. My instinct would say we should use `useLayoutEffect` but we would want to ignore the first render as we are already doing it inside the state initializers

kof:
> The only question is when do we need to recreate the sheets when the theme changes for example.

Everything that renders on the clinent to CSSOM should use `useLayoutEffect`, it's the phase that is run by react right before they insert dom nodes and they would run it only if they didn't suspend the lifecycles, so if you do it anywhere else, there is a risk of doing the work while it could be delayed.

HenriBeck:
I added support for recreating the sheet when the theme changes.

There is still a nasty flow issue inside the `theming` package which comes from upgrading `flow` which I can't seem to get ignored by flow.

kof:
where is the flow issue?

HenriBeck:
Inside the `node_modules/theming`

kof:
yes but where do I need to test/look here?

HenriBeck:
@kof just pushed a branch to `theming` which upgrades flow there. There are still eight issues left which I wasn't able to resolve yesterday evening for three hours.

.eslintrc.js
```diff
@@ -10,6 +10,10 @@ module.exports = {
     __VERSION__: true,
     CSS: true
   },
+  rules: {
+    'import/prefer-default-export': 'off',
+    'import/no-named-as-default': 'off'
```
kof: why not turn of the rule locally where needed?

kof: Ideally we should not have to configure eslintrc anywhere except of one central place for cssinjs org

packages/react-jss/src/createUseStyles.js
```diff
@@ -0,0 +1,148 @@
...
+      : // $FlowFixMe
+        (): Theme => noTheme
+
+  const useLayoutEffect = isInBrowser ? React.useLayoutEffect : React.useEffect
+
+  return (data: any) => {
+    const context = React.useContext(JssContext)
+    const theme = useTheme()
+
+    if (context.disableStylesGeneration) {
```
eps1lon: Is there some guard in the provider that prevents this value from changing? Otherwise we get an error from react about changing number of hooks calls.

packages/react-jss/src/createUseStyles.js
```diff
@@ -0,0 +1,148 @@
...
+            theme,
+            index,
+            sheetOptions
+          })
+          const newDynamicRules = staticSheet ? addDynamicRules(staticSheet, data) : undefined
+          const newClasses = staticSheet ? getSheetClasses(staticSheet, dynamicRules) : undefined
+
+          setStaticSheet(newStaticSheet)
+          setDynamicRules(newDynamicRules)
+          setClasses(newClasses)
```
eps1lon: Are these batched or can each call trigger a separate render? If so can this be an issue? It looks like those should live in a single state i.e. use `useReducer`.

.eslintrc.js
kof: will fix it since it's quicly done

packages/react-jss/src/createUseStyles.js
HenriBeck: We implemented it because someone wanted it to use it in a way where it's not changed. Though there is no rule against changing this. I will move it further down

packages/react-jss/src/createUseStyles.js
HenriBeck: The ` setState` calls are batched from react AFAIK. But we could also change whole implementation to a `useReducer` call.

kof: setState is batched since quite some time in react, you can call any amount of setState, the state is getting merged and scheduled

damianobarbati:
Is this making it into next alpha release so we can start playing with it? :)

kof:
yes

packages/react-jss/src/createUseStyles.js
eps1lon: It's batched in event handlers but not in general (which is why `unstable_batchedUpdates` exists). Don't know about effects though. Does anybody have a resource comparing setState behavior in render, layout effects, passive effects and event handlers?

https://github.com/facebook/react/issues/10231#issuecomment-316644950 makes me question that they are indeed batched. Link is old though and may very well not apply anymore.

I would expect react to batch anything that it invokes itself (event handlers, renders, effects etc). So the code would be fine.

kof:
@HenriBeck I merged it and I am going to release 10.0.0-alpha.17 ... to let everyone try this version ... we still have some new bugs in the issues and dynamic updates benchmark from react-native-web doesn't work.

Also we need to try to reduce the amount hooks we use, because I think they cost us CPU time and we . are loosing perf benefits against HOC interface.

kof:
releasing it despite failing tests,  I think it is again something with browserstack,  if not we have to fix it in the next release

kof:
v10.0.0-alpha.17 is out

Issue #974, opened 2019-01-07
Make sure styled-system is supported


kof:
Styled-System by @jxnblk https://github.com/jxnblk/styled-system seems like a  very solid approach for building design systems. I think we should look into it and see how it can be supported. 

## Todo

- [ ] See how it can be supported 
- [ ] Add it to documentation with links
- [ ] Make codesandbox examples
- [ ] Send a PR to styled-system and link the examples

jxnblk:
Haven't really tested it recently, but styled-system *should* "just work" with `styled-jss`. Definitely let me know if you run into any bugs though

kof:
I wonder if it also can work with react-jss and the future version of it based on hooks (prototype https://github.com/cssinjs/react-jss-hook)

HenriBeck:
It should already work with react-jss

Eddort:
this only works with react-jss @ alpha-10
but if you want to use something like `<Box p = {[1, 1/2, 1/4]}> </ Box>` you will notice a bug [#1044](https://github.com/cssinjs/jss/issues/1044)

kof:
Putting high prio on that bug.

HenriBeck:
I think the issue is that `styled-system` is accessing the theme via props and not from our `themed styles` function.

kevinSuttle:
Would love to see this also.

kof:
@jxnblk do you see a good way to support styled-system without enforcing use of components, but with functions instead?

kevinSuttle:
What is the relationship between https://material-ui.com/system/basics/ and JSS here?

kof:
MUI style system is an independent effort by @oliviertassinari inspired by styled-system. It is a separate package https://github.com/mui-org/material-ui/tree/next/packages/material-ui-system and seems to have no direct dependency on JSS, but can be used together.

I am also porting an SC-like, similar to styled-jss into the react-jss package to support styled-system and SC-pattern users.

kevinSuttle:
See also: https://github.com/emotion-js/emotion/tree/master/packages/primitives-core

kof:
@kevinSuttle what specific use case from primitive-core are you interested in?

kevinSuttle:
Nothing in particular. It's just another approach—one that isn't included in the Prior Art section here: https://material-ui.com/system/basics/#prior-art

Also odd that `styled-system` isn't mentioned here.

kof:
it is mentioned there)

[...]

PR #1094, opened 2019-04-23
Styled API

kof:
Implements #974

## Todo
- [x] use the new hooks based api
- [x] Unsupported rule rendering bug https://github.com/cssinjs/jss/issues/1122
- [x] styled accepts multiple style objects or object and functions `styled('div')({}, () => ({}) ...)`
- [x] shouldForwardProp(propName) as options property `styled('div', {shouldForwardProp})(style...)`
- [x] Add emotion's `label` 
- [x] Changelog


packages/jss/tests/unit/createGenerateId.js
```diff
@@ -2,7 +2,7 @@
```
kof: changes here are unrelated, but for some reason after I added tests for styled api, this tests started to fail, something weird started happening with module id, like there is a new moduleId module evaluation, I couldn't fully understand that, but I just started resetinng moduleId here.

packages/react-jss/src/styled.js
```diff
@@ -0,0 +1,47 @@
...
+        if (isTag && !isPropValid(prop)) continue
+        childProps[prop] = props[prop]
+      }
+      // $FlowFixMe flow seems to not know that `classes` will be provided by the HOC, not by element creator.
+      childProps.className = className ? `${className} ${classes.css}` : classes.css
+      return React.createElement(as || type, childProps)
+    }
+
+    // $FlowFixMe flow seems to not know that `classes` will be provided by the HOC, not by element creator.
+    return withStyles({css: style}, {...options, injectTheme: true})(Styled)
```
HenriBeck: This implementation is good for now, but I would like to optimize it, later on, to not render two extra react components for example.

I'm not sure if we should use the hooks implementation in here yet, as this would make the `styled` API only usable with `react@16.8` currently, which in my opinion is a little bit early.

kof: I actually think performance is critical and this amount of components for each styled one is really bad, that's why I think hooks api is the only way forward with this API. For those who use older react versions they still can use the old styled implementation.

packages/react-jss/src/types.js
```diff
@@ -22,20 +20,31 @@ export type Context = {|
...
 
-export type HOCProps<Theme, Props> = Props & {|
+export type HOCProps<Theme, Props> = Props & {
   theme?: Theme,
   jssContext: Context,
   innerRef: any
-|}
+}
+
+export type Classes = {[string]: string}
```
kof: we could export this one from the core

packages/react-jss/src/types.js
```diff
@@ -22,20 +20,31 @@ export type Context = {|
...
-  classes: {}
+  classes: Classes
 }
 
 export type DynamicRules = {
   [key: string]: BaseRule
 }
 
-export type ThemedStyles<Theme> = (theme: Theme) => StaticStyles
+type StaticStyle = {}
```
kof: we should use this one from the core too, for the sake of a sigle source of truth

packages/react-jss/tests/styledSystem.js
```diff
@@ -0,0 +1,186 @@
...
+    const registry = new SheetsRegistry()
+    const Div = styled('div')(space)
+    const renderer = TestRenderer.create(
+      <JssProvider registry={registry} generateId={createGenerateId()}>
+        <ThemeProvider theme={theme}>
+          <Div px={1} />
+        </ThemeProvider>
+      </JssProvider>
+    )
+    // TODO we should not need a static rule in such cases.
```
kof: this really needs fixing

.flowconfig
```diff
@@ -6,4 +6,6 @@
 
 [options]
 include_warnings=true
-esproposal.export_star_as=enable
\ No newline at end of file
+esproposal.export_star_as=enable
+suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
+suppress_comment= \\(.\\|\n\\)*\\$FlowIgnore
```
kof: This  can be used for things which need to be ignored but need no fixing

packages/react-jss/src/createUseStyles.js
```diff
@@ -96,7 +96,8 @@ const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions
 
     useEffectOrLayoutEffect(
       () => {
-        if (state.sheet && state.dynamicRules) {
+        // We only need to update the rules on a subsequent update and not in the first mount
+        if (state.sheet && state.dynamicRules && !isFirstMount.current) {
```
kof: this can be in changelog

Issue #1098, opened 2019-05-08
@keyframes identifiers aren't escaped and don't work if contain, e.g., parentheses


synchronos-t:
__Expected behavior:__
@keyframes identifiers should work even if they contain injected parentheses and should escape them properly.

__Describe the bug:__
Material UI, from version 4.0.0-alpha.6 onwards, has been injecting HOCs to JSS class names with parentheses, e.g. `Connect(ComponentName)` (for react-redux connect()), instead of `Connect--ComponentName`. These work well with class names and the parentheses are properly escaped. However, when used with `@keyframes`, they appear in styles unescaped and hence stop working.

__Codesandbox link:__ 
https://codesandbox.io/s/74m7qv4qrj

The above Codesandbox creates two buttons, which should be spinning, but don't. If you remove the connect() from TestComponent, the left button starts spinning again.

The right button (TestComponent2) tries to directly put parentheses in the identifier name, which causes the identifier to drop everything from the opening parenthesis onwards in `@keyframes` as well. (I'm not sure if it's needed for JSS to allow direct referring to ids containing parentheses, however.)

The related issue at Material UI: https://github.com/mui-org/material-ui/issues/15604

HenriBeck:
I think we can just remove all non A-Z characters from the name of the component

kof:
we have a escape utility, why not use it?

HenriBeck:
Actually, the problem is that we are not escaping the `keyframes` selector. We are doing it for normal class names

PR #1100, opened 2019-05-09
Add escaping of keyframe names

HenriBeck:
__Corresponding issue (if exists):__ #1098

### DIFF
--- a/packages/jss/tests/integration/sheet.js
+++ b/packages/jss/tests/integration/sheet.js
@@ -497,7 +497,7 @@ describe('Integration: sheet', () => {
       `)
     })
 
-    it('should work with tso referenced keyframes', () => {
+    it('should work with multiple referenced keyframes', () => {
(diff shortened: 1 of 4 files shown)


HenriBeck:
@kof, please approve. Added a regression test for it.

packages/jss/tests/integration/sheet.js
```diff
@@ -529,6 +529,36 @@ describe('Integration: sheet', () => {
       `)
     })
 
+    it('should correctly escape the name', () => {
+      const localJss = create({
+        createGenerateId() {
+          return (rule, sheet) => `${sheet.options.classNamePrefix}-${rule.key}-id`
```
kof: can you move it to the createGenerateId we use everywhere in tests? Seems like this could be just used there without any issues, I mean without changing the class name in all assertions.

kof:
@HenriBeck please add a changelog

Issue #837, opened 2018-09-05
[jss-template] Nesting, etc?


trusktr:
__Is your feature request related to a problem? Please describe.__

Can we nest rules when using jss-template? If so, how?

__Describe the solution you'd like__

If not, this would be nice.

__Are you willing to implement it?__

No time to at the moment.

HenriBeck:
Currently, we don't support nesting rules inside template strings.

kof:
It might be interesting to enable a way for nesting because currently, you can't do the style block in a template string entirely if needs nesting:
```js
const styles = {
  btn: {
    float: 'left',
    '& span': `
      color: red;
    `
  }
}
```
I see 2 potential ways to enable nesting at low cost

1. Using a template property (easiest, less nince)

```js
const styles = {
  btn: {
    template: `
      float: left;
   `,
    '& span': `
      color: red;
    `
  }
}
```

2. Actual nesting support

```js
const styles = {
  btn: `
    float:  left;
    & span: {
      color: red;
    }
  `
}
```

Now the problem here is that we can't know when the inner block starts/ends without a full parser, because `{}` can be inside of `content` prop and background `url()`. What we could do though, we could enforce a specific constraint on that, which would allow us to detect it. 

- For the opening curly brace, the constraint would be - "it needs to be on the same line with &", if & is found and no { - throw an Error.
-  For the closing brace it would be - "it needs to be on a separate new line", so if opening brace was found before and not closed, we are in a nested block, if no closing brace on a new line was found - throw an error.

Performance should be almost not impacted in this case.

PR #1103, opened 2019-05-12
Implements nesting in template string parser

kof:
Implements #837

I think I found a way to implement it without degrading performance, LOC is still pretty small.

The problem it is solving: you can't express a nested rule with a template string. If you used a template string initially to describe the rule, as soon as you want to add nesting, you will have to rewrite it to objects.

```
const styles = {
  btn: `
    float:  left;
    & span: {
      color: red;
    }
  `
}
```

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -44,7 +44,7 @@ module.exports = config => {
       frameworks: ['benchmark'],
       // Using a fixed position for a file name, m.b. should use an args parser later.
       files: [process.argv[4] || 'packages/jss/benchmark/**/*.js'],
-      preprocessors: {'packages/jss/benchmark/**/*.js': ['webpack']},
+      preprocessors: {'packages/**/benchmark/**/*.js': ['webpack']},
       reporters: ['benchmark'],
       // Some tests are slow.
       browserNoActivityTimeout: 20000
--- a/packages/jss-plugin-template/benchmark/tests/parse.js
+++ b/packages/jss-plugin-template/benchmark/tests/parse.js
@@ -1,3 +1,4 @@
+import stylis from 'stylis'
 import parse from '../../src/parse'
 
 const css = `
@@ -37,8 +38,21 @@ const css = `
   font-variant: normal normal;
   border-spacing: 0px;
 `
+
+stylis.set({
+  global: false,
+  keyframe: false,
+  prefix: false,
+  compress: false,
+  semicolon: true
+})
+
 suite('Parse', () => {
   benchmark('parse()', () => {
     parse(css)
+  })
+
+  benchmark('stylis()', () => {
+    stylis('#id', css)
   })
 })
--- a/packages/jss-plugin-template/src/parse.js
+++ b/packages/jss-plugin-template/src/parse.js
@@ -1,7 +1,7 @@
 // @flow
 import warning from 'tiny-warning'
 
-const semiWithNl = /[;\n]|[}\n]/
+const semiWithNl = /\n/
 
 /**
  * Naive CSS parser.
@@ -28,7 +28,7 @@ export default (cssText: string): Object => {
           warning(false, `[JSS] Missing opening curly brace in "${decl}".`)
           continue
         }
-        nestedRuleProp = decl.substr(0, openCurlyIndex).trim()
+        nestedRuleProp = decl.substr(0, openCurlyIndex - 1)
         style[nestedRuleProp] = {}
         continue
       }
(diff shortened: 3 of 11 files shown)


kof:
Also cc @eps1lon

karma.conf.js
```diff
@@ -44,7 +44,7 @@ module.exports = config => {
       frameworks: ['benchmark'],
       // Using a fixed position for a file name, m.b. should use an args parser later.
       files: [process.argv[4] || 'packages/jss/benchmark/**/*.js'],
-      preprocessors: {'packages/jss/benchmark/**/*.js': ['webpack']},
+      preprocessors: {'packages/**/benchmark/**/*.js': ['webpack']},
```
kof: was broken

packages/jss-plugin-template/benchmark/tests/createSheet.js
```diff
@@ -4,7 +4,7 @@ import template from '../../src/index'
 import parse from '../../src/parse'
 
 const options = {Renderer: null}
-const jss = create(options).use(template())
+const jss = create(options).use(template({cache: false}))
```
kof: disabling cache here to have a realistic comparison without caching, I didn't document it, because I don't think user should need it, so for now it's a private one, just for tests

package.json
```diff
@@ -15,7 +15,7 @@
     "format:ci": "yarn format --list-different",
     "test": "karma start --single-run",
     "test:watch": "karma start",
-    "test:bench": "cross-env BENCHMARK=true karma start --single-run",
+    "test:bench": "cross-env NODE_ENV=production BENCHMARK=true karma start --single-run",
```
kof: removes dead from the bench~!

packages/jss-plugin-template/benchmark/tests/parse.js
```diff
@@ -37,8 +38,21 @@ const css = `
...
+  compress: false,
+  semicolon: true
+})
+
 suite('Parse', () => {
   benchmark('parse()', () => {
     parse(css)
   })
+
+  benchmark('stylis()', () => {
```
kof: using stylis parser just to see a comparable other value in perspective

packages/jss-plugin-template/src/parse.js
```diff
@@ -1,29 +1,81 @@
...
     if (!decl) continue
+
+    const ampIndex = decl.indexOf('&')
+
+    if (ampIndex !== -1) {
+      const openCurlyIndex = decl.indexOf('{')
+      if (openCurlyIndex === -1) {
+        warning(false, `[JSS] Missing opening curly brace in "${decl}".`)
+      }
+      const key = decl.substring(0, openCurlyIndex - 1)
```
HenriBeck: Should we trim the selector here?

packages/jss-plugin-template/src/parse.js
```diff
@@ -1,29 +1,81 @@
...
+      const nestedStyle = {}
+      rules[rules.length - 1][key] = nestedStyle
```
HenriBeck: Why do we add the key here to the last object?

packages/jss-plugin-template/src/parse.js
```diff
@@ -1,29 +1,81 @@
...
+      rules.push(nestedStyle)
+      continue
+    }
+
+    // We are closing a nested rule.
+    if (decl === '}') {
+      rules.pop()
```
HenriBeck: Shouldn't we add the rule when we pop it to the style?

packages/jss-plugin-template/src/index.js
```diff
@@ -2,13 +2,27 @@
...
+type Options = {|cache: boolean|}
+
+export default function templatePlugin(options: Options = {cache: true}): Plugin {
+  const onProcessStyle = style => {
+    if (typeof style !== 'string') {
+      return style
+    }
+
+    if (style in cache) {
+      return cache[style]
```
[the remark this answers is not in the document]
kof: the key is the template string, if template string is the same, why should the parsed object be different? What is the problem I am not seeing?

packages/jss-plugin-template/src/parse.js
kof: yes

packages/jss-plugin-template/src/index.js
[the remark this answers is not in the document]
HenriBeck: Nvm, I thought we are caching based on the keys and not the style 🤦‍♂

packages/jss-plugin-template/src/parse.js
kof: rules array represents an array of nested styles, each level, separate array entry, so we need to do both, reference the nested style in the parent object and add it to the array to have a pointer to the current style object where declarations will be added next in that pass

packages/jss-plugin-template/src/parse.js
kof: we already did, we are just removing it from the array, since we basically finished adding declarations to that rule

HenriBeck:
@kof should we merge this?

PR #1105, opened 2019-05-14
Export StyleSheet from jss to facilitate testing

timsnadden:
By exporting StyleSheet from the package it is now possible to test that functions are executed as expected. e.g. 
```
import { StyleSheet } from 'jss';
...
const attachSpy = jest.spyOn(StyleSheet.prototype, 'attach');
...
expect(attachSpy).toHaveBeenCalledTimes(3);
```

### DIFF
--- a/packages/jss/src/index.js
+++ b/packages/jss/src/index.js
@@ -7,7 +7,7 @@
  * @license MIT
  */
 import Jss from './Jss'
-import type StyleSheet from './StyleSheet'
+
 import type {
   ConditionalRule,
   KeyframesRule,
@@ -38,16 +38,7 @@ export type {
   ContainerRule
 } from './types'
 
-export type {
-  Jss,
-  StyleSheet,
-  ConditionalRule,
-  KeyframesRule,
-  StyleRule,
-  ViewportRule,
-  SimpleRule,
-  FontFaceRule
-}
+export type {Jss, ConditionalRule, KeyframesRule, StyleRule, ViewportRule, SimpleRule, FontFaceRule}
 
 /**
  * Export a constant indicating if this browser has CSSTOM support.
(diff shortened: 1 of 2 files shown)


packages/jss/src/index.js
```diff
@@ -7,7 +7,7 @@
  * @license MIT
  */
 import Jss from './Jss'
-import type StyleSheet from './StyleSheet'
```
timsnadden: I wasn't sure whether this could be removed or not. It uses the same identifier (StyleSheet) so if it's to be retained I'll need to work out how to do this. I've not worked with flow types before

HenriBeck:
Why do you need to test that the stylesheets are correctly attaching and detaching? For me, this is an implementation detail from `react-jss` and shouldn't be required to test from other libraries or application

timsnadden:
@HenriBeck The code I'm testing is not using `react-jss`. 

I agree about not testing the functionality of other libraries. In fact, this is why I wanted to be able to spy on `attach` and `detach` rather than trying to inspect the DOM. It's jss' job to ensure that calling `attach`, `detach`, `update` does what it says on the tin. My code is a HOC that is similar to `withStyle`. I'm wanting to ensure that the logic inside some `useEffect` hooks are correct around calls to methods of `SheetsRegistry` `SheetsManager` and `StyleSheet`. The first two are already exported but not `StyleSheet`. I hope that explains it. 

Is there some disadvantage to exporting `StyleSheet`? My only guess is that not exporting it is an indication that you don't want users to be interacting with it directly.

HenriBeck:
> Is there some disadvantage to exporting StyleSheet? My only guess is that not exporting it is an indication that you don't want users to be interacting with it directly.

Yeah, that's precisely the point. This would open up to people directly creating instances which we don't want.

Maybe check our tests and see how we make sure that the stylesheets are attached.
There are basically two ways:

When running in a real/simulated browser, you could check that the stylesheet was added to the DOM.

If not, you could also wrap the component in a `JssProvider` and collect every sheet via a `SheetsRegistry`. You would have access to the instances and check the `attached` property.

kof:
Yes, please test using existing public API's. It's already possible.

PR #1111, opened 2019-05-23
Fix for case where `global` is unset

buu700:
In line with recent changes to `src/utils/escape.js`, this handles situations where `global` doesn't exist (which is often the case in browsers, depending on the webpack configuration).


kof:
We are relyling on global in many places, it needs to be fixed as part of build step. I am not aware how you got to this issue in the first place, please create an issue first.

Issue #677, opened 2018-02-04
Add a rehydrate function 


kof:
Initially I went against creating a separate rehydrate function, because we decided to not actually do rehydration, but instead just regenerate the styles, because it is fast enough and more robust. 

So if I did `jss.rehydrate('.style-selector')´ it would be basically inserting the styles after this element and just removing the old style node.

Now as some people had issues with mismatching selectors from SSR, it makes sense to add some more logic to rehydrate, which will make sure in dev mode that user gets proper error report.

1. Check the JSS versions matching between ssr and client.
1. Check the amount of JSS instances matches between ssr and client.
1. Check the module id matches between ssr and client.

All checks could be done for e.g. if ssr generates a comment string which contains the information (dev mode only). Rehydration call checks the comment and prints a warning or error.

kof:
cc @jedwards1211

Aetherall:
Indeed, It could be really usefull to have those error reports. Actually managing packages which uses jss as dependency is a pain

PR #1117, opened 2019-06-10
[react-jss] removed classNamePrefix from getSheetOptions

chrisands:
what's new:
- edited tests to fit new requirements

classNamePrefix hardcoded inside `react-jss` creates inconsistencies between packages and should be managed in inside `jss` or by developer (via `classNamePrefix` prop).

### DIFF
--- a/docs/react-jss-hoc.md
+++ b/docs/react-jss-hoc.md
@@ -349,6 +349,57 @@ export default function render(req, res) {
 }
 ```
 
+You also can use`classNamePrefix={false}` in `JssProvider` to disable additional prefix that is compiled.
+
+```javascript
+import React from 'react'
+import {renderToString} from 'react-dom/server'
+import {JssProvider, SheetsRegistry, createGenerateId} from 'react-jss'
+import MyApp from './MyApp'
+
+export default function render(req, res) {
+  const sheets = new SheetsRegistry()
+  const generateId = createGenerateId()
+
+  const body = renderToString(
+    <JssProvider classNamePrefix={false} registry={sheets} generateId={generateId}>
+      <MyApp />
+    </JssProvider>
+  )
+
+  // Any instances using `useStyles` within `<MyApp />` will have gotten sheets
+  // from `context` and added their Style Sheets to it by now.
+
+  return res.send(
+    renderToString(
+      <html lang="en">
+        <head>
+          <style type="text/css">{sheets.toString()}</style>
+        </head>
+        <body>{body}</body>
+      </html>
+    )
+  )
+}
+```
+
+```javascript
+import React from 'react'
+import withStyles from 'react-jss'
+
+const styles = {
+  container: {
+    background: 'white'
+  }
+}
+
+const MyApp = ({classes}) => <div className={classes.container}>MyApp</div>
+
+export default withStyles(styles)(MyApp)
+```
+
+Class names will compile to `container-0-1-1` instead of `MyApp-container-0-1-1`
+
 #### React tree traversing
 
 For traversing the React tree outside of the HTML rendering, you should add `disableStylesGeneration` property.
--- a/packages/react-jss/src/JssProvider.js
+++ b/packages/react-jss/src/JssProvider.js
@@ -18,7 +18,7 @@ type Props = {|
   jss?: Jss,
   registry?: SheetsRegistry,
   generateId?: GenerateId,
-  classNamePrefix?: string,
+  classNamePrefix?: string | false,
   disableStylesGeneration?: boolean,
   media?: string,
   children: Node,
@@ -30,7 +30,7 @@ export default class JssProvider extends Component<Props> {
     registry: PropTypes.instanceOf(SheetsRegistry),
     jss: PropTypes.instanceOf(defaultJss.constructor),
     generateId: PropTypes.func,
-    classNamePrefix: PropTypes.string,
+    classNamePrefix: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
     disableStylesGeneration: PropTypes.bool,
     children: PropTypes.node.isRequired,
     media: PropTypes.string,
@@ -73,6 +73,9 @@ export default class JssProvider extends Component<Props> {
     // Merge the classname prefix
     if (classNamePrefix) {
       context.sheetOptions.classNamePrefix += classNamePrefix
+    }
+    if (classNamePrefix === false) {
+      context.sheetOptions.classNamePrefix = classNamePrefix
     }
 
     if (media !== undefined) {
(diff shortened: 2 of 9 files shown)


kof:
Can you pls expand on what you mean by inconsistencies? I don't quite understand right now.

chrisands:
@kof  I'll try.

As I understand, `createGenerateId` is generate classNames. In `react-jss` there are function `getSheetOptions` that changes behaviour of `classNamePrefix` prop by adding `options.name` as prefix while it is in dev mode.

When I wrote that fix I thought - why in even exist in `react-jss`. It broke my application and I could even understand why this happens.

But right now I understand that It helps to find where a component is come from. 

Shouldn't it be in `createGenerateId` so every `jss` developer get that help? 
Or at least documented somewhere about that behavior? I spend a lot of time to understand why this happens.

kof:
Actually, the plan is to change the behavior to always use prefix in both production and development for better DX. If user wants to have shorter class names in production they can still enable it manually with 1 setting.


We did that already in core jss package, but we forgot to do this in react-jss.

Another question is why your application broke, can you create a codesandbox example with the problem you have right now?

HenriBeck:
Is your problem related to SSR? It usually helps first creating an issue as we could have pointed to the issue more quickly. 

As you said, this is, so it’s easier to understand in dev mode from which component the rule comes from. If your setup is correct, then this shouldn’t cause any issues.

Having said that, I don’t think this is the fix to your solution.

chrisands:
@kof that is wonderful news. Thank you.

I already found out what was a problem. Problem was that `NODE_ENV` wasn't set on a server, but it was on Webpack config. And when I ran my production build classNames were mismatched because on a server it generated with prefix `options.name` and on a client without. And when I was deleting SSR styles there were no applied styles on my app.

kof:
So  if you like you can change the PR to always add the prefix. Additionally, we need to enable turning off the prefix completely, e.g. by using {classNamePrefix: false} on JssProvider. This would also need to be documented.

packages/react-jss/src/utils/sheets.js
```diff
@@ -33,15 +33,15 @@ const getStyles = <Theme>(options: Options<Theme>) => {
...
-
   return {
     ...options.sheetOptions,
     ...options.context.sheetOptions,
     index: options.index,
     meta: `${options.name}, ${typeof options.styles === 'function' ? 'Themed' : 'Unthemed'}`,
-    classNamePrefix: options.context.sheetOptions.classNamePrefix + classNamePrefix,
+    classNamePrefix: `${options.context.sheetOptions.classNamePrefix}${options.name.replace(
+      /\s/g,
+      '-'
```
HenriBeck: Just for readability, could you keep the `classNamePrefix ` variable? I find it a bit hard to read now

kof:
- [ ] docs
- [ ] {classNamePrefix: false} on JssProvider

chrisands:
I am trying to fix #1119 with no success, but I noticed that when I use a hook `createUseStyles` `options.name` becomes 'Hook-'. When the whole app will be written with `react-jss` hooks, every component prefix will be 'Hook-' and that makes no sense.
Also, hooks don't work with class components. Not sure if `react-jss` should throw away class support.

HenriBeck:
#1119 is a bit more complicated.

packages/react-jss/src/JssProvider.js
```diff
@@ -18,7 +18,7 @@ type Props = {|
   jss?: Jss,
   registry?: SheetsRegistry,
   generateId?: GenerateId,
-  classNamePrefix?: string,
+  classNamePrefix?: string | any,
```
HenriBeck:
```suggestion
  classNamePrefix?: string | false,
```

packages/react-jss/src/index.d.ts
```diff
@@ -47,6 +47,8 @@ interface Options extends StyleSheetFactoryOptions {
 
 type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
 
+declare function createUseStyles<S extends Styles<string> | ThemedStyles<any>>(styles: S): object
```
HenriBeck: Please revert this again, I will add proper types this week

packages/react-jss/src/types.js
```diff
@@ -9,7 +9,7 @@ export type Managers = {[key: number]: SheetsManager}
 
 type StyleSheetOptions = {
   ...StyleSheetFactoryOptions,
-  classNamePrefix: string
+  classNamePrefix: string | any
```
HenriBeck:
```suggestion
  classNamePrefix: string | false
```

chrisands: It throws flow error `Cannot add options.context.sheetOptions.classNamePrefix and prefix because boolean literal false [1] is incompatible
with string`

kof: sure, you have to enable this in the type of sheet options

chrisands: I did, but I still get the same error. Should I commit without checks?

kof: yes if you want us to look at the problem

docs/react-jss-hoc.md
```diff
@@ -349,6 +349,57 @@ export default function render(req, res) {
...
+    background: 'white'
+  }
+}
+
+const MyApp = ({classes}) => <div className={classes.container}>MyApp</div>
+
+export default withStyles(styles)(MyApp)
+```
+
+Class names will compile to `container-0-1-1` instead of `MyApp-container-0-1-1`
```
kof: I actually thought that setting it to false will also turn on the minify option in jss id generator. I think we would be misusing `classNamePrefix ` if we did it over that prop, basically I thought we need a way to turn on minified classes in react-jss.

@HenriBeck

chrisands: So basically it should do the same thing as `createGenerateId({ minify: true })`?

kof:
I just looked into JssProvider, we have 2 undocumented props, `media` and `id`, where id can be `{minify: false}` and that's what is passed to `createGenerateId(options)`, but we never document3ed it.

kof:
So those are 2 separate things, `classNamePrefix` is a prefix for the subtree that is rendered inside of a particular provider, `displayName` of a component is a second prefix that adds info about the specific component and  id option `minify` is a flag to remove all prefixes.

We need to better document all 3.

kof:
> I already found out what was a problem. Problem was that NODE_ENV wasn't set on a server, but it was on Webpack config. And when I ran my production build classNames were mismatched because on a server it generated with prefix options.name and on a client without. And when I was deleting SSR styles there were no applied styles on my app.

I think this issue once implemented would have fixed it https://github.com/cssinjs/jss/issues/677

kof:
@chrisands I am sorry for the confusion I have created here, I actually forgot how things work. I will document prefixing in a separate PR right now. I think there is actually no code changes needed, just documentation. 

Code change that we actually need is improving warnings to avoid things like different NODE_ENV.

chrisands:
Thank you

kof:
@chrisands check out https://github.com/cssinjs/jss/pull/1147/files

PR #1121, opened 2019-06-11
Match ts types to flow types, theming is optional

nikuda:

Match TypeScript type defs with flow ones, which seems to be the correct one.

`theming` was required in TypeScript defs.


HenriBeck:
Thanks 👍

PR #1124, opened 2019-06-12
Export useTheme in TypeScript declaration

nikuda:

Export `useTheme` in TypeScript declaration.

Still missing: Declaration and export of `createUseStyles` fn.


HenriBeck:
Thanks 👍 

I'm working on definitions for `createUseStyles`

kof:
Lets always merge PR's with changelogs, we should probably make it fail the PR.

Issue #747, opened 2018-07-06
Create a package with css() interface


kof:
An interface similar to  [glamor](https://github.com/threepointone/glamor) that allows to use objects and strings would be a good fit for many people. One of the biggest benefits would be support of [highlighting](https://github.com/cssinjs/jss/issues/591#issuecomment-402826872)

romelperez:
Also, it can help with safe typing with flow or typescript is applicable.

kof:
How?

HenriBeck:
We can striclty type the object that the css function receives.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

kof:
We can not jss.createStyleSheet?

HenriBeck:
Not really

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

HenriBeck:
The css function can just take an object with css styles

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

romelperez:
I mean, if you use JavaScript normally, it would work without typing. Maybe it can be expensive to check all properties on runtime. But we can add type checking for flow or typescript when used. In TypeScript there is [typestyle](https://github.com/typestyle/typestyle).

kof:
> Not really. The problem is the nested syntax.

css() would do the same

kof:
Actually I am  not sure it can't be fully typed, its not easy though.

romelperez:
Well, maybe adding typescript support to all packages is an entire task. I don't use flow, I don't know if the typing can be set up when the packages are consumed.

kof:
Can we clean up this issue from the comments which don't provide any value, we should use a chat for stuff like this. Issues are here to document something.

kof:
I am removing my comments from this thread, its useless.

romelperez:
As a future improvement, we can add type checking with flow or typescript after setup when defining object properties. A good example is how [typestyle](https://github.com/typestyle/typestyle) does it with typescript.

kof:
Common guys, css() function has nothing directly todo with types, its not the reason. I will update the original description with the actual reason. It can only help with typing if you reduce the styling interface from everything that is hard to type and this is not the intention of the issue.

Please clean up this comments mess.

kof:
I totally forgot that we have it already: https://github.com/dan-lee/glamor-jss

kof:
[...]

PR #1129, opened 2019-06-12
New css interface

kof:
Implements #747, for now just objects syntax.


packages/css-jss/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "css-jss",
```
TrySound: why not `jss-css`?

packages/css-jss/src/createCss.js
```diff
@@ -0,0 +1,31 @@
...
+    for (const i in args) {
+      const arg = args[i]
+      if (arg) Object.assign(style, arg)
+    }
+    if ('label' in style) {
+      label = style.label
+      delete style.label
+    }
+    const key = `${label}${ruleIndex++}`
+    sheet.addRule(key, style)
```
TrySound: Will this deduplicate the same style created inside of component render?

packages/css-jss/package.json
kof: because  like react-jss is extension for react,  css-jss is extension for css interface

packages/css-jss/src/createCss.js
kof: yes, if object refs are not the same, to avoid this, one will need to use a babel plugin (planned)

kof: Alternatively I could use JSON.string() and compare by that ...

TrySound: Does jss serialize styles internally? Maybe use serialized value to get the key?

kof: Nope, we optimized that away in JSS.

kof: Actually with this logic I can also remove the ref cache and just use string based keys and hope that user will use babel plugin to avoid the need for JSON.stringify()

packages/css-jss/package.json
kof: I am not sure though, if this sounds weird we can do the other way around, because one can think of it as of an extension of jss too

kof: actually I start thinking it makes more sense as css-jss

kof: wdyt @HenriBeck @oliviertassinari

HenriBeck: I like `css-jss` more, just as it's an interface for `jss` and we usually name these like `-jss`

packages/css-jss/src/index.js
```diff
@@ -0,0 +1,18 @@
...
+import createCss from './createCss'
+import type {Css, Style} from './types'
+
+const jss = createJss(preset())
+const defaultSheet = jss.createStyleSheet().attach()
+
+export type {Css, Style}
+
+export const create = (sheet?: StyleSheet): Css => {
+  if (sheet) defaultSheet.detach()
```
HenriBeck: Why would we detach in such a case? The user still could have used the default sheet

packages/css-jss/src/index.js
```diff
@@ -0,0 +1,18 @@
...
+  return createCss(sheet || defaultSheet)
+}
+
+export default create()
```
HenriBeck:
```suggestion
export default createCss(defaultSheet)
```

And make sheet non-optional in `create`

packages/css-jss/src/index.js
kof: Yeah, not sure what to do here, the problem I see is that if user creates custom jss/sheet, this sheet will already be in the DOM with no rules.

So maybe I need to additionally check if there are rules

HenriBeck: Would this be a big benefit? I don’t think it will make a big difference in performance

packages/css-jss/src/types.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+export type ClassName = string
+
+export type Style = {} | null | void | '' | ClassName
```
HenriBeck: Using ’ClassName’ here is a bit confusing here. Is it really a class name or a css string?

packages/css-jss/src/index.js
kof: Just the fact that there would be an empty style tag in the head, though evtl it doesn't matter for client side and won't happen on the server so fuck it, I am removing the line.

packages/css-jss/src/types.js
kof: well it corresponds the className prop

kof: actually using it to be more explicit here about the string

kof: but actually it doesn't matter that much, moving it to just string

kof:
merging it now to make it into the next alpha, if you find anything else, write it here or  https://github.com/cssinjs/jss/issues/747

PR #1138, opened 2019-06-19
Small readability improvements

kof:


packages/react-jss/src/createUseStyles.js
```diff
@@ -52,16 +52,20 @@ const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions
...
+        classes = getSheetClasses(sheet, dynamicRules)
       }
 
-      const dynamicRules = sheet && addDynamicRules(sheet, data)
-
       return {
         sheet,
         dynamicRules,
-        classes: sheet ? getSheetClasses(sheet, dynamicRules) : {}
+        classes: classes || {}
```
HenriBeck: Classes should always be an object, so the user can safely access any props when `disableStylesGeneration` is enabled

PR #1149, opened 2019-07-03
HOC should not attach sheets until mount

### LINKED ISSUE #1143: Incorrect "Rule is not linked" warning on server when using function rules
__Expected behavior:__
No warning.

__Describe the bug:__
All function rules are triggering the following warning when run on the server: *Warning: [JSS] Rule is not linked. Missing sheet option "link: true".*

I've traced the problem back to the fact that in `Jss.js`, `options.Renderer` is set to `null` when not in the browser. This eventually leads to the code in `BaseStyleRule.prop()` falling through to the warning because `this.renderable` and `this.renderer` are not defined at [styleRule.js#L71](https://github.com/cssinjs/jss/blob/master/packages/jss/src/plugins/styleRule.js#L71).

The warning is incorrect: the sheet option `link` is actually `true`, and the fact that `this.renderable` and `this.renderer` are not defined is correct because on the server, we don't need anything to happen in `prop()` because we'll call `toString()` later the render the CSS.

I would contribute a test to replicate the bug, but I don't think it's possible with the existing karma setup because we need `'is-in-browser'` to report `false` to trigger the bug. So, I've written a jest test (see link below) to demonstrate.

__Codesandbox link:__ 
https://repl.it/@felthy/JssInvalidWarningDuringSSR

__Versions:__
 - jss: 10.0.0-alpha.20

__jest test source:__
```javascript
/**
 * @jest-environment node
 */
import warning from 'tiny-warning'
import {create, SheetsRegistry, SheetsManager} from 'jss/dist/jss.cjs'
import functionPlugin from 'jss-plugin-rule-value-function/dist/jss-plugin-rule-value-function.cjs'
import isInBrowser from 'is-in-browser'

jest.mock('tiny-warning')

describe('React-JSS: rendering in an SSR context', () => {
  it('should render a dynamic rule without warnings', () => {
    expect(isInBrowser).toBe(false)

    const jss = create().use(functionPlugin())
      rule: () => ({
        color: 'red'
      })
    }, {
      link: true
    }).attach()
    sheet.update({})
    const css = sheet.toString()

    expect(css).toContain('color: red;')
    expect(warning).not.toHaveBeenCalled()
  })
})

```

felthy:
Fixes #1143.

Per the discussion in #1140 and, prior to that, #1089, sheets should not be attached during server-side rendering.

Now that #1148 has been merged, server rendering is working for the hook API (`createUseStyles()`) but the HOC (`withStyles()`) hasn't been updated yet so still warns (incorrectly) `Rule is not linked. Missing sheet option "link: true"` on the server.

This PR brings the HOC implementation into line with the hook implementation - adding the sheet to the `SheetsRegistry` when the component is constructed, and not attaching (managing) it until the component is mounted (which never occurs on the server).

### DIFF
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -1,5 +1,6 @@
 // @flow
 import React, {Component, type ComponentType, type Node} from 'react'
+import isInBrowser from 'is-in-browser'
 import hoistNonReactStatics from 'hoist-non-react-statics'
 import {type StyleSheet, type Classes} from 'jss'
 import {ThemeContext} from 'theming'
@@ -113,16 +114,14 @@ const withStyles = <Theme>(styles: Styles<Theme>, options?: HOCOptions<Theme> = 
         super(props)
 
         this.state = WithStyles.createState(props)
-        const context = props.jssContext
-        const {sheet} = this.state
-        if (sheet && context.registry) {
-          context.registry.add(sheet)
-        }
-      }
-
-      componentDidMount() {
-        if (this.state.sheet) {
-          WithStyles.manage(this.props, this.state)
+        if (isInBrowser) {
+          WithStyles.manage(props, this.state)
+        } else {
+          const {registry} = props.jssContext
+          const {sheet} = this.state
+          if (sheet && registry) {
+            registry.add(sheet)
+          }
         }
       }
 
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -113,7 +113,17 @@ const withStyles = <Theme>(styles: Styles<Theme>, options?: HOCOptions<Theme> = 
         super(props)
 
         this.state = WithStyles.createState(props)
-        WithStyles.manage(props, this.state)
+        const context = props.jssContext
+        const {sheet} = this.state
+        if (sheet && context.registry) {
+          context.registry.add(sheet)
+        }
+      }
+
+      componentDidMount() {
+        if (this.state.sheet) {
+          WithStyles.manage(this.props, this.state)
+        }
       }
 
       componentDidUpdate(prevProps: HOCProps<Theme, Props>, prevState: State) {
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -114,14 +114,13 @@ const withStyles = <Theme>(styles: Styles<Theme>, options?: HOCOptions<Theme> = 
         super(props)
 
         this.state = WithStyles.createState(props)
+        const {registry} = props.jssContext
+        const {sheet} = this.state
+        if (sheet && registry) {
+          registry.add(sheet)
+        }
         if (isInBrowser) {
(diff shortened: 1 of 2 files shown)


packages/react-jss/src/withStyles.js
```diff
@@ -113,7 +113,17 @@ const withStyles = <Theme>(styles: Styles<Theme>, options?: HOCOptions<Theme> =
...
+        const context = props.jssContext
+        const {sheet} = this.state
+        if (sheet && context.registry) {
+          context.registry.add(sheet)
+        }
+      }
+
+      componentDidMount() {
+        if (this.state.sheet) {
+          WithStyles.manage(this.props, this.state)
```
HenriBeck: This is wrong. This would only after the first render meaning a flash of unstyled HTML!

We should check in the constructor if we render on the server and then not manage the sheet

felthy: Thanks, I overlooked that.

packages/react-jss/src/withStyles.js
```diff
@@ -113,7 +114,15 @@ const withStyles = <Theme>(styles: Styles<Theme>, options?: HOCOptions<Theme> =
         super(props)
 
         this.state = WithStyles.createState(props)
-        WithStyles.manage(props, this.state)
+        if (isInBrowser) {
+          WithStyles.manage(props, this.state)
+        } else {
+          const {registry} = props.jssContext
```
HenriBeck: This should also run on the client to be in sync with the hook version

felthy: Or, should we change the hooks version to be like this instead? After all, `manageSheet()` also calls `registry.add()`.

felthy: For now though I'll do as the hooks API does

kof:
Changelog forgotten!

kof:
Looking at this, we shouldn't be calling .attach in constructor at all, this should be done in `componentDidMount`

HenriBeck:
Wouldn’t we then have a flash of uncomputed styles?

kof:
Nope, didMount happens sync after rendering, so one can still do stuff like that, before browser starts drawing

kof:
I had a discussion with the core team about this a while ago, it sounds dangerous but it actually is guaranteed.

BenF-Lumedic:
Unsure if this is the right place to ask this but I noticed a change in test behavior (jest+enzyme, react-jss components) which i'm wondering is a side-effect of this update?

It's not a problem, I'm just curious if I'm understanding the functionality I'm seeing & if there are any other changes I missed that might contribute (I just updated react-jss from around ~8.6.* to 10.0.0).

Previously components wrapped in `withStyles` and `shallow` rendered with enzyme would give an error when no `ThemeProvider` was present. 

After updating the same components `shallow` rendered don't give any errors. Using the enzyme method `mount` does still give an error.

I think this makes sense, because enzyme it isn't fully mounting the component with `shallow`, so there is no need for react-jss to find the ThemeProvider to actually attach the necessary sheets?

PR #1151, opened 2019-07-05
Replace global polyfill with globalThis

TrySound:
https://twitter.com/IAmTrySound/status/1147174216095211520

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,6 @@
     "rollup": "^1.1.2",
     "rollup-plugin-babel": "^4.3.2",
     "rollup-plugin-commonjs": "^9.2.0",
-    "rollup-plugin-node-globals": "^1.4.0",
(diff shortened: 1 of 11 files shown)


kof:
some new errors here https://travis-ci.org/cssinjs/jss/builds/554779472#L721

kof:
seems like it doesn't work on IE10, if so  I think this package wasn't ever widely used

TrySound:
Yes, ie10 support is not provided. Why do you still support it?

kof:
because react does too

TrySound:
Not really. React requires polyfills
https://reactjs.org/docs/react-dom.html#browser-support

kof:
we can ask to add polyfills too if that solves the problem

kof:
we actually already do

PR #1152, opened 2019-07-05
Update browsers.json

kof:
WIP
https://github.com/cssinjs/jss/issues/1158


TrySound:
Yay, passed

kof:
nope, I am creating a new list of browsers, that was just a test

AleshaOleg:
@kof try to use latest OS version (like for android use 9/10 instead of 8) or in some cases even another OS (for example Windows instead MacOS). That helped me once

PR #1153, opened 2019-07-05
Use mui like global ponyfill

TrySound:
https://twitter.com/IAmTrySound/status/1147174216095211520

With globalThis polyfill we have a browser support problem. Let's try
this instead
https://github.com/mui-org/material-ui/blob/master/packages/material-ui-utils/src/ponyfillGlobal.js


packages/jss/src/utils/moduleId.js
```diff
@@ -1,8 +1,10 @@
+import globalThis from './ponyfillGlobal'
```
kof: let's just call it `globalThis`?

kof:
TLDR when using `global`,  users webpack creates module wrappers and prevents tree shaking, so we now try to not use `global` at all.

PR #1154, opened 2019-07-06
Demo link is not work

uxitten:
Link is out of service


kof:
We changed the name of plugins and forgot to change it in examples repo, so the solution to this is to update examples)

Issue #1159, opened 2019-07-09
[rule-value-function|camel-case] Cannot read property 'fallbacks' of undefined (or null)


akomm:
Exactly the same issue, as [this](https://github.com/cssinjs/jss/issues/489), except the null / undefined value comes from a rule-function:

```js
{
  font: () => null
}
```
**update: added sandbox**
Version: 10.0.0-alpha.22
Sandbox: https://codesandbox.io/s/hungry-glitter-pb2p2

HenriBeck:
What version are you using and could you please create a codesandbox where this issue exists?

akomm:
Version: alpha.22
Sandbox: https://codesandbox.io/s/hungry-glitter-pb2p2

Must be a bug introduced in alpha, because in the sandbox switching ONLY jss to latest 9.x.x fixed the issue. :)

akomm:
This works:
```js
jss.createStyleSheet({
  a: undefined,
});
```
This works:
```js
jss.createStyleSheet({
  a: {
    fontSize: () => undefined,
  },
});
```

This is does not work:
```js
jss.createStyleSheet({
  a: () => undefined,
});
```

PluginRegistry checks rule.style for null in onProcessRule and passes it to onProcessStyle, however onProcessStyle runs through plugins and does not check whether the result is null/undefined. So when the function plugin returns null or undefined, it then passes it to camelCase. So bug is not in the plugin, but jss PluginRegistry. Will provide PR soon.

PR #1160, opened 2019-07-10
[jss] fix PluginRegistry passing null|undefined to onProcessStyle hooks (#1159)

akomm:
Fixes issue: #1159

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,12 +4,9 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
-## 10.0.0-alpha.22 (2019-7-10)
-
-- [jss] fix PluginRegistry passing null|undefined to onProcessStyle hooks [#1159](https://github.com/cssinjs/jss/issues/1159)
-
 ### Bug fixes
 
+- [jss] fix PluginRegistry passing null|undefined to onProcessStyle hooks [#1159](https://github.com/cssinjs/jss/issues/1159)
 - [react-jss] Update dynamic nested rules need ([#1144](https://github.com/cssinjs/jss/pull/1144))
 - [react-jss] withStyles shouldn't attach() on SSR ([#1149](https://github.com/cssinjs/jss/pull/1149), [#1157](https://github.com/cssinjs/jss/pull/1157))
 
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,10 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+## 10.0.0-alpha.22 (2019-7-10)
+
+- [jss] fix PluginRegistry passing null|undefined to onProcessStyle hooks [#1159](https://github.com/cssinjs/jss/issues/1159)
+
 ### Bug fixes
 
 - [react-jss] Update dynamic nested rules need ([#1144](https://github.com/cssinjs/jss/pull/1144))
--- a/packages/jss/src/PluginsRegistry.js
+++ b/packages/jss/src/PluginsRegistry.js
@@ -71,6 +71,7 @@ export default class PluginsRegistry {
     for (let i = 0; i < this.registry.onProcessStyle.length; i++) {
       // $FlowFixMe
       rule.style = this.registry.onProcessStyle[i](rule.style, rule, sheet)
+      if (!rule.style) return
     }
   }
 
--- a/packages/jss/tests/integration/plugins.js
+++ b/packages/jss/tests/integration/plugins.js
@@ -407,13 +407,13 @@ describe('Integration: plugins', () => {
     })
 
     it('should not call the hook if rule style is null or undefined', () => {
-      let localExecuted = 0
+      let hookExecuted = 0
 
       jss = create({
         plugins: [
           {
             onProcessStyle: style => {
-              localExecuted++
+              hookExecuted++
               return style
             }
           }
@@ -425,24 +425,24 @@ describe('Integration: plugins', () => {
         b: undefined
       })
 
-      expect(localExecuted).to.be(0)
+      expect(hookExecuted).to.be(0)
     })
 
     it('should not call the next hook if rule style resolves to null or undefined', () => {
-      let localHookExecuted = 0
-      let localNextHookExecuted = 0
+      let firstHookExecuted = 0
+      let secondHookExecuted = 0
 
       jss = create({
         plugins: [
           {
             onProcessStyle: style => {
-              localHookExecuted++
+              firstHookExecuted++
               return typeof style === 'function' ? style() : style
             }
           },
           {
             onProcessStyle: style => {
-              localNextHookExecuted++
+              secondHookExecuted++
               return style
             }
           }
@@ -456,8 +456,8 @@ describe('Integration: plugins', () => {
         })
       }).to.not.throwException()
 
-      expect(localHookExecuted).to.be(2)
-      expect(localNextHookExecuted).to.be(0)
+      expect(firstHookExecuted).to.be(2)
+      expect(secondHookExecuted).to.be(0)
     })
 
     it('should pass the style object to the next hook', () => {
--- a/packages/jss/tests/integration/plugins.js
+++ b/packages/jss/tests/integration/plugins.js
@@ -406,27 +406,58 @@ describe('Integration: plugins', () => {
       `)
     })
 
-    it('should not call the hook if rule has no .style', () => {
+    it('should not call the hook if rule style is null or undefined', () => {
       let localExecuted = 0
-      jss.use({
-        onProcessStyle: (style, rule, passedSheet) => {
-          receivedStyle = style
-          receivedRule = rule
-          receivedSheet = passedSheet
-          localExecuted++
-          return style
-        }
-      })
-      sheet = jss.createStyleSheet({
-        '@media all': {
-          a: {color: 'red'}
-        }
+
+      jss = create({
+        plugins: [
+          {
+            onProcessStyle: style => {
+              localExecuted++
+              return style
+            }
+          }
+        ]
       })
 
-      expect(receivedStyle).to.be(newStyle)
-      expect(receivedRule.type).to.be('style')
-      expect(receivedSheet).to.be(sheet)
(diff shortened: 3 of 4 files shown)


akomm:
@HenriBeck 

Karma flaky?

HenriBeck:
yes in the CI it is very flaky

packages/jss/src/PluginsRegistry.js
```diff
@@ -71,6 +71,7 @@ export default class PluginsRegistry {
     for (let i = 0; i < this.registry.onProcessStyle.length; i++) {
       // $FlowFixMe
       rule.style = this.registry.onProcessStyle[i](rule.style, rule, sheet)
+      if (!rule.style) return
```
kof: onProcessStyle  was supposed to always return a new style object, can we pls move this check to the plugin?

kof: The function rule plugin needs to return previous style value when user function returns empty

packages/jss/tests/integration/plugins.js
```diff
@@ -406,27 +406,58 @@ describe('Integration: plugins', () => {
       `)
     })
 
-    it('should not call the hook if rule has no .style', () => {
```
kof: tests should also be in the plugin

packages/jss/src/PluginsRegistry.js
kof: onProcessStyle was designed similar to a [].map function

akomm: For me it looks more like a `[].reduce`. Given the following arguments, you might want to reconsider it:

- It would make behavior inconsistent between passing a null|undefined style directly and having a plugin that produces this output. 
- Given example of the problem I provided, the style is not an `{...}`, but a function. So basically the plugin would have to invoke the function, check for null|undefined and if it is, return the function back if its null|undefined. You would expect the user function being a map of input -> style, but it behaves differently to 1. providing null|undefined as value (works) 2. what user would expect as a result of returning nothing (no style rule applied, like in 1.)
- You would have to implement checks in every plugin (even some internal fail with exception because currently they do not, for example keyframes internal).
- You would run hooks even thought it does not make any sense to run them

The last two points also have performance implications. While if it did make sense, it would be not a question of performance, but it also does not make much sense in reflection of the first two points, so why? It makes it also more error prone, because now you can do more things wrong in plugins and you can do more things wrong as consumer of such a plugin, by expecting a function plugin to resolve to value returned and not to the function provided itself, because it returned null|undefined. And this is only an example of issues I encountered in short period of time. I am pretty sure this leads to more soon.

packages/jss/tests/integration/plugins.js
akomm: Yes. But I think it depends on the points I gave in the other comment :)

changelog.md
```diff
@@ -4,6 +4,10 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+## 10.0.0-alpha.22 (2019-7-10)
```
HenriBeck: Please remove the version, this will be automatically added by our build script and put the changelog entry into the `Bug fixes` section

packages/jss/src/PluginsRegistry.js
HenriBeck: > The function rule plugin needs to return previous style value when user function returns empty

I don't think that's correct. It should return an empty object in my opinion.

kof: @HenriBeck if there was no style object before - yes it should return empty object

akomm: @kof
> if there was no style object before - yes it should return empty object, I think though there is always a style object, because there should be an empty one by default

There can be no style object, it you pass a function as `rule.style` and the function value plugin is active.

@HenriBeck 
Then [this](https://github.com/akomm/jss/blob/fix-1159/packages/jss/src/PluginsRegistry.js#L62) should not bail out, but run call `onProcessStyle` with empty style object too.

Also note, `null|undefined` is easier checked for changes, than new empty object created on each run/update, even though it did not change and is still empty.

kof: I have tried to reproduce/fix it and ended up with this 
 https://github.com/cssinjs/jss/pull/1164 

I tried first to support  a falsy value directly in the core, but this resulted in so many falsy value checks in different places that I immediately regretted it and reverted to just using a default empty object in the plugin.

I think this is an edge case.

kof: I wonder if we should warn when onProcessStyle doesn't return an object to ensure this connsistently.

HenriBeck:
This can be closed now, right?

packages/jss/src/PluginsRegistry.js
akomm: The question preventing me from moving it to the plugin in this PR is still unaswere though.

akomm: The question preventing me from moving it to the plugin in this PR is still unaswered though.
Btw, bailing out would have required only 1 check AFAIK. Because other plugins would not get to check, as they never get falsy value then. But nvm.

akomm: What argument is there to continue process onProcessStyle further plugins with empty object? So that maybe a scenario is possible where plugin inserts something into this empty object? Okay. Might be useful. Then why should not be the same possible when doing exactly same thing just without function value plugin? See https://github.com/cssinjs/jss/pull/1160#discussion_r303336699

kof: > Btw, bailing out would have required only 1 check AFAIK. Because other plugins would not get to check, as they never get falsy value then. But nvm.

yeah and then we have a behaviour where any plugin can prevent other hooks being called by returning null, that wouldn't be easy to debug.

kof: >Then why should not be the same possible when doing exactly same thing just without function value plugin? See #1160 (comment)

Can you explain this pls?

akomm: `PluginRegistry.onProcessRule` does falsy-check `rule.style`. If its true, `PluginRegistry.onProcessStyle` is invoked. 

This means, when you provide style this way:
```js
jss.createStyleSheet({a: undefined})
```
**No** `onProcessStyle` hooks will be invoked. The resulting `rule.style` is falsy.

When you provide the style this way:
```js
jss.createStyleSheet({a: () => undefined})
```
**All** `onProcessStyle` hooks will be invoked. The resulting `rule.style` is (optimistically) never falsy, as falsy must be replaced with empty object by plugins.

You would except same outcome, but the outcome is different for this two scenarios.

If you do not want to handle falsy values by those `onProcessStyle` hooks, hence the check before in `onProcessRule`, why do you want to handle falsy values (by replacing with empty object) afterwards? 

Is there any good reason doing so? I gave an example, what it could make possible: A plugin that injects something into the empty object. If we would exit early, it would not be possible. Then the final question is, why should it not be possible when the style is false on first pass through `onProcessRule` too? Or the other way around: if bailing out on falsy value is preferred way and there is no intention to inject something via plugin into empty style objects, why do we do it in `onProcessStyle` ? It then just pointlessly runs through hooks and also provides inconsistent behavior where you expect consistency.

akomm: At least, given current fix you merged, `onProcessRule` should initialize `rule.style` to empty object, if it is falsy, instead of skipping `onProcessStyles`. Or you might reconsider to do it the other way around.

kof: > PluginRegistry.onProcessRule does falsy-check rule.style. If its true, PluginRegistry.onProcessStyle is invoked.

Yes the idea is that no `onProcessStyle` hook is invoked if there is no style to process. The hook is called `onProcessStyle`, sounds logical to me. It doesn't stop the call chain of that hook in the middle, it doesn't start it at all if there is no style.

kof: I think the main reason for the StyleRule.style to not been set to empty object by default is the emptiness check in some cases, because it's more work to check if `{}` is empty than a falsy value.

kof: > All onProcessStyle hooks will be invoked. The resulting rule.style is (optimistically) never falsy, as falsy must be replaced with empty object by plugins.

I agree though that in this case style is technically empty but is initialized with an empty object. We have been treeting function value/rules as an edge case so far, because in those cases, we don't know what style object will be in the future.

kof: Evtl. making sure that style object is never falsy is a better way for the future. cc @HenriBeck

kof: Another thing is that the check `if (rule.style)` in PluginsRegistry also checks it for rules which don't have a style map at all, which means that check will still have to exist.

kof: That check would be more explicit if it would be checking specific rule types which are supposed to have style object, but that would limit the plugins, because a plugin wouldn't be able to create a custom rule with a style prop,  whatever reason for this could be.

akomm: > I think the main reason for the StyleRule.style to not been set to empty object by default is the emptiness check in some cases, because it's more work to check if `{}` is empty than a falsy value.

Would this in `onProcessRule`:
```js
if (!rule.style) rule.style = {};
this.onProcessStyle(rule.style, rule, sheet)
```

Not still be a falsy check?

kof: There are other places I believe which would go iterate over empty style object even though it's empty, but it might be not too bad, gotta check.

In any case default object should be set in the corresponding rule classes.

kof: Adding //@flow to the tests and then changing JssStyle type to something more specific would reveal all the places probably, because currently it's just `Object` and that's just as `any` in the latest flow versions.

kof: Let's open a separate issue for this investigation? Because this PR got diverged from it's original purpose and I would rather close it and start fresh.

akomm: good idea

PR #1164, opened 2019-07-16
Support falsy value from fn rule

kof:
https://github.com/cssinjs/jss/issues/1159


HenriBeck:
Don't forget the changelog 😉

PR #1168, opened 2019-08-13
Consistent usage of the CSS browser API

oliviertassinari:
Related to https://github.com/mui-org/material-ui/issues/16953. I'm not 100% convinced its the right solution, but it's worth trying.

*I hope you don't mind the created branch, I will delete it after.*

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,16 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+## 10.0.0-alpha.24 (2019-8-13)
+
+- Consistent usage of the CSS browser API ([#1168](https://github.com/cssinjs/jss/pull/1168))
+
+### Bug fixes
+
+- [react-jss] Update dynamic nested rules need ([#1144](https://github.com/cssinjs/jss/pull/1144))
+- [react-jss] withStyles shouldn't attach() on SSR ([#1149](https://github.com/cssinjs/jss/pull/1149), [#1157](https://github.com/cssinjs/jss/pull/1157))
+
+
 ## 10.0.0-alpha.23 (2019-7-20)
 
 ### Bug fixes
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@ Since you are interested in what happens next, in case, you work for a for-profi
 ### Improvements
 
 - [react-jss] Improve TypeScript definitions and add missing definition for `createUseStyles` ([1155](https://github.com/cssinjs/jss/pull/1155))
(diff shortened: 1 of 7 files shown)


kof:
I wonder what didn't work, in node this code shouldn't run?

oliviertassinari:
I have tried to summarize the issue in this comment: https://github.com/mui-org/material-ui/issues/16953#issuecomment-520956695.

kof:
yeah I mean how it comes `window.CSS === undefined`?

oliviertassinari:
@kof It's curious. I don't know. At the same time, should we always use CSS or window.CSS. Is there a reason for mixing the two?

kof:
`CSS` and `window.CSS` should be exactly the same, since window is the global. `window.CSS` is a more explicit way, I don't mind changing it the way you did it, but I don't understand what's going on, I fear there is some sort of confusion caused by something else, mb polyfills

HenriBeck:
I think `CSS` should be safer because it tries to access `CSS` directly on the global object and not `window` on the global object which could be `undefined` and then it would crash.

kof:
`window` IS the global object in a browser environment, it is safe, but I think something gets messed up on the way. Anyways I am merging it now.

kof:
Forgot, @oliviertassinari ca you please add a changelog entry?

changelog.md
```diff
@@ -13,6 +13,12 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+## 10.0.0-alpha.24 (2019-8-13)
```
kof: we have a Next headline which will be replaced with the release version at release time

kof: so below Next is the next release changes

packages/jss-plugin-default-unit/src/defaultUnits.js
```diff
@@ -2,9 +2,9 @@
 
 import {hasCSSTOMSupport} from 'jss'
 
-const px = hasCSSTOMSupport ? window.CSS.px : 'px'
-const ms = hasCSSTOMSupport ? window.CSS.ms : 'ms'
-const percent = hasCSSTOMSupport ? window.CSS.percent : '%'
+const px = hasCSSTOMSupport && CSS ? CSS.px : 'px'
```
TrySound: You need to make sure CSS is defined or check with window.CSS
image

kof: hasCSSTOMSupport did that check already, I am just making flow happy

kof:
released

oliviertassinari:
Thanks for the release!

PR #1176, opened 2019-08-21
jss-plugin-nested docs fix

antvaset:
Just a small fix that resolves the confusion around how referencing local rules works.


HenriBeck:
Thanks 👍

Issue #1166, opened 2019-08-01
type Plugin should be JSPlugin


jacobbogers:
__Describe the bug:__

The index.d.ts of the jss-plugin-xxx has the following
```typescript
import {Plugin} from 'jss'

export default function jssPluginSyntaxNested(): Plugin
```

Should be `JSPlugin`

__Versions (please complete the following information):__
 - jss: "10.0.0-alpha.7"
 - Browser: chrome
 - OS: Windows

Shall i fix it?

HenriBeck:
Why would this need to be `JSPlugin`? we don't even have such a type.
Also, alpha 7 is outdated, please upgrade to the latest alpha and check if your problem still exists.

jacobbogers:
In typescript I am using the @types/jss with version "9.5.8"  
I simple installed with `npm i --save-dev @types/jss` if this outdated can you update the npm registry to the most current one

in the `index.d.ts` is shows 
`export function create(options?: Partial<JSSOptions>): JSS;``

and if you look up `JSSOptions` it shows 
```
export interface JSSOptions {
	createGenerateClassName(): GenerateClassName;
	plugins: ReadonlyArray<JSSPlugin>;  // <--- JSSPlugin not Plugin
	virtual: boolean;
	insertionPoint: string | HTMLElement;
}
```

HenriBeck:
Yeah, so that's your problem. `@types/jss` is incompatible with `jss` v10. In v10 we moved the types so you don't need `@types/jss` anymore.

jacobbogers:
Oh while you replied , I adjusted my original message))
I am not using explicit version numbers during `npm i`,  Is it possible for you to have the npm fetch the the correct one

HenriBeck:
Just do `npm install jss@next` and remove the `@types/jss` dependency

jacobbogers:
Ok

jacobbogers:
Also for the plugins?

HenriBeck:
yes

jacobbogers:
Ok, works

I also noticed the new jss.create({ plugins }) needs an array of type plugins instead of functions creating plugins `(options:Options) => Plugin `

PR #1186, opened 2019-09-05
Change Plugin to JSSPlugin

trinhhunganh:
Currently @types/jss exports JSSPlugin and not Plugin, so we need to refactor to JSSPlugin otherwise tsc raises error.


HenriBeck:
See #1166 for an explanation of how to fix your issue

trinhhunganh:
> See #1166 for an explanation of how to fix your issue
Awesome. Thank you @HenriBeck !

PR #1189, opened 2019-09-10
fix: Cannot read property rules of undefined

vtereshyn:
Not sure if it works for you, but I guess we shouldn't come into this condition if `rule` is undefined

image


kof:
I think the problem is somewhere before.

matteoferigo:
With this change, the issue will appear later (eg: rules is used even in row `199`.

To fix this issue is simply needed to prevent rules update when rule instance doesn't exist anymore

kof:
closing this PR since it isn't mergeable, please feel free to reopen later

PR #1194, opened 2019-09-22
Fix the CI

kof:


kof:
Good news, I just figured out that all IE's keep css rules inserted using insertRule API when style node is detached, that's why we still have a duplication of nodes in IE (cc @felthy).

I think JSS StyleSheet class needs to be refactored. Method `deploy()` needs to be removed and become a purely DOMRenderer implementation detail.

We are currently control the deployment from both StyleSheet and  DOMRenderer [see](https://github.com/cssinjs/jss/blob/2b2c1a05fd7640bbf5c0736847754e3a98790dc6/packages/jss/src/StyleSheet.js#L115), but this needs to happen in DOMRenderer and needs to be aware of browser specific differences.

kof:
the ci is finally green again

PR #1198, opened 2019-09-22
Fix fallbacks support in jss-plugin-vendor-prefixer

### LINKED ISSUE #1177: Multiple fallbacks for the same property name not working
**Expected behavior:**
When an array is used for the fallbacks value it should render multiple properties for the same property name

**Describe the bug:**
The array of fallbacks are being render like this:

```
.Component-withFallbacks-2-2-1 {
  0: [;
  1: o;
  2: b;
  3: j;
  ...
  29: c;
  30: t;
  31: ];
  display: flex;
}
```

The jss vendor prefix plugin seems to be changing it into the string
`"[object Object], [object Object]"`

**Codesandbox link:**
[https://codesandbox.io/s/jss-fallbacks-problem-gbvkf](https://codesandbox.io/s/jss-fallbacks-problem-gbvkf)

 - jss: 10.0.0-alpha.24
 - Browser: Chrome
 - OS Windows and macOS

kof:
fixes #1177

### DIFF
--- a/packages/jss-plugin-vendor-prefixer/src/index.js
+++ b/packages/jss-plugin-vendor-prefixer/src/index.js
@@ -15,12 +15,13 @@ export default function jssVendorPrefixer(): Plugin {
     }
   }
 
-  function onProcessStyle(style, rule) {
-    if (rule.type !== 'style') return style
-
+  function prefixStyle(style) {
     for (const prop in style) {
       const value = style[prop]
-
+      if (prop === 'fallbacks' && Array.isArray(value)) {
+        style[prop] = value.map(prefixStyle)
+        continue
+      }
       let changeProp = false
       const supportedProp = vendor.supportedProperty(prop)
       if (supportedProp && supportedProp !== prop) changeProp = true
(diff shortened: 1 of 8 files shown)


packages/jss-plugin-vendor-prefixer/src/index.js
```diff
@@ -15,12 +15,13 @@ export default function jssVendorPrefixer(): Plugin {
...
   }
 
-  function onProcessStyle(style, rule) {
-    if (rule.type !== 'style') return style
-
+  function prefixStyle(style) {
     for (const prop in style) {
       const value = style[prop]
-
+      if (prop === 'fallbacks' && Array.isArray(value)) {
```
HenriBeck: Should we also handle fallbacks which are just an object?

kof: adding a test

kof: added a test, it already works.

Issue #1204, opened 2019-10-03
[react-jss]: wrong 'transition' parse


chrisands:
__Expected behavior:__
`transition: 'ease-in-out 0.3s'` parses to `transition: ease-in-out 0.3s` or

__Describe the bug:__
`transition 'ease-in-out 0.3s'` parses to `transition: false 0.3s ease 0s`

> Note:
If don't use function in `createUseStyles` than `transition: 'ease-in-out 0.3s` parses to `transition: false 0.3s`

__Codesandbox link:__ 
https://codesandbox.io/s/stoic-hoover-91hdt

__Versions (please complete the following information):__
 - jss: 10.0
 - Browser [e.g. chrome, safari]: firefox, chrome
 - OS [e.g. Windows, macOS]: macOS

I would suggest that this is not `react-jss` problem, but `jss` parser.

kof:
I think we fixed something very similar a while ago in prefixer cc @AleshaOleg

chrisands:
Also, it doesn't parse `!important` on property value. Property just being ignored.

AleshaOleg:
Just another case for transition/transition-property properties, which is not covered. @chrisands thanks!

Here is PR: https://github.com/cssinjs/css-vendor/pull/169

AleshaOleg:
Fixed in `css-vendor@2.0.7`

AleshaOleg:
@kof https://github.com/cssinjs/jss/pull/1208

kof:
fixed, gonna be in 10.0.1

PR #1208, opened 2019-10-03
Updated css-vendor to 2.0.7

AleshaOleg:

Updated `css-vendor` to `2.0.7`.

https://github.com/cssinjs/jss/issues/1204

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,9 @@
-__What would you like to add/fix?__
+## Corresponding issue (if exists):
 
-__Corresponding issue (if exists):__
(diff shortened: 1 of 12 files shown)


kof:
pls add a changelog

AleshaOleg:
@kof which version will be next? `10.0.1`? Because I can't even find changelog for `10.0.0`

kof:
You don't need to define the next version, here is an example of how changelog looks like before a version is released: https://github.com/cssinjs/jss/blob/0934862f1340ae983e78c921419b386f6f9d0d92/changelog.md

AleshaOleg:
@kof

kof:
merged, thanks @AleshaOleg , sorry for waiting

PR #1209, opened 2019-10-04
Fix: typos

zelda-at-tempus:
Typos and wording in the introduction to the react-jss docs

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -4,7 +4,7 @@ React-JSS integrates [JSS](https://github.com/cssinjs/jss) with React using the 
 
 Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
 
-**The HoC based API is deprecated as of v10 and will be removed in an upcoming major version. You can still perform a lazy migration as described [here](https://reacttraining.com/blog/using-hooks-in-classes/). HoC specific docs are available [here](./react-jss-hoc.md).**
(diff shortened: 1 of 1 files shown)


docs/react-jss.md
```diff
@@ -4,7 +4,7 @@ React-JSS integrates [JSS](https://github.com/cssinjs/jss) with React using the
 
 Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
 
-**HOC based API is deprecated as of v10 and will be removed in any upcomming major version. You can still make a lazy migration as described [here](https://reacttraining.com/blog/using-hooks-in-classes/). HOC specific docs are available [here](./react-jss-hoc.md).**
+**The HoC based API is deprecated as of v10 and will be removed in an upcoming major version. You can still perform a lazy migration as described [here](https://reacttraining.com/blog/using-hooks-in-classes/). HoC specific docs are available [here](./react-jss-hoc.md).**
```
kof: react refers them as `HOC` in the documentation

kof: > an upcoming

means v11, any means it will be any of the major versions, but not necessarily v11

zelda-at-tempus: You are totally correct about the `HoC` thing. In retrospect, I don't know why I thought that change was necessary.

Re: an vs any, I just thought an sounded better. I can switch it back if you like.

kof: I would even add "may" because we haven't decided yet in which version and some people expressed negative input on that, so we might do it later

HenriBeck: Maybe ”one of the upcoming versions”

zelda-at-tempus: What about
>The HOC based API is deprecated as of v10 and may be removed in a future version.

kof: yes!

kof:
Thank you!

Issue #1211, opened 2019-10-10
[react-jss] wrong displayName of components given by withStyles HOC


dannystyleart:
__Expected behavior:__
Our jest snapshots ( using enzyme ) should not failing due components's displayName change.

__Describe the bug:__
Migrated from 8.6.1 to 10.0.0 our jest snapshots are failing. We have set up the theming context as the docs and the code is working fine in browser but the components's displayName have changed from Jss($componentName) to JssContextSubscriber.

![Failing tests](https://snipboard.io/n4G1o9.jpg)

__Codesandbox link:__ 
Please create a codesandbox.io with the issue. Make it as minimal as possible as this will help us find the bug quicker.
Here is our case in a nutshell: https://codesandbox.io/s/react-jss-withstyles-hoc-displayname-fails-04eob

__Versions (please complete the following information):__
 - jss: 10.0.0
- React: 16.10.2
- enzyme: 3.10.0
- enzyme-adapter-react-16: 1.15.0
- enzyme-to-json: 3.4.2
- jest: 24.9.0
 - OS [Linux deepin/Window 10/ Mac Os Catalina & Mojave]

HenriBeck:
You will just need to update your snapshot tests. This change is expected when upgrading to version 10 of react jss

dannystyleart:
@HenriBeck So complete loss of component naming is expected?

I have looked up how the withStyles HOC is doing name creation and it does kinda same as before.

My problem is that when snapshots are created from a component that contains jss styled components, all children has the same name.

If I'm changing a children in the markup and leaving the props the snapshot will no differ and not catches the change.

HenriBeck:
No, the `Jss(FormControl)` should be the child of `JssContextSubscriber` now as you can also see that in the diff

dannystyleart:
@HenriBeck It is.
We use the original HOC approach that is exported by default, our components are defined like in the codesandbox link above.

We found that the main reason that causes this is in the [withStyles.js:181](https://github.com/cssinjs/jss/blob/b1602dcd6a5514e29c899f23d11cb6346a1cf6aa/packages/react-jss/src/withStyles.js#L181)

`
    JssContextSubscriber.displayName = 'JssContextSubscriber'
`

Changing this line to fallowing, would solve the problem:
`
    JssContextSubscriber.displayName = `Jss(${displayName})`
`

PR #1214, opened 2019-10-14
Fix(withStyles): HOC return component with standard displayName

dannystyleart:
react-jss withStyles HOC returns component that has non-conventional naming.
Each styled component had the displayName "JssContextSubscriber" that makes hard jest snapshot testing (shallow)

#1211 

@HenriBeck Here is the possible solution I have mentioned before.

### DIFF
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,23 +1,23 @@
 {
   "dist/react-jss.js": {
-    "bundled": 169391,
-    "minified": 58384,
+    "bundled": 169398,
+    "minified": 58389,
     "gzipped": 19101
   },
   "dist/react-jss.min.js": {
-    "bundled": 112715,
-    "minified": 41775,
+    "bundled": 112722,
+    "minified": 41780,
     "gzipped": 14178
   },
   "dist/react-jss.cjs.js": {
-    "bundled": 27041,
-    "minified": 11651,
-    "gzipped": 3862
+    "bundled": 27048,
+    "minified": 11656,
+    "gzipped": 3863
   },
   "dist/react-jss.esm.js": {
-    "bundled": 26079,
-    "minified": 10816,
-    "gzipped": 3740,
+    "bundled": 26086,
+    "minified": 10821,
+    "gzipped": 3741,
     "treeshaked": {
       "rollup": {
         "code": 1841,
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,22 +1,22 @@
 {
   "dist/react-jss.js": {
-    "bundled": 169377,
-    "minified": 58382,
-    "gzipped": 19100
+    "bundled": 169391,
+    "minified": 58384,
+    "gzipped": 19101
   },
   "dist/react-jss.min.js": {
-    "bundled": 112701,
-    "minified": 41773,
-    "gzipped": 14177
+    "bundled": 112715,
+    "minified": 41775,
+    "gzipped": 14178
   },
   "dist/react-jss.cjs.js": {
-    "bundled": 27027,
-    "minified": 11649,
-    "gzipped": 3861
+    "bundled": 27041,
+    "minified": 11651,
+    "gzipped": 3862
   },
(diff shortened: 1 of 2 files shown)


HenriBeck:
@dannystyleart We still have the `Jss()` in the component.

https://github.com/cssinjs/jss/blob/f911113f554de5ab7e6d9bea359dbde577d986f6/packages/react-jss/src/withStyles.js#L50

I don't think to change this makes a lot of sense. This is then more an issue of `shallow` rendering. 

Also, `shallow` testing your styled component doesn't make a lot of sense, in my opinion, because you never render your actual component.

dannystyleart:
@HenriBeck I know that shallow not renders the styles.
However we use it as a primer check for structural checks and at that point we don't need to render styling.
By every exported component having the same name would ruin these tests ( yes is painful for 100k+ line of code ).
Yes, withStyles hoc itself does the correct naming but it is wrapped into the JssContextSubscriber which does not have correct naming.

HenriBeck:
> However we use it as a primer check for structural checks and at that point we don't need to render styling.

What kind of structural checks? In shallow render, you only render a single component. Before you would only render our component, and now only the `JssContextSubscriber`

dannystyleart:
@HenriBeck example:

We have button component ( supporting numerous props )
We have TextField component ( doing the same )

We have a complex component like a Form that contains Buttons,TextFields, and the form has states that affect visibility and/or props assignment of subcomponents.

We are testing those constellations by settings state/ props of the Form component.

Lets see some real examples:
We have a Modal component that can contain Modal part components namely( ModalHeader, ModalBody, ModalFooter ).
Before v10 we had this snapshot:
```
exports[`<Modal /> should should render correctly with subcomponents 1`] = `
<Jss(ModalPortalComponent)
  backdrop={true}
  backdropProps={
    Object {
      "className": "TEST-MODALBACKDROPTHEME",
      "onClick": [Function],
    }
  }
  isOpen={false}
  onKeyboardClose={[Function]}
  transitionSpeed={180}
>
  <div
    className="TEST-MODAL TEST-MODALTHEME"
    onClose={[Function]}
  >
    <Jss(ModalHeaderComponent)
      height={54}
    >
      Title
    </Jss(ModalHeaderComponent)>
    <Jss(ModalBodyComponent)>
      Content
    </Jss(ModalBodyComponent)>
    <Jss(ModalFooterComponent)
      height={50}
    >
      Footer
    </Jss(ModalFooterComponent)>
  </div>
</Jss(ModalPortalComponent)>
`;
```
After updating we got this:
```
exports[`<Modal /> should should render correctly with subcomponents 1`] = `
<JssContextSubscriber
  backdrop={true}
  backdropProps={
    Object {
      "className": "TEST-MODALBACKDROPTHEME",
      "onClick": [Function],
    }
  }
  isOpen={false}
  onKeyboardClose={[Function]}
>
  <div
    className="TEST-MODAL TEST-MODALTHEME"
    onClose={[Function]}
  >
    <JssContextSubscriber>
      Title
    </JssContextSubscriber>
    <JssContextSubscriber>
      Content
    </JssContextSubscriber>
    <JssContextSubscriber>
      Footer
    </JssContextSubscriber>
  </div>
</JssContextSubscriber>
`;
```

Without the displayName of the component the developer will not be able to recognize where and what component has been changed in an instant way.

I hope now the point is clear.

dannystyleart:
@HenriBeck any toughts on this? Otherwise how could we achieve the goal of testing via snapshots?

HenriBeck:
@dannystyleart With `shallow` rendering, you are not testing state and the component itself.

You only render one level deep the component. Any other components which are being used (including children) are never rendered just for your information. So changing some elements in `ModalHeaderComponent` would not make your above test fail.

Nevertheless, I think we can do this change as this is only about dev tools and will have no other effect on the code.

HenriBeck:
@dannystyleart please run `yarn build` again to update our snapshots again and then commit the changes

dannystyleart:
@HenriBeck thanks

dannystyleart:
@HenriBeck When it is likely to be published / released as non-alpha version?

HenriBeck:
Probably on the weekend. I would like to also include #1218 in the first patch version

Issue #1188, opened 2019-09-10
Cannot read property rules of undefined


vtereshyn:
Hi. Not sure how to reproduce locally or create a code snippet with a reproducible issue, but got this today 5 times on stage environment. It's a really weird and very important issue since I can't debug it
Hope you can help. 

Thank you in advance

![image](https://user-images.githubusercontent.com/32247411/64632869-a54fbc00-d402-11e9-99a4-86b24e5ea01a.png)

vtereshyn:
https://github.com/cssinjs/jss/pull/1189

vtereshyn:
https://github.com/cssinjs/jss/pull/1189

vtereshyn:
So, this issue brokes an application in production. Of course its my bad to use package alpha version in production, but here is a result.
React-jss is not ready for use in prod builds

HenriBeck:
Could you please include more info about your setup and under which conditions you are using `react-jss` so we can look into this?

vtereshyn:
@HenriBeck I couldn't create a codesandbox with a reproducible example, but I can try to describe:

I have a component with dynamic styles. This is FunctionComponent which uses `createUseStyles`. There I have a few CSS properties which dynamically change depends of Component props. 
I use this component on several pages. So when I return from one of the pages to another I got the error described above. In the console, I saw that the problem in `onUpdate` method or smth like that.

Also, before I had an issue with dynamical styles. My styles didn't update properly. So I needed a full page reload to update CSS. This is weird. 

But I'm looking forward to trying to help you in fixing these issues. 

Now I resolved these issue using different classnames depends on Component props.

HenriBeck:
The not updating of styles should be fixed in #1190

vtereshyn:
Cool, I'll update my jss versions and check. Thanks for the update.
Maybe my problem relates to #1190

HenriBeck:
We haven't released a version with that fix yet

vtereshyn:
@HenriBeck yes, I know. So I'm waiting for the new version and then update

felthy:
I've been having this problem too. It won't be fixed by #1190, it's a different problem.

The crash can be triggered by using a dynamic style inside a media query e.g.:

```javascript
  button: {
    '@media (min-width: 0px)': {
      color: () => 'red'
    }
  }
```

The crash occurs if you unmount a component that uses this sheet, while there are still other instances mounted.

Here's a codesandbox:

[...]

PR #1228, opened 2019-11-06
Fix: Cannot read property rules of undefined #1189

matteoferigo:
https://github.com/cssinjs/jss/issues/1188

Prevent onUpdate execution when `rule` instance doesn't exist yet.

## Todo
- [ ] Add tests if possible
  - can't execute tests following `contributing.md` (seems outdated)
  - it's a simple bugfix
- [ ] Add documentation
  - it's a simple bugfix

### DIFF
--- a/packages/jss/src/RuleList.js
+++ b/packages/jss/src/RuleList.js
@@ -180,7 +180,10 @@ export default class RuleList {
   /**
    * Execute plugins, update rule props.
    */
-  onUpdate(data: Object, rule: Rule, options?: Object = defaultUpdateOptions) {
+  onUpdate(data: Object, rule?: Rule, options?: Object = defaultUpdateOptions) {
+    // Should be undefined during hot-reloading
+    if (!rule) return
+
     const {
       jss: {plugins},
       sheet


packages/jss/src/RuleList.js
```diff
@@ -180,7 +180,10 @@ export default class RuleList {
   /**
    * Execute plugins, update rule props.
    */
-  onUpdate(data: Object, rule: Rule, options?: Object = defaultUpdateOptions) {
+  onUpdate(data: Object, rule?: Rule, options?: Object = defaultUpdateOptions) {
+    // Should be undefined during hot-reloading
+    if (!rule) return
```
HenriBeck: I guess this makes sense as we are typing it as optional anyway. Not sure how flow would even allow something like this

kof: Flow will have a problem here, rule isn't optional. We need a clear reproduction of this case to understand how rule can be undefined and we need to fix it far before, so that it doesn't come to this state.

kof:
@matteoferigo are you planing to work more on this? Just so we can close it if not and people don't wait.

matteoferigo:
I don't know what triggers this error, I tried to quickly create a reproducible test case but it's not simple (in my situation, it happens during hot-reloading within a Next.js project, when JSS is trying to update an unmounted page).
This issue doesn't affect the production environment (because there is no hot-reloading)… so I close it and maybe someday I will re-open it when I will find what is triggering this issue

Issue #1202, opened 2019-10-01
[react-jss] Not working HMR with RHL (v10.0.0+)


BEGEMOT9I:
__Expected behavior:__
HMR must working after styles changing.

__Describe the bug:__
An application with `HMR` `(react-hot-loader, react-jss-hmr)` does not work in place with new version of `react-jss`:
![Снимок экрана 2019-10-01 в 17 35 32](https://user-images.githubusercontent.com/11138075/65972115-06066d80-e472-11e9-9f78-0362bbde1fa5.png)
In this case, it seems, the problem is that the HOC `withStyles` returns a `JssContextSubscriber` instead of a HOC-function.

__Versions (please complete the following information):__
 - jss: 10.0.0-alpha.25
 - react-jss: 10.0.0-alpha.27
 - Browser [e.g. chrome, safari]: all
 - OS [e.g. Windows, macOS]: all
Feel free to add any additional versions which you may think are relevant to the bug.

kof:
Fix https://github.com/cssinjs/jss/pull/1237 is going to be released with 10.0.1, feel free to try out

BEGEMOT9I:
@kof #1237 does not fix this problem, since it is typical only for the HOC. However, the #1230 just allows you to use the HOC to the fullest without a plugin. Or are you going to rewrite the HOC so that it uses the hook functionality?

kof:
Alright, thanks for testing. Reopening then. Yeah I think we should refactor HOC to use hooks interface.

PR #1230, opened 2019-11-09
[react-jss] added fix for HMR and RHL

BEGEMOT9I:
Corrects the generation of component styles during HMR. Eliminates the need for [react-jss-hmr](https://github.com/cssinjs/react-jss-hmr) package.

https://github.com/cssinjs/jss/issues/1202

Fix

## Todo
- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage
 lerna-debug.log
 yarn-error.log
 .idea
+.vscode
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,23 +1,23 @@
 {
   "dist/react-jss.js": {
-    "bundled": 169398,
-    "minified": 58389,
-    "gzipped": 19101
+    "bundled": 169602,
+    "minified": 58465,
+    "gzipped": 19117
   },
   "dist/react-jss.min.js": {
-    "bundled": 112722,
-    "minified": 41780,
-    "gzipped": 14178
+    "bundled": 112926,
+    "minified": 41856,
+    "gzipped": 14195
   },
   "dist/react-jss.cjs.js": {
-    "bundled": 27048,
-    "minified": 11656,
-    "gzipped": 3863
+    "bundled": 27244,
+    "minified": 11732,
+    "gzipped": 3882
   },
   "dist/react-jss.esm.js": {
-    "bundled": 26086,
-    "minified": 10821,
-    "gzipped": 3741,
+    "bundled": 26282,
+    "minified": 10897,
+    "gzipped": 3757,
     "treeshaked": {
       "rollup": {
         "code": 1841,
--- a/packages/react-jss/src/withStyles.js
+++ b/packages/react-jss/src/withStyles.js
@@ -20,6 +20,7 @@ import {manageSheet, unmanageSheet} from './utils/managers'
 import getSheetClasses from './utils/getSheetClasses'
 
 interface State {
+  generationIndex: number;
   dynamicRules: ?DynamicRules;
   sheet: ?StyleSheet;
   classes: {};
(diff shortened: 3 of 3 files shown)


BEGEMOT9I:
@HenriBeck Hi. Could you see the travis, please.

HenriBeck:
> @HenriBeck Hi. Could you check travis, please.

@BEGEMOT9I could you please explain how this would fix HMR with `withStyles`? I thought we would need to recreate the sheet in `componentDidUpdate`

BEGEMOT9I:
@HenriBeck Sorry, my mistake, I forgot to add the manager updating into the cDU. We can't use the prevState in comparison, since then managers will be updated twice.

kof:
This PR looks promising, @BEGEMOT9I please check it out, since it will probably fix that issue
https://github.com/cssinjs/jss/pull/1237

BEGEMOT9I:
@kof this PR fixes only updating styles in the HOC, but not in the hook. A changes in the hook does not correlate with current changes.

kof:
I think we should close this PR and make HOC use hooks

BEGEMOT9I:
@kof. Okay. I could try to write a version with using hook on holidays. But I'm not sure if this realization will be as productive as the current one.

kof:
You mean using the createUseStyles we already have, should be only less perf because it is a HOC, should be streight forward to port

BEGEMOT9I:
@kof I wrote a functional that adapts the use of a hook into a HOC. Along the way, I conducted a [little testing](https://github.com/BEGEMOT9I/test-react-jss/) (rather for myself). The final version is here: https://github.com/BEGEMOT9I/test-react-jss/blob/master/src/lib/utils/react-jss/withStyles.tsx. There are some typing issues in TS that need to be fixed.

kof:
@BEGEMOT9I we need something mergeable that migrates the HOC to hooks.

BEGEMOT9I:
@kof Yeah, I understand. This is the preview version. If what is written can be put into the project, then I can create a pr, but a little later

kof:
I can't tell unless I see a diff and tests passing

kof:
also would need to run a bench to make sure it's not slower than the original version

BEGEMOT9I:
@kof I wroted base version, but it does not pass some tests. Can you see them and express some thoughts? Because the test of `should not inject theme with themed classes`, for example, should not pass in principle, since in the hock there is a check for the need to throw the topic by whether the variable `styles` an object (https://github.com/cssinjs/jss/blob/master/packages/react-jss/tests/withStylesTheming.js#L43 and https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/withStyles.js#L39). I will try to fix it soon.

BEGEMOT9I:
@kof Fixed all cases where tests fell, with the exception of `should not inject theme with themed classes` (see comment before).

BEGEMOT9I:
`should not inject theme with themed classes` - fixed (my mistake). Removed `JssProvider`, which should increase the speed of rendering the HOC.

BEGEMOT9I:
@HenriBeck some progress?

kof:
@BEGEMOT9I are you still interested in landing this PR?  How can I help?

BEGEMOT9I:
@kof Yeah. This is really a very necessary functionality. While it is not in the library, we have to use our own implementation of the hook with support for HMR.

kof:
So what steps do we need to accomplish this?

BEGEMOT9I:
I don't know. Maybe you need to test the performance (which you wrote about). And so, we already use this version of the HOC in our company at production.

kof:
Or maybe I should communicate which tests I was using so far, so that you can run them too

BEGEMOT9I:
Yes, you can. I tested the speed of rendering and re-rendering of 1000 elements, but it was with the dev tools

kof:
It's a good test in the sense of E2E, another one is https://cssinjs.github.io/react-native-web/benchmarks/

We have some more micro benchmarks https://github.com/cssinjs/jss/tree/master/packages/jss/benchmark/tests but it won't help in this case

BEGEMOT9I:
@kof Ok, I will try the first version

PR #1235, opened 2019-11-25
fix(typescript): compat with non-dom consumers

Akuukis:

 Problem: jss doesn't depend on browser env but the types do.
 Workaround: Copy/paste subset of interfaces from `dom.d.ts` (TS v3.7.2):
 - `CSSStyleRule`
 - `StyleSheet`
 - `CSSRule`

## Post Scriptum
Pre-commit hook failed, and following readme didn't solve that. Here's the excerpt:
```
$ git commit -m'fix(typescript): compat with non-dom consumers'
 ✔ Running tasks for **/*.{js,md,json,ts}
 ↓ Running tasks for **/*.{js,md} [skipped]
   → No staged files match **/*.{js,md}
yarn run v1.15.2
$ yarn check:flow && yarn check:ts && yarn check:snapshots && yarn check:lint
$ flow check --max-warnings=0
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ packages/css-jss/src/createCss.js:3:20

Cannot resolve module jss-preset-default.

     1│ // @flow
     2│ import {create as createJss} from 'jss'
     3│ import preset from 'jss-preset-default'
     4│ import type {Jss} from 'jss'
     5│ // eslint-disable-next-line no-unused-vars
     6│ import type {Css, StyleArg} from './types'
```

### DIFF
--- a/packages/jss/src/dom-monkeypatch.d.ts
+++ b/packages/jss/src/dom-monkeypatch.d.ts
@@ -1,0 +1,638 @@
+/**
+ * Problem: jss doesn't depend on browser env but the types do.
+ * Workaround: Copy/paste subset of interfaces from `dom.d.ts` (TS v3.7.2):
+ * - `CSSStyleRule`
+ * - `StyleSheet`
+ * - `CSSRule`
+ *
+ * Note: Typescript interfaces nicely merge when full dom types are present, but `declare var` doesn't.
+ */
+
+/** A single CSS rule. There are several types of rules, listed in the Type constants section below. */
+interface CSSRule {
+  cssText: string
+  readonly parentRule: CSSRule | null
+  readonly parentStyleSheet: CSSStyleSheet | null
+  readonly type: number
+  readonly CHARSET_RULE: number
+  readonly FONT_FACE_RULE: number
+  readonly IMPORT_RULE: number
+  readonly KEYFRAMES_RULE: number
+  readonly KEYFRAME_RULE: number
+  readonly MEDIA_RULE: number
+  readonly NAMESPACE_RULE: number
+  readonly PAGE_RULE: number
+  readonly STYLE_RULE: number
+  readonly SUPPORTS_RULE: number
+  readonly UNKNOWN_RULE: number
+  readonly VIEWPORT_RULE: number
+}
+
+/** A CSSRuleList is an (indirect-modify only) array-like object containing an ordered collection of CSSRule objects. */
+interface CSSRuleList {
+  readonly length: number
+  item(index: number): CSSRule | null
+  // [index: number]: CSSRule  // Index signatures does not merge nicely.
+}
+
+/** An object that is a CSS declaration block, and exposes style information and various style-related methods and properties. */
+interface CSSStyleDeclaration {
+  alignContent: string
+  alignItems: string
+  alignSelf: string
+  alignmentBaseline: string
+  animation: string
+  animationDelay: string
+  animationDirection: string
+  animationDuration: string
+  animationFillMode: string
+  animationIterationCount: string
+  animationName: string
+  animationPlayState: string
+  animationTimingFunction: string
+  backfaceVisibility: string
+  background: string
+  backgroundAttachment: string
+  backgroundClip: string
+  backgroundColor: string
+  backgroundImage: string
+  backgroundOrigin: string
+  backgroundPosition: string
+  backgroundPositionX: string
+  backgroundPositionY: string
+  backgroundRepeat: string
+  backgroundSize: string
+  baselineShift: string
+  blockSize: string
+  border: string
+  borderBlockEnd: string
+  borderBlockEndColor: string
+  borderBlockEndStyle: string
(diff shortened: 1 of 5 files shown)


Akuukis:
bump - anything I can do to get this solved?

Akuukis:
@kof @HenriBeck Happy new Year! Anything I can help with?

This PR is blocking us to upgrade dependencies for our SSR backend codebase. I'd love to have motion on this.

HenriBeck:
@Akuukis 

Wouldn't an easier fix to be adding `dom` to your `lib` in your `tsconfig.json`?

Akuukis:
@HenriBeck 

JSS library doesn't depend on `dom` environment, and thus types shouldn't too. As it stands, JSS may share with `dom` 600 out of 20,000 LOC and imports/reuses those types for convenience, but that has many subtle problems:
1. `dom` types are a external undocumented dependency on their own
2. with `dom` it forces consumer to lie to itself about in what environment it's code is running
3. with `dom` other 19,400 LOC pollute global namespace
4. with `dom` other 19,400 LOC puts unnecessary pressure on typescript server
5. jss may not use all properties/methods of those types - then to require a full type is unnecessary burden on consumer.
6. jss may not populate all properties/methods of those types when returned - then types would not be accurate.

This PR already fixes first 4 problems for all non-dom consumers. Furthermore it's a first step forward to review and address last 2 problems later.

If every non-`dom` consumer just adds `dom` to `tsconfig.json` then they will have first 4 problems, and there's no motion to review and address last 2 problems later.

HenriBeck:
@Akuukis What do you think about a separate entry point then?

Because technically, the `DomRenderer` should also not be used on the server at all/available.

kof:
The problem is that JSS has both dom and non-dom rendering and it decides automatically what it uses, so from the user perspective, it could potentially decide to use both. I think some sort of indication from the user like a separate entry point would be a possible way.

kof:
I wonder though what if you are using an isomorphic setup, so you use the same logic and entry points for both server and client.

Akuukis:
I think that separate entry points won't make a difference - `dom` types are put in global scope, and whenever library is imported one or other way, or not imported at all but just available in dependency tree, those types will be there.

If non-dom rendering API is a sub-set of `dom` types then I think that this PR is the best approach:
- for non-dom renderer, only the provided types will be present and consumer will see accurate sub-set.
- for dom renderer, the provided types will merge with global `dom` types, and consumer will see a accurate full set.

kof:
How is react doing it with react-dom?

Akuukis:
It has two entry-points - root for `dom` consumers and `react-dom/server` for nodeJS consumers. 
- at root there's no reference to `dom` types so it's a undocumented external dependency (a `/// <reference lib="DOM" />` tag would solve that)
- at `react-dom/server` there's a little monkey-patch for `Node.ReadableStream` at start of file. That monkey-patch is available only when `react-dom/server` is imported.

Type-wise, those are two packages in a one - both entrypoints don't import anything in common.

kof:
So we should do that as well, right?

HenriBeck:
I think this would be the correct solution to just expose a different entry point for the server where the `DomRenderer` is not exported

Akuukis:
Oh, I missed that `DomRenderer` is not used on server at all. Yes then let's do like that!

Would you create a PR? I don't have a clear idea how to split it best.

HenriBeck:
@kof Let's go forward with this for now. I think the change with the different `server` entry point we should put a little bit more thought on the structure

packages/jss/src/dom-monkeypatch.d.ts
```diff
@@ -0,0 +1,638 @@
...
+
+/** A CSSRuleList is an (indirect-modify only) array-like object containing an ordered collection of CSSRule objects. */
+interface CSSRuleList {
+  readonly length: number
+  item(index: number): CSSRule | null
+  // [index: number]: CSSRule  // Index signatures does not merge nicely.
+}
+
+/** An object that is a CSS declaration block, and exposes style information and various style-related methods and properties. */
+interface CSSStyleDeclaration {
```
kof: Is this for server-side rendering? Is react-dom doing that as well? I think ultimately this creates a maintenance burden because now we will have to maintain those types.

We def. need a way to avoid this. If anything types from styles should be generated as from mdn-data package and we have a separate issue for this.

kof:
> I think the change with the different server entry point we should put a little bit more thought on the structure

I think we need to do this to properly solve this problem.

kof:
If this solution generates more issues, I would rather have this original issue open than 10 new once.

kof:
I am closing this due to inactivity, if you wish to make it work, lets talk.

PR #1237, opened 2019-11-28
refactor createUseStyles to work with react-hot-loader

bs85:
`isFirstRender` is broken when using react-hot-loader, as it unmounts and remounts the components, but preserves hook state all the way through, including `isFirstRender`.

The `removeDynamicRules` hook runs first, and an error is thrown once `updateDynamicRules` runs as the rules have been removed.

With those changes everything should run in the proper order.

### DIFF
--- a/packages/react-jss/src/createUseStyles.js
+++ b/packages/react-jss/src/createUseStyles.js
@@ -20,13 +20,6 @@ const useEffectOrLayoutEffect = isInBrowser ? React.useLayoutEffect : React.useE
 
 const noTheme = {}
 
-const reducer = (prevState, action) => {
-  if (action.type === 'updateSheet') {
-    return action.payload
-  }
-  return prevState
-}
-
 const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions<Theme> = {}) => {
   const {index = getSheetIndex(), theming, name, ...sheetOptions} = options
   const ThemeContext = (theming && theming.context) || DefaultThemeContext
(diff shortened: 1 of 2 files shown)


kof:
Tests are passing, ignore the CI, it's a permissions probl. Looks good at the first glance. Need to check if we have all tests  for cases like
1. 2 elements of the same component reusing the same style sheet if there are no dynamic rules
2. 2 elements are reusing the same static sheet and dynamic rules are created and updatable
3. 1 of 2 elements unmounts - static sheet that is in use by the other element is still there, dynamic rules used only by that element are removed.

kof:
@HenriBeck @eps1lon do you see any problems with this approach?

kof:
@HenriBeck do you see anything wrong with this PR? I would merge it tomorrow otherwise.

packages/react-jss/src/createUseStyles.js
```diff
@@ -42,111 +35,75 @@ const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions
...
-
-      return {
-        sheet,
-        dynamicRules,
-        classes: classes || {}
-      }
-    })
-
-    useEffectOrLayoutEffect(
+    const [sheet, dynamicRules] = React.useMemo(
```
HenriBeck: If I remember correctly there was a problem with using `React.useMemo` as it can be rerun by React at a given point of time.

I think @eps1lon knew more about this.

packages/react-jss/src/createUseStyles.js
```diff
@@ -42,111 +35,75 @@ const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions
...
-        if (state.sheet && state.dynamicRules && !isFirstMount.current) {
-          updateDynamicRules(data, state.sheet, state.dynamicRules)
+        if (sheet && dynamicRules && !isFirstMount.current) {
+          updateDynamicRules(data, sheet, dynamicRules)
         }
       },
       [data]
     )
 
     useEffectOrLayoutEffect(
```
HenriBeck: Can't we just move this above the `updateDynamicRules` hook?

packages/react-jss/src/createUseStyles.js
kof: I think you are talking about the theoretical situation in async mode, not sure if it was made real.

packages/react-jss/src/createUseStyles.js
kof: That cleanup depends on the sheet argument, the one above on data.

kof: Seems correct to me

kof:
@bs85 going to release it together with one more important bugfix that I have been working on lately

bs85:
Great, thank you!

PR #1239, opened 2019-12-08
feat: export defaultJss and add createJss

anyexinglu:
1.why export `defaultJss`?

Sometimes, we need to use the type from jss, such as `import { StyleSheet, JssOptions } from 'jss'` 

But we cannot `npm install jss`, since it will brings two versions of jss (one from jss itself, one from react-jss's dependency)

The solution is to export the original jss to avoid two versions

2.why add `createJss`?

Sometime, we want to customize the `create(preset())`, such as `create({...preset(), createGenerateId})`, but the jss from react-jss/src/jss.js doesn't support.

So add a function named `createJss`, the usage is:
```
import { createJss } from 'react-jss';

// `createGenerateId` is a customize function to solve the 'Server-side rendering' problem within a component library, just like material design.
//  In which, we cannot use `<JssProvider generateId={generateId}><MyApp /></JssProvider>` solution.
const jss = createJss({
  createGenerateId
})
```

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/react-jss/src/jss.js
+++ b/packages/react-jss/src/jss.js
@@ -1,5 +1,12 @@
 // @flow
-import {create} from 'jss'
+import {create, type JssOptions} from 'jss'
 import preset from 'jss-preset-default'
+
+export const createJss = (options?: Partial<JssOptions>) => {
+  return create({
+    ...preset(),
+    ...(options || {})
+  })
+}
 
 export default create(preset())
--- a/packages/react-jss/src/index.js
+++ b/packages/react-jss/src/index.js
@@ -1,15 +1,16 @@
 // @flow
+import defaultJss from 'jss'
 import withStyles from './withStyles'
 
 export {ThemeProvider, withTheme, createTheming, useTheme} from 'theming'
 export {default as createUseStyles} from './createUseStyles'
 export {default as JssProvider} from './JssProvider'
-export {default as jss} from './jss'
+export {default as jss, createJss} from './jss'
 export {SheetsRegistry, createGenerateId} from 'jss'
 export {default as JssContext} from './JssContext'
 export {default as styled} from './styled'
 export {default as jsx, create as createJsx} from './jsx'
-export {withStyles}
+export {withStyles, defaultJss}
 
 // Kept for backwards compatibility.
 export default withStyles


packages/react-jss/src/jss.js
```diff
@@ -1,5 +1,12 @@
 // @flow
-import {create} from 'jss'
+import {create, type JssOptions} from 'jss'
 import preset from 'jss-preset-default'
 
+export const createJss = (options?: Partial<JssOptions>) => {
```
kof: We could export the `create` function from JSS and the default presets, so one can combine both of them

kof: If wee do this, we don't really need to export `defaultJss` from JSS, since one could just use the create function.

kof: So I would just do this:

```js
export {create as createJss} from 'jss'
export {default as preset} from 'jss-preset-default'
```

HenriBeck:
> But we cannot npm install jss, since it will brings two versions of jss (one from jss itself, one from react-jss's dependency)

If `react-jss` and `jss` itself are on the same version since version 10, they should be deduplicated in the resulting bundle, in my opinion.

kof:
@HenriBeck there is always a danger that user doesn't update their version but update react-jss and there will be duplication, so it's safer if they can import it from react-jss package

anyexinglu:
> > But we cannot npm install jss, since it will brings two versions of jss (one from jss itself, one from react-jss's dependency)
> 
> If `react-jss` and `jss` itself are on the same version since version 10, they should be deduplicated in the resulting bundle, in my opinion.

Sometimes jss and react-jss will not be deduplicated.
For example:

- Currently, my package.json and yarn.lock are:
```
// package.json
"dependencies": {
  "jss": "^10.0.0",
  "react-jss": "^10.0.0",
}
```
```
// yarn.lock
react-jss@^10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/react-jss/-/react-jss-10.0.0.tgz#fdde48eabe85c807a0729518e13eae25fc46e59a"
  integrity sha1-/d5I6r6FyAegcpUY4T6uJfxG5Zo=
  dependencies:
    "@babel/runtime" "^7.3.1"
    "@emotion/is-prop-valid" "^0.7.3"
    css-jss "10.0.0"
    hoist-non-react-statics "^3.2.0"
    is-in-browser "^1.1.3"
    jss "10.0.0"
    jss-preset-default "10.0.0"
    prop-types "^15.6.0"
    shallow-equal "^1.2.0"
    theming "3.2.0"
    tiny-warning "^1.0.2"

css-jss@10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/css-jss/-/css-jss-10.0.0.tgz#58fc9d19593ddcf4857855743c6066a113fefd17"
  integrity sha1-WPydGVk93PSFeFV0PGBmoRP+/Rc=
  dependencies:
    "@babel/runtime" "^7.3.1"
    jss "10.0.0"
    jss-preset-default "10.0.0"

jss@10.0.0, jss@^10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/jss/-/jss-10.0.0.tgz#998d5026c02accae15708de83bd6ba57bac977d2"
  integrity sha1-mY1QJsAqzK4VcI3oO9a6V7rJd9I=
  dependencies:
    "@babel/runtime" "^7.3.1"
    csstype "^2.6.5"
    is-in-browser "^1.1.3"
    tiny-warning "^1.0.2"
```

- I run `yarn add jss`.
- Then, my package.json and yarn.lock are:
```
// package.json
"dependencies": {
  "jss": "^10.0.3",
  "react-jss": "^10.0.0",
}
```
```
// yarn.lock
react-jss@^10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/react-jss/-/react-jss-10.0.0.tgz#fdde48eabe85c807a0729518e13eae25fc46e59a"
  integrity sha1-/d5I6r6FyAegcpUY4T6uJfxG5Zo=
  dependencies:
    "@babel/runtime" "^7.3.1"
    "@emotion/is-prop-valid" "^0.7.3"
    css-jss "10.0.0"
    hoist-non-react-statics "^3.2.0"
    is-in-browser "^1.1.3"
    jss "10.0.0"
    jss-preset-default "10.0.0"
    prop-types "^15.6.0"
    shallow-equal "^1.2.0"
    theming "3.2.0"
    tiny-warning "^1.0.2"

css-jss@10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/css-jss/-/css-jss-10.0.0.tgz#58fc9d19593ddcf4857855743c6066a113fefd17"
  integrity sha1-WPydGVk93PSFeFV0PGBmoRP+/Rc=
  dependencies:
    "@babel/runtime" "^7.3.1"
    jss "10.0.0"
    jss-preset-default "10.0.0"

jss@10.0.0, jss@^10.0.0:
  version "10.0.0"
  resolved "http://bnpm.corp.yiran.com/jss/-/jss-10.0.0.tgz#998d5026c02accae15708de83bd6ba57bac977d2"
  integrity sha1-mY1QJsAqzK4VcI3oO9a6V7rJd9I=
  dependencies:
    "@babel/runtime" "^7.3.1"
    csstype "^2.6.5"
    is-in-browser "^1.1.3"
    tiny-warning "^1.0.2"

jss@^10.0.3:
  version "10.0.3"
  resolved "https://bnpm.pdd.net/jss/-/jss-10.0.3.tgz#5c160f96aa8ce8b9f851ee0b33505dcd37f490a4"
  integrity sha1-XBYPlqqM6Ln4Ue4LM1BdzTf0kKQ=
  dependencies:
    "@babel/runtime" "^7.3.1"
    csstype "^2.6.5"
    is-in-browser "^1.1.3"
    tiny-warning "^1.0.2"
```

There will be two versions of jss in my yarn.lock and node_modules

If it is intentional? Personally,  I thinks there are two ways:

- jss should be devDependency and peerDependency in react-jss, but not dependency. (Since the users of this package must already has a version of jss)
- as this pull request does, export jss as defaultJss from react-jss, so users can `yarn add react-jss && yarn remove jss`

kof:
hey @anyexinglu I would love this problem solved, but I don't want to increase amount of APIs more than necessary.

The reason I proposed

```js
export {create as createJss} from 'jss'
export {default as preset} from 'jss-preset-default'
```

Is because you can solve the same dependencies problem you have, but using interfaces we already have in jss and the default present, basically only exposing them from this package, without introducing new functions.


If you are interesting to amend the changes you made to what I suggest, I am happy to merge it.

Otherwise, I am gonna have to close this PR for the time being.

Issue #1240, opened 2019-12-13
[jss-isolate] Reseting vendor-suffixed selectors not working due to resulting css rule being invalid 


shamakhov:
__Expected behavior:__
Resulting reset rules are valid

__Describe the bug:__
I am using Material UI widgets with vendor selectors like `::-ms-input-placeholder`, `-moz-placeholder`, etc, and applying isolate plugin results in this errors in Chrome:
<details> 
  <summary>
Warning: [JSS] Failed to execute 'insertRule' on 'CSSStyleSheet': Failed to parse the rule</summary> <pre>'.MuiPaper-root,
.MuiPaper-rounded,
.MuiPaper-elevation0,
.MuiPaper-elevation1,
.MuiPaper-elevation2,
.MuiPaper-elevation3,
.MuiPaper-elevation4,
.MuiPaper-elevation5,
.MuiPaper-elevation6,
.MuiPaper-elevation7,
.MuiPaper-elevation8,
.MuiPaper-elevation9,
.MuiPaper-elevation10,
.MuiPaper-elevation11,
.MuiPaper-elevation12,
.MuiPaper-elevation13,
.MuiPaper-elevation14,
.MuiPaper-elevation15,
.MuiPaper-elevation16,
.MuiPaper-elevation17,
.MuiPaper-elevation18,
.MuiPaper-elevation19,
.MuiPaper-elevation20,
.MuiPaper-elevation21,
.MuiPaper-elevation22,
.MuiPaper-elevation23,
.MuiPaper-elevation24,
.MuiTable-root,
.MuiTable-root caption,
.MuiTable-stickyHeader,
.MuiTableHead-root,
.MuiTableRow-root,
.MuiTableRow-root.Mui-selected,
.MuiTableRow-root.MuiTableRow-hover:hover,
.Mui-selected,
.MuiTableRow-hover,
.MuiTableRow-head,
.MuiTableRow-footer,
.MuiTableCell-root,
.MuiTableCell-head,
.MuiTableCell-body,
.MuiTableCell-footer,
.MuiTableCell-sizeSmall,
.MuiTableCell-sizeSmall:last-child,
.MuiTableCell-sizeSmall.MuiTableCell-paddingCheckbox,
.MuiTableCell-sizeSmall.MuiTableCell-paddingCheckbox:last-child,
.MuiTableCell-sizeSmall.MuiTableCell-paddingCheckbox > *,
.MuiTableCell-paddingCheckbox,
[...]

PR #1241, opened 2019-12-13
Validate each isolate selector before adding it to resulting rule

shamakhov:
#1240 
## Todo

- [ ] Add tests if possible
- [ ] Add documentation


kof:
That's an interesting problem, these selectors are not invalid if I understand the case correctly, it's just so they are vendor prefixed and insertRule api throws if the vendor used is not supported by that browser.

Doing this at runtime for every selector is probably way too much overhead. It actually sounds like we need to make sure that we only render vendor prefixed selectors in the browser that understands that prefix. 

One way to do it is to ask users to use vendor-prefixer package for prefix detection and only supply

kof:
Closing due to inactivity.

Issue #1246, opened 2019-12-24
css-jss package missing ts types


chrisands:
__Is your feature request related to a problem? Please describe.__
When I open project with `react-jss` I get an error `File '*/node_modules/css-jss/src/index.d.ts' not found.ts` in `jsconfig.json`

__Describe the solution you'd like__
Add `index.d.ts` file and copy types from flow `types.js`

__Are you willing to implement it?__
Yes

chrisands:
Merged #1247

PR #1247, opened 2019-12-24
Add TS types for css-jss packages

chrisands:
#1246 

Add typescript types for css-jss package

### DIFF
--- a/packages/css-jss/src/index.d.ts
+++ b/packages/css-jss/src/index.d.ts
@@ -1,0 +1,5 @@
+export type Style = {} | null | void | string
+
+export type StyleArg = Style | Array<Style>
+
+export type Css = (...args: StyleArg[]) => string
--- a/packages/css-jss/src/index.d.ts
+++ b/packages/css-jss/src/index.d.ts
@@ -1,8 +1,8 @@
-import {Jss, Style} from 'jss'
+import {Jss, JssStyle} from 'jss'
 
-export {Style}
+export {JssStyle}
 
-export type StyleArg = Style | Array<Style>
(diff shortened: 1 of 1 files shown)


packages/css-jss/src/index.d.ts
```diff
@@ -0,0 +1,5 @@
+export type Style = {} | null | void | string
```
kof: you should be able to use https://github.com/cssinjs/jss/blob/master/packages/jss/src/index.d.ts#L4

chrisands: Thank you, updated code

packages/css-jss/src/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import {Style} from 'jss'
+
+export {Style}
+
+export type StyleArg = Style | Array<Style>
+
+export type Css = (...args: StyleArg[]) => string
```
HenriBeck: The definitions for `createCss` and the default export are missing here. These definitions will not work by it's own

chrisands: Done

packages/css-jss/src/index.d.ts
```diff
@@ -0,0 +1,15 @@
+import {Jss, Style} from 'jss'
```
HenriBeck:
```suggestion
import {Jss, JssStyle} from 'jss'
```

I think this is named `JssStyle` now

chrisands: Rebased to master branch and updated imports

kof:
thanks @chrisands and @HenriBeck

PR #1218, opened 2019-10-20
Update typescript definitions

HenriBeck:
__What would you like to add/fix?__

Adjusts the typescript definitions

__Corresponding issue:__
#1215
#1210
#1203
#1199

PR #1250, opened 2019-12-28
Fix #1199. Type alias circularly references

### LINKED ISSUE #1199: CSS Properties TypeScript definitions
__Expected behavior:__
Being able to auto-complete on CSS properties using the `createUseStyles`.

__Describe the bug:__
Not being able to do the auto-complete of CSS properties because of a `// @ts-ignore` block of the `Style` definition of: https://github.com/cssinjs/jss/blob/master/packages/jss/src/index.d.ts#L3

If I remove the `// @ts-ignore` block, I get an error from TypeScript about a circular dependency: `Type alias 'Style' circularly references itself.`

I am not sure wether or not to remove the circularly `Style` reference, or rework it to be:

```ts
export type Style = css.StandardProperties<string | number | (() => string | number)> & {
  [key: string]: { [key: string]: Style } | string | number
}
```

__Codesandbox link:__ 
https://codesandbox.io/s/magical-blackwell-04it0

 - jss:
 - Browser [e.g. chrome, safari]:
 - OS [e.g. Windows, macOS]:

hackwaly:

- #1199

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -1,9 +1,9 @@
 import * as css from 'csstype'
 
-// @ts-ignore
-export type Style = css.StandardProperties<string | number | (() => string | number)> & {
-  [key: string]: Style | string | number
+type StyleP<T> = css.StandardProperties<string | number | (() => string | number)> & {
+  [key: string]: T | string | number;
 }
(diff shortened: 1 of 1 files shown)


HenriBeck:
Thanks for the PR, I applied the fix in #1218

Issue #1256, opened 2020-01-08
[react-jss] getSheetOptions argument options replaced by default 


yeliex:
__Expected behavior:__
`createStyleSheet ` could get options.generateId correctly

__Describe the bug:__
https://github.com/cssinjs/jss/blob/17f54cfbe63009838c48fd2a12751302698ef200/packages/react-jss/src/utils/sheets.js#L34-L57 input options replace default options correctly

__Versions (please complete the following information):__
 - jss: 10.0.3

yeliex:
@kof

kof:
Can you please explain in a greater detail? Ideally use the structure in the template and evtl a codesandbox example of the bug that reproduces the problem.

yeliex:
@kof 
as type defined https://github.com/cssinjs/jss/blob/17f54cfbe63009838c48fd2a12751302698ef200/packages/jss/src/index.d.ts#L132-L140
`withStyles` method can accept an option with prop `generateId`

but in https://github.com/cssinjs/jss/blob/17f54cfbe63009838c48fd2a12751302698ef200/packages/react-jss/src/utils/sheets.js#L49-L57 , the property would be replaced by `options.context.generateId `, and the passed one would be ignored

For example, 
```js
import injectSheet from 'react-jss'
const StyledComponent = injectSheet(styles, {
    generateId: (rule) => rule.key
});
```
the passed generateId would be ignored

yeliex:
```js
return { 
     ...options.sheetOptions, 
     index: options.index, 
     meta, 
     classNamePrefix, 
     link, 
     generateId: options.context.generateId // undefined
   }  // => {generateId: undefined}
```

kof:
Understood, cc @HenriBeck

yeliex:
The similar code snip in the old version:

https://github.com/cssinjs/jss/blob/v9.7.0/src/Jss.js#L84-L90

kof:
gonna be released with the next version, subscribe to releases!

yeliex:
@kof Can you publish a new release, we want to update `jss` in our component library but now processed because of the issue.

PR #1257, opened 2020-01-08
fix: react-jss/src/utils/sheets.js getSheetOptions use options.genera…

yeliex:
#1256 
react-jss/src/utils/sheets.js getSheetOptions use options.generateId first

## Todo

- [x] Add tests if possible
- [ ] Add documentation


kof:
can you please add a test?

yeliex:
@kof @HenriBeck Test added

I export getSheetOptions for testing the change only

kof:
@yeliex thanks for doing it. The problem with this kind of tests - they don't tell us why it is testing it, it doesn't describe the use case. In addition, it tests internal implementation detail.

Tests should be written in the way the API is actually used by the user, so that on the long run we understand better why things work the way it work.

yeliex:
@kof I have tried to write tests for exported APIs at first, but there have some troubles.

1. `createStyleSheet` is the main method of `sheet.js` and used by `createUseStyle` and `wthStyle`, so it is the best to write tests for it, but if want to do that, I have to create an instance for context and, because it depends on `context.managers`
2. `withStyles` and `createUseStyle` are two APIs which calls `createStyleSheet`, but they cannot get options with method return.

So, I just wrote tests for  `getSheetOptions `, and the tests make the edit in this PR as expected.

kof:
How did you discover that bug, if I asked you to create a reproducible codesandbox example, how would you do it?

NinoFocus:
example: https://codesandbox.io/s/boring-wood-s8uvg

kof:
I would try to simply it and then, what do you expect to test, class names?

yeliex:
> I would try to simply it and then, what do you expect to test, class names?

@kof  The rendered className of component `<Button size="large">Hello JSS</Button>` should be `ui-button ui-button-large` as output of `generateId` in `createUseOwnStyles`, but got `root-1-2-5 large-1-2-6`.

kof:
right, so you want to test the generated css like this https://github.com/cssinjs/jss/blob/master/packages/react-jss/test-utils/createBasicTests.js#L49

yeliex:
@kof @HenriBeck Test update, review please

cc @NinoFocus

kof:
looks almost ready, we have shared tests between hoc and hooks here https://github.com/cssinjs/jss/blob/master/packages/react-jss/test-utils/createBasicTests.js

Probably not ideal location to have it in test-utils.

yeliex:
> Probably not ideal location to have it in test-utils.
@kof What is the meaning of `Probably not ideal location`? Would you want me to transfer the test to test-utils?

kof:
I ment that the fact you didn't find shared tests is probably because having test generator function in test-utils isn't a good idea. Right now your test should go there though, to be written once and used for both.

yeliex:
@kof 
I found the test utils, actually.  And I did not do that because the way to create `styled component` is different for `hok` and `hooks` in these tests 
```
// Component for hoc
const withCustom = styles =>
        withStyles(styles, {
          generateId: rule => `ui-${rule.key}`
        })

      const StyledComponent = withCustom({
        button: {
          color: 'red'
        }
      })(props => <button className={props.classes.button} type="button" />)

// for hooks 
const useCustom = styles =>
        createUseStyles(styles, {
          generateId: rule => `ui-${rule.key}`
        })

      const useStyles = useCustom({
        button: {
          color: 'red'
        }
      })

      const StyledComponent = props => {
        const classes = useStyles(props)

        return <button className={classes.button} type="button" />
      }
```
It is different with the tests in `createBaseTests`, the core code to call `createUseStyles` and `withStyles`  are not reusable.

kof:
it's possible, see https://github.com/cssinjs/jss/pull/1263/files  there is an options as second argument

Your commits are preserved in that PR, so I am gonna close this one and merge the other one, let's continue there

PR #1261, opened 2020-01-12
Fixed compatibility issue with jss-plugin-rule-value-function

olee:
When using jss-plugin-template with a function rule, the style string would not be parsed.
Example:
```js
const styles = (props) => `
    background-color: ${props.bgColor};
`;
```

By also checking for a string value in onProcessStyle, this will work properly.

### DIFF
--- a/packages/jss-plugin-template/src/index.js
+++ b/packages/jss-plugin-template/src/index.js
@@ -9,6 +9,13 @@ const onProcessRule = rule => {
   }
 }
 
+const onProcessStyle = style => {
+    if (typeof style === 'string') {
+        return parse(style);
+    }
+    return style;
+}
+
 export default function templatePlugin(): Plugin {
-  return {onProcessRule}
+  return {onProcessRule, onProcessStyle}
 }


kof:
Closing due to inactivity.

olee:
Sorry, but this is still an issue, however I didn't know and have the time to write a test for this right.
Could someone maybe point me to the right place for this or maybe even provide one for the example above?

kof:
A big downside of using template strings inside of the function is that parsing of those can't be done once and cached forever ... it's going to be parsed on every update. So maybe it's not a good idea after all without a better technical solution.

olee:
But right now it's a fact that using function rules with  jss-plugin-template just breaks it completely, even though it should work.
Also correct me if I'm wrong, but doesn't jss-plugin-template even work in a way, that the css inside the template is not processed at all (parsing, prefixing)?
That would mean it could potentially provide better performance than using style objects.

kof:
as a rule of thumb, if something is not documented and has no tests, it is not officially supported

kof:
We need a way to parse the string once, after which update only values. Parsing right now is very simple but it's incomplete, so we need to think about the future versions.

An option is to use a tagged template and interpolate funcitons which receive props.

PR #1266, opened 2020-01-16
[react-jss] fixed ts typings for hook, created common interface for options

BEGEMOT9I:
Fix

## Description
TS-error in this case:
```ts
import React from 'react'
import { createUseStyles } from 'react-jss'
import { createTheming } from 'theming'

const context = React.createContext({})
const theming = createTheming(context)

const styles = {
  container: {
    backgroundColor: 'green'
  }
}
const useWithoutThemeStyles = createUseStyles(styles, { theming })

export { useWithoutThemeStyles }
```
Error:
```
Argument of type '{ theming: Theming<{}>; }' is not assignable to parameter of type 'CreateUseStylesOptions'.
  Object literal may only specify known properties, and 'theming' does not exist in type 'CreateUseStylesOptions'.
```


kof:
merged, added changelog, thanks @BEGEMOT9I

Issue #1156, opened 2019-07-07
Fix examples links


kof:
Since we changed plugin names, we broke examples links.

Also, mb it's better to maintain examples repo in the jss monorepo?

m4theushw:
I wanna make my first contribution. Is this an issue yet?

kof:
Yes, totally

kof:
Depending how far you think you can take it, this issue may be just fixing links or moving examples into the monorepo. The last one wouldn't be easy.

m4theushw:
I'm copying the [examples](https://github.com/cssinjs/examples/) repo to the monorepo and fixing the npm scripts to point to the right locations. GitHub Pages will be enabled on branch `master` to keep examples [like this](http://cssinjs.github.io/examples/perdido/) working?

kof:
Wait a sec, lets sync this with @HenriBeck there is also another option to make all examples in sandbox format, so that they are in the repo and loadable with codesandbox here is an example https://github.com/cssinjs/egghead/tree/master/from-sass-to-cssinjs/default-values

kof:
In any case we need to decide if we really want them in the monorepo or not.

HenriBeck:
@kof, Yes codesandbox would be the preferred way IMO. It allows to directly fork them and start using them which is very nice.

I think they can be in the monorepo in a `examples/` directory. The question is if codesandbox supports this

kof:
yes and also embed on a site

kof:
for codesandbox it's just a directory, that monorepo package.json will need stuff that codesandbox needs

kof:
I am sometimes wondering though how do we decide what goes into monorepo and what not, one way to think about it is monorepo packages should have/need the same setup, so that we can have same management from the root, but examples turns out different, so I am not sure any more

m4theushw:
@kof [react-admin](https://github.com/marmelab/react-admin) follows the same logic. They have a monorepo and also some examples.

I think I'll move examples to an `examples/` directory and embed them in the docs using codesandbox. This way we can remove the compiled files. Each example will have its own package.json.

kof:
@m4theushw you mean in jss repository? `jss/examples`

m4theushw:
@kof Yes, like @HenriBeck suggested

kof:
alright, let's go with this

kof:
released in 10.1.0

PR #1271, opened 2020-01-28
Migrate examples to monorepo

m4theushw:
#1156 

This PR copies examples from https://github.com/cssinjs/examples to the monorepo and formats them to the codesandbox format.


examples/angular/app.js
```diff
@@ -19,8 +19,4 @@ const sheet = jss.createStyleSheet(styles).attach()
 
 angular.module('myApp', []).controller('myController', $scope => {
   $scope.classes = sheet.classes
-  $scope.showSource = () => {
-    // eslint-disable-next-line no-restricted-globals
-    location.href = 'https://github.com/cssinjs/examples/tree/master/angular'
```
kof: I think we should keep all links to the repo (new url) because it won't be obvious how to contribute to the github repo when they see a codesandbox, we just need a different string instead of "view source",  "View the source on GitHub"

m4theushw: I had to remove the "View on Github" button because it's styled by a [stylesheet](https://github.com/cssinjs/examples/blob/gh-pages/calendar/index.html#L6) outside the example files. This file would not be available to codesandbox. Can we put an absolute link to the CSS?

kof: Or we could leave it out and link to github same place where we are going to link to codesandbox, so that there is no need for doing it inside of example.

m4theushw: Ok, when I change the documentation site repo I'll re-add the button 'View on Github'.

kof:
What's the status here? Should we review/merge?

m4theushw:
@kof You can make a first review. There are some improvements to be done yet:
- The React SSR example doesn't work in codesandbox (how should we support it?).
- Replace class components with functional components.

I could make these improvements in another PR.

kof:
SSR over codesandbox can be done even better, because you can have an
actual server there

m4theushw:
React SSR example updated to work on codesandbox.

kof:
Hey @m4theushw let's create list of todos so that we know what is left and can let other folks help

m4theushw:
@kof All examples from https://github.com/cssinjs/examples with the exception of the `plugins` folder were converted to codesandbox. I don't know if we should convert these remaining demos because the docs already describe their output.

kof:
@m4theushw idk, it might still make sense, because people can try run them quickly and see what happens without having to install a lot, it might help

m4theushw:
@kof ok, I can convert them too. But would not be worth to merge this PR and I work on them in another smaller PR? These plugins' demos are not mentioned in the docs so we'll not have an impact now.

kof:
Sure, lets do that.

kof:
For some reason CI is failing

kof:
Oh just realized the branch is on your fork,  closing this one in favor of https://github.com/cssinjs/jss/pull/1306

Issue #1262, opened 2020-01-12
.addRule with @font-face does not work the second time


geigerzaehler:
__Expected behavior:__
Using the following code I’d expect the stylesheet to contain _two_ `@font-face` declrations, one for each call of `.addRule`.
~~~js
const sheet = jss.createStyleSheet()
sheet.addRule('@font-face', { name: 'font1' })
sheet.addRule('@font-face', { name: 'font2' })
sheet.attach()
~~~

__Describe the bug:__
The stylesheet contains only the first `@font-face` declration.

__Codesandbox link:__ 
https://codesandbox.io/s/xenodochial-chatelet-rpqrc

__Versions (please complete the following information):__
 - jss: 10.0.3 (was working in 10.0.2)
 - Browser [e.g. chrome, safari]: all
 - OS [e.g. Windows, macOS]: all

kof:
will be fixed in 10.0.5 or higher

kof:
released in 10.1.0

PR #1279, opened 2020-02-07
Fix when declaring multiple @font-face

m4theushw:
#1262 

When adding multiple `@font-face` the key passed to `FontFaceRule` after the first entry follows the pattern: `@font-face-dX`. This was causing the 2nd rule to be ignored since this is not a valid CSS declaration. My solution checks if the key starts with `@font-face`.

## Todo

- [x] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/jss/.size-snapshot.json
+++ b/packages/jss/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "dist/jss.js": {
-    "bundled": 61836,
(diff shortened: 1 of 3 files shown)


kof:
Thanks @m4theushw looks good to me, fixing some minor thing and merging from a separate PR to run the CI https://github.com/cssinjs/jss/pull/1280

PR #1284, opened 2020-02-16
Allow a Comment node as insertion point

cmfcmf:

--

The `InsertionPoint` typings only allow `string`s and `HTMLElement`s. The following code currently results in a type error, but works fine:

```ts
import { create } from "jss";

const comment = document.createComment("hi");
// ... add comment to the DOM ...

const jss = create({ insertionPoint: comment });
```

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

^- Not sure if any of them are needed for this PR.

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -34,7 +34,7 @@ export type JssValue =
   | null
   | false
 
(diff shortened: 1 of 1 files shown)


kof:
thank you

kof:
released in 10.1.0

Issue #1057, opened 2019-03-12
Weird inconsistent behaviour with dynamics value


VincentLanglet:
__Codesandbox link:__ 
https://codesandbox.io/s/p2wljkm9zm

In this example you can see
```tsx
const styles = () =>
  createStyles({
    text: {
      backgroundColor: "red", // WORKING
      color: () => "blue", // WORKING
      "&:hover": {
        marginLeft: 20, // WORKING,
        marginRight: () => 20 // WORKING,
      }
    },
    someText: () => ({
      backgroundColor: "red", // WORKING
      color: () => "blue", // NOT WORKING
      "&:hover": {
        marginLeft: 20, // WORKING, 
        marginRight: () => 20 // WORKING,
      }
    }),
    otherText: {
      backgroundColor: "red", // WORKING
      color: () => "blue", // WORKING
      "&:hover": () => ({
        marginLeft: 20, // NOT WORKING,
        marginRight: () => 20 // NOT WORKING,
      })
    },
    finalText: () => ({
      backgroundColor: "red", // WORKING
      color: () => "blue", // NOT WORKING
      "&:hover": () => ({
        marginLeft: 20, // WORKING,
        marginRight: () => 20 // NOT WORKING,
      })
    })
  });
```

I understand the fact function inside function does not work, like color of someText, but in this case why marginRight of someText does work ?
And why marginLeft of otherText does not work ? Even weirder, it works for finalText...

If it works when it shouldn't at least it's not a big deal. But I think the case of marginLeft of otherText should be a correct one, and actually it doesn't work.

HenriBeck:
`marginRight ` works because `&:hover` will become a separate rule on the sheet.
```js
someText: () => ({
      backgroundColor: "red", // WORKING
      color: () => "blue", // NOT WORKING
      "&:hover": {
        marginLeft: 20, // WORKING, 
        marginRight: () => 20 // WORKING,
      }
    }),
```
will become:
```js
someText: () => ({
      backgroundColor: "red", // WORKING
      color: () => "blue", // NOT WORKING
    }),
 "someTextSelector:hover": {
        marginLeft: 20, // WORKING, 
        marginRight: () => 20 // WORKING,
      }
```

kof:
We should warn when user tries to do unsupported stuff when possible.

HenriBeck:
I think why `marginLeft` inside `otherText` doesn't work is because we split dynamic rules from the static rules. I think if we would to do this just with jss it would work.

[...]

PR #1285, opened 2020-02-19
Add warning when using a function value inside a function rule

m4theushw:
#1057 

Add warning when user tries to use a function value inside a function rule.

## Todo

- [x] Add tests if possible
- [x] Add documentation

### DIFF
--- a/packages/jss-plugin-rule-value-function/src/index.js
+++ b/packages/jss-plugin-rule-value-function/src/index.js
@@ -1,4 +1,5 @@
 /* @flow */
+import warning from 'tiny-warning'
 import {
   createRule,
   type Rule,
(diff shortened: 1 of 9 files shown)


packages/jss-plugin-rule-value-function/src/index.js
```diff
@@ -55,6 +56,13 @@ export default function functionPlugin() {
         // Empty object will remove all currently defined props
         // in case function rule returns a falsy value.
         styleRule.style = fnRule(data) || {}
+
+        for (const prop in styleRule.style) {
```
kof: We could avoid doing these checks in production by wrapping it in  `if (process.env.NODE_ENV === 'development')`

TrySound: Then we could even avoid warning package and use only `console.warn`.

kof: I would keep warning for consistency, but it doesn't matter technically, just in case in the future we want to replace warning, it will be easier if it's used consistently

m4theushw:
@kof updated! I had to change `webpack.config.js` because `process.env.NODE_ENV` was undefined.

kof:
LGTM, thank you!

kof:
released in 10.1.0

Issue #1297, opened 2020-03-11
JssProvider ignores generateId prop when id prop is also set


felthy:
__Expected behavior:__
To work with multiple React trees, we need to pass a `generateId` prop [as documented](https://cssinjs.org/react-jss/?v=v10.0.0-alpha.9#multi-tree-setup). But to get proper minification, we also need to pass the `id` prop due to how the `minify` variable is initialized here: https://github.com/cssinjs/jss/blob/b36ef074e69aef21dc764a54f572d7ced76041bc/packages/react-jss/src/utils/sheets.js#L35-L38

Given the code below, I would expect my `generateId` function to be used to generate classnames.

```
const id = {
  minify: true
}
const generateId = createGenerateId(id)

ReactDOM.render(
  <JssProvider id={id} generateId={generateId}>
    ...
  </JssProvider>,
  root
)
```

__Describe the bug:__

My custom `generateId` is not used, because `context.generateId` is overwritten with a newly created generator here: https://github.com/cssinjs/jss/blob/b36ef074e69aef21dc764a54f572d7ced76041bc/packages/react-jss/src/JssProvider.js#L79-L81

`prevContext` above is initialized from `initialContext` (an empty object), so this if clause is always true when an `id` prop is present.

Furthermore, `createGenerateId` here is the default implementation exported by `jss`, even if jss itself has been configured to use a custom `createGenerateId` function.

__Versions (please complete the following information):__
 - jss: `10.0.4`
 - Browser: Google Chrome
 - OS: macOS

kof:
released in 10.1.0

PR #1298, opened 2020-03-11
Fix generateId ignored by JssProvider when there is an id prop (#1297)

felthy:
#1297 

Avoid overwriting custom `generateId` when there is also an `id` prop.

## Todo

- [x] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "dist/react-jss.js": {
-    "bundled": 169027,
-    "minified": 58353,
+    "bundled": 169023,
+    "minified": 58350,
     "gzipped": 19101
   },
   "dist/react-jss.min.js": {
-    "bundled": 112062,
-    "minified": 41612,
+    "bundled": 112058,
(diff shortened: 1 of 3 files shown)


kof:
LGTM, I think you missed that we have this describe though, we could move that test as an it inside https://github.com/cssinjs/jss/blob/63e2d33cc7a8ffe7c7cadfd537faebd4a6a5f6d6/packages/react-jss/src/JssProvider.test.js#L195

kof:
Need to merge it now, since I want to do a release today, so tbc here https://github.com/cssinjs/jss/pull/1304

felthy:
@kof re the existing `describe('id prop')`, I didn't use it because the one you referenced is nested inside of `describe('nested child JssProvider')`, which has nothing to do with this test. There wasn't a relevant top-level `describe` so I added a new one.

kof:
true, I didn't notice myself it's inside there, most provider tests are there, lol, so maybe we should just rename the thing and keep doing all the things in nested providers, because it makes sure that standalone works as well as nested.

PR #1303, opened 2020-03-15
Move getDynamicStyles into react-jss package

kof:
Just moving a function.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,10 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+### Improvements
+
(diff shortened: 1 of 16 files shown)


Vadorequest:
@kof This broke our builds. https://github.com/mui-org/material-ui/issues/20132

Could a revert be made ASAP until a proper way of fixing this is found?

Vadorequest:
I believe this supposedly private function was actually used by other libraries, such as material-ui. This becomes a breaking change and should be done using a major release change.

kof:
Should be fixed now  in 10.1.1

Vadorequest:
Just confirming it works again. Thank you for the quick fix!

PR #1306, opened 2020-03-15
Migrate examples to monorepo

kof:
#1156 

This PR copies examples from https://github.com/cssinjs/examples to the monorepo and formats them to the codesandbox format.


kof:
released in 10.1.0

Issue #1309, opened 2020-03-16
[react-jss] Theming type is not exported


dfernandez-asapp:
__Expected behavior:__

Since the function `createTheming` is exported by `react-jss`, I'll expect to import the `Theming` type from the same place:

```
import { createTheming,Theming } from 'react-jss';

const result : Theming<MyCustomTheme> = createTheming<MyCustomTheme>(ThemeContext);
```

__Describe the bug:__

```
import {
createTheming,
Theming // Fails Theming is not exported
} from 'react-jss';
```

__Codesandbox link:__ https://codesandbox.io/s/vigorous-keller-fsw88?fontsize=14&hidenavigation=1&theme=dark

__Versions (please complete the following information):__
 - jss: 10.1.1
 - Browser [e.g. chrome, safari]: Chrome
 - OS [e.g. Windows, macOS]: macOS

kof:
Will be released in 10.1.1 or higher

kof:
Released in v10.2.0

Issue #1308, opened 2020-03-16
[react-jss] createUseStyles TypeScript typings ignores Theme generic argument for theming option


dfernandez-asapp:
__Expected behavior:__

This code fails to compile due to type errors:
```
type MyCustomTheme = {color: string};
const ThemeContext = React.createContext<MyCustomTheme>({color: 'black'});
const theming = createTheming<MyCustomTheme>();

const useStyles = createUseStyles<MyCustomTheme>(theme => ({
   someName: {color: theme.color}
}), { 
  theming /* <-- Fails Theming<MyCustomTheme> is not assignable to Theming<object> */
});
```

__Describe the bug:__
The generic function `createUseStyles` receives a `Theme` type as a parameter: 
`createUseStyles<MyCustomTheme>`. It also receives a `theming` option to pass the implementation of `ThemeContext`. However, the `Theme` template parameter is never used in for the declaration of options.

__Codesandbox link:__ https://codesandbox.io/s/headless-night-xu428?fontsize=14&hidenavigation=1&theme=dark

__Versions (please complete the following information):__
 - jss: 10.1.1
 - TypeScript: 3.8.3
 - Browser [e.g. chrome, safari]: Chrome 
 - OS [e.g. Windows, macOS]: macOS

kof:
Will be released in 10.1.1 or higher

kof:
Released in v10.2.0

PR #1310, opened 2020-03-16
TypeScript fixes for #1308 and #1309

dfernandez-asapp:

#1308
#1309

This PR changes the type definition of `createUseStyles<Theme>(styles, options?: CreateUseStyles)` so it passes the `Theme` type parameters to the options type: `createUseStyles<Theme>(styles, options?: CreateUseStyles<Theme>)`.

It also re-exports the `Theming` type, so users of `createTheming` can do explicit typing of the result: `const theming : Theming = createTheming(...` without having to add the `theming` package to the direct dependencies.

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/react-jss/src/index.d.ts
+++ b/packages/react-jss/src/index.d.ts
@@ -45,9 +45,11 @@ interface WithStylesProps<S extends Styles | ((theme: unknown) => Styles)> {
  */
 type WithStyles<S extends Styles | ((theme: unknown) => Styles)> = WithStylesProps<S>
 
-interface BaseOptions extends StyleSheetFactoryOptions {
+export interface DefaultTheme {}
+
+interface BaseOptions<Theme = DefaultTheme> extends StyleSheetFactoryOptions {
   index?: number
-  theming?: Theming<object>
+  theming?: Theming<Theme>
 }
 
 interface WithStylesOptions extends BaseOptions {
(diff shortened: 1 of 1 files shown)


kof:
Is this considered a breaking change?

dfernandez-asapp:
Is not a breaking change:
* If you don't use TypeScript you are not affected by this.
* If you use TypeScript, expressions like `createUseStyles<XYZ>()`, `createTheming(Context)` or `createTheming<XYZ>(Context)` are not affected since the generic argument for the options will take a default (that is the same as before `{}`).

The main difference is that before this PR you can't pass a custom `theming` to `createUseStyles` in TypeScript without having to cast it to `Theming<{}>` and to cast it to `Theming` you need to import the type from the `theming` module.

A few things not included in the PR:
* I didn't add any tests, because I didn't see any typings tests (eg dtslint or tsd) in the project. This is a compile time check that cannot be tested with a normal unit test.
* I didn't do any change to Flow typings, I don't use Flow and I don't know if the Flow types definition has the same issue or not.

kof:
I am going to merge it and release with the next release, but since I already released yesterday, I am gonna wait for more PRs to get merged.

kof:
Merged, Thanks

kof:
Released in v10.2.0

PR #1314, opened 2020-03-16
add styled to type exports in 'react-jss'

jeremy-coleman:
for some reason, react-jss's index.d.ts does not export the styled function. these 2 lines of code fix that (only changed react-jss/index.d.ts , NOT the index.js file)


kof:
styled interface isn't documented nor officially release, please hold on with that and using it

jeremy-coleman:
but it's 3x faster than mui styled :(

kof:
good to know, how did you measure?

PR #1317, opened 2020-03-18
Add SwissDev Jobs as a user

Vrq:
## Todo

- [x ] Add tests if possible
- [x ] Add changelog if users should know about the change
- [x ] Add documentation

### DIFF
--- a/docs/users.md
+++ b/docs/users.md
@@ -37,3 +37,6 @@
 [Onepixel](https://www.onepixel.com/)
 
 [React Admin](https://marmelab.com/react-admin/)
+
+[SwissDev Jobs](https://swissdevjobs.ch/jobs/JavaScript/All)
+


Vrq:
@kof would you mind having a look? :)

Vrq:
@HenriBeck could you check it out, please? :)

HenriBeck:
@Vrq how are you using `jss`? I can't find any style tag which would indicate the usage of `jss`

Vrq:
@HenriBeck we started using it this year and the rollout in production is still ongoing but you can find it for example on: https://swissdevjobs.ch/open

kof:
@Vrq I don't see jss on that site either, I looked into class names and looked into the header for generated style tags, but I am gonna just trust you on this and merge it.

Vrq:
Thanks @kof! 
Maybe it's because of the caching, but it should be there:
code

PR #1322, opened 2020-03-30
Add Tradie Training to users

ryan-senn:
Thanks for this great project!


ryan-senn:
Actually I might close this, we use it on our internal admin section for now, it's not public facing so maybe not a great example? Might re-open if we migrate the publicly facing website.

PR #1343, opened 2020-05-09
fix(react-jss): Media query styles applied only to the first element in function

### LINKED ISSUE #1320: Media query styles applied only to the first element in function
__Expected behavior:__
media query should apply styles to all the component that uses it

__Describe the bug:__
only the first component gets the styles of media query

__Codesandbox link:__ 
https://codesandbox.io/s/react-jss-playground-bz42d

 - jss:10.1.1
 - Browser saw it on chrome, havent tested others
 - OS windows10:

image

pranshuchittora:
Closes #1320 

Fixed `react-jss`
## Todo

- [ ] Add tests if possible
- [ ] Add documentation

P.S. Will add tests if the preliminary changes are acceptable.
Demo (React JSS + React Hot Loader) -> https://youtu.be/utjjiwLn7lo

### DIFF
--- a/packages/react-jss/src/createUseStyles.test.js
+++ b/packages/react-jss/src/createUseStyles.test.js
@@ -1,16 +1,72 @@
 /* eslint-disable react/prop-types */
-import createUseStyles from './createUseStyles'
-import createBasicTests from '../test-utils/createBasicTests'
 
-const createStyledComponent = (styles, options) => {
-  const useStyles = createUseStyles(styles, options)
-  const Comp = props => {
-    useStyles(props)
-    return null
-  }
-  return Comp
+import expect from 'expect.js'
+import React from 'react'
+import TestRenderer from 'react-test-renderer'
+import {stripIndent} from 'common-tags'
+
+import {SheetsRegistry, JssProvider, ThemeProvider, createUseStyles} from '.'
+
+const createGenerateId = () => {
+  let counter = 0
+  return rule => `${rule.key}-${counter++}`
+}
+
+const theme: Object = {
+  background: 'yellow',
+  background2: 'red'
 }
 
 describe('React-JSS: createUseStyles', () => {
-  createBasicTests({createStyledComponent})
+  it('should render multiple elements with applied media query', () => {
+    const registry = new SheetsRegistry()
+    const useStyles = createUseStyles(themeObj => ({
+      wrapper: () => ({
+        padding: 40,
+        background: themeObj.background,
+        textAlign: 'left',
+        '@media (min-width: 1024px)': {
+          backgroundColor: themeObj.background2
+        }
+      })
+    }))
+
+    const Comp = () => {
+      const classes = useStyles(theme)
+      return <div className={classes.wrapper} />
+    }
+
+    const a = [1, 2]
+    TestRenderer.create(
+      <JssProvider registry={registry} generateId={createGenerateId()}>
+        <ThemeProvider theme={theme}>
+          {a.map(item => (
+            <Comp key={item} />
+          ))}
+        </ThemeProvider>
+      </JssProvider>
+    )
+    expect(registry.toString()).to.be(stripIndent`
+    .wrapper-0 {}
+    .wrapper-d0-1 {
+      padding: 40px;
+      background: yellow;
+      text-align: left;
+    }
+    @media (min-width: 1024px) {
+      .wrapper-d0-1 {
+        background-color: red;
+      }
+    }
+      .wrapper-d1-2 {
+        padding: 40px;
+        background: yellow;
+        text-align: left;
+      }
+    @media (min-width: 1024px) {
+      .wrapper-d1-2 {
+        background-color: red;
+      }
+    }`)
+  })
 })
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,23 +1,23 @@
 {
   "dist/react-jss.js": {
-    "bundled": 130189,
-    "minified": 46655,
-    "gzipped": 16051
+    "bundled": 171071,
+    "minified": 59339,
+    "gzipped": 19438
   },
   "dist/react-jss.min.js": {
-    "bundled": 103803,
-    "minified": 38662,
-    "gzipped": 13692
+    "bundled": 114048,
+    "minified": 42598,
+    "gzipped": 14460
   },
   "dist/react-jss.cjs.js": {
-    "bundled": 26922,
-    "minified": 11621,
-    "gzipped": 3848
+    "bundled": 27447,
+    "minified": 12023,
+    "gzipped": 3885
   },
   "dist/react-jss.esm.js": {
-    "bundled": 25951,
-    "minified": 10777,
-    "gzipped": 3726,
+    "bundled": 25941,
+    "minified": 10783,
+    "gzipped": 3728,
     "treeshaked": {
       "rollup": {
         "code": 1838,
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,23 +1,23 @@
 {
   "dist/react-jss.js": {
-    "bundled": 169946,
-    "minified": 58911,
-    "gzipped": 19311
+    "bundled": 130189,
+    "minified": 46655,
+    "gzipped": 16051
   },
   "dist/react-jss.min.js": {
-    "bundled": 112923,
-    "minified": 42170,
-    "gzipped": 14335
+    "bundled": 103803,
+    "minified": 38662,
+    "gzipped": 13692
   },
   "dist/react-jss.cjs.js": {
-    "bundled": 26398,
-    "minified": 11532,
-    "gzipped": 3759
+    "bundled": 26922,
+    "minified": 11621,
+    "gzipped": 3848
   },
   "dist/react-jss.esm.js": {
-    "bundled": 24901,
-    "minified": 10301,
-    "gzipped": 3602,
+    "bundled": 25951,
+    "minified": 10777,
+    "gzipped": 3726,
     "treeshaked": {
       "rollup": {
         "code": 1838,
--- a/packages/react-jss/src/createUseStyles.js
+++ b/packages/react-jss/src/createUseStyles.js
@@ -20,6 +20,13 @@ const useEffectOrLayoutEffect = isInBrowser ? React.useLayoutEffect : React.useE
 
 const noTheme = {}
 
+const reducer = (prevState, action) => {
+  if (action.type === 'updateSheet') {
+    return action.payload
+  }
+  return prevState
+}
+
 const createUseStyles = <Theme: {}>(styles: Styles<Theme>, options?: HookOptions<Theme> = {}) => {
   const {index = getSheetIndex(), theming, name, ...sheetOptions} = options
   const ThemeContext = (theming && theming.context) || DefaultThemeContext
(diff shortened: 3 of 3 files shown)


kof:
I definitely need a test before I can look into the solution, there might be multiple ways to fix it.

kof:
hey @pranshuchittora  I am about to make a release, but sadly I can't merge it because it has no test

pranshuchittora:
Pardon @kof  was a bit busy with the work, will write tests for sure, by the end of this week.

anthonygrignoux:
Hello, is there any news about that PR?

pranshuchittora:
@kof I have added the test for the following fix.
And once again I apologize for the delay.

packages/react-jss/src/createUseStyles.test.js
```diff
@@ -1,16 +1,72 @@
...
+}
+
+const theme: Object = {
+  background: 'yellow',
+  background2: 'red'
 }
 
 describe('React-JSS: createUseStyles', () => {
-  createBasicTests({createStyledComponent})
+  it('should render multiple elements with applied media query', () => {
```
kof, line 21: I wonder if this test should be applied to both hook and hoc based implementation, just to be sure both work same way

pranshuchittora: Can you pls elaborate, like exactly which components are needed to be tested.

kof: createBasicTests() is the function that runs all the tests which use the passed function which creates a styled component. You picked the test file that only serves as an entry point to generate tests using hooks interface,

If we want to run the test with both interfaces, you need to add the test inside createBasicTests

Also it seems like you removed createBasicTests() call entirely from this file, which means you don't run all the tests using hooks interface at all now

kof:
hey @pranshuchittora, any updates?

pranshuchittora:
Hi @kof , It would to great if anyone can help me with the tests.

mattbeaudry:
hungry for this fix, might have time to take a look this weekend (or next) to lend a hand with tests, assuming it hasn't been merged by then. :)

pranshuchittora:
I apologize for blocking this. Due to some huge academic work, I won't be able to work on this in the coming days. 
Hence closing this PR.
I am extremely sorry for that.

kof:
I think it is very close to a mergeable state, lets keep it open, mb someone will make the required tests changes

pranshuchittora:
Sounds great. Before proceeding with FTs It would be great if you can pull this PR locally and test the new changes manually.

alee8:
Any updates on this PR?  We make a lot of use of media query for responsive UI and have had to find ways to work around this issue (e.g., static object, static media query prop key).

mattbeaudry:
@kof If i understand what you needed for the updated test correctly, then I think I have a fix, but let me know if more is needed. I made a PR on @pranshuchittora jss fork https://github.com/pranshuchittora/jss/pull/1

mattbeaudry:
How's this look for the test updates: https://github.com/pranshuchittora/jss/pull/1

EZEDSEA:
Looks like tests are failing for the build cuz of an outdated snapshot on popout.   I had to update the snapshot in my branch too.  Might just want to do that before we merge this.

jonathan-l-rivera:
Any updates? Really could use this fixed right now. It's blocking a project that uses dynamically set media queries. Any word?

kof:
Hey everyone, I have other priorities and unless someone submits a fully mergeable PR or close to one where my involvement would be limited, nothing is gonna happen.

Asking about the status is not helpful. If you want to help - fix it.

jonathan-l-rivera:
> Hey everyone, I have other priorities and unless someone submits a fully mergeable PR or close to one where my involvement would be limited, nothing is gonna happen.
> 
> Asking about the status is not helpful. If you want to help - fix it.

@kof it looks like @mattbeaudry did 29 days ago. If that's not sufficient, that wasn't made clear.

kof:
Didn't realize  this was based off the branch with the fix, so it is passing the tests I suppose. Did anyone try the changes from this PR on a code base?  

Those are massive and I am a bit afraid they will break stuff.

EZEDSEA:
> Didn't realize this was based off the branch with the fix, so it is passing the tests I suppose. Did anyone try the changes from this PR on a code base?
> 
> Those are massive and I am a bit afraid they will break stuff.

@kof I just pulled down the branch and tested it against our codebase.  I can confirm that everything seems to work as expected and it DID fix some bugs we had that is effected by this issue.

kof:
@EZEDSEA good to know, I am gonna test it again soon and release a .pre version for everyone to test

kof:
Lol, those tests pass old lib code as well https://github.com/cssinjs/jss/pull/1442#issuecomment-762440005

kof:
We can't merge this until we have proper tests, validating the fix.

EZEDSEA:
> We can't merge this until we have proper tests, validating the fix.

@kof I noticed you ended up merging it.  Were @mattbeaudry tests sufficient in the end or do we still need to add some more tests to validate?

kof:
I merged it because its a good test to have, but it doesn't verify the behavior of the changes made in this PR at ALL, so basically it has nothing to do with this PR. We still need tests that verify changes that are made here and fail on the code base before the changes were made.

alee8:
**Summary of Current Situation with this PR #1343**
@kof I investigated the work that @pranshuchittora did in this PR and to try to provide verification tests.  To summarize:

1. @pranshuchittora found that this bug was introduced with v10.0.1 by the change to [packages/react-jss/src/createUseStyles.js](https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/createUseStyles.js) in [this commit 9f7924e](https://github.com/cssinjs/jss/commit/9f7924eb79bf1e78816db7aaa4cf324471)
2. His fix was to [revert the change](https://github.com/cssinjs/jss/commit/9f7924eb79bf1e78816db7aaa4cf324471)
3. @pranshuchittora produced a test that verified rendering multiple elements using media query and theme function ([this was accepted as generally useful dynamicStyles test.](https://github.com/cssinjs/jss/pull/1237/files)) However, the test needs to additionally show that a viewport change would still render the multiple elements correctly. 
4. @EZEDSEA confirmed that this reverted code does fix the issues in their code base. @mattbeaudry can comment as well on whether it fixes his code base.

**An Effective Verification Test**
For @pranshuchittora test [to verify the fix]( https://github.com/cssinjs/jss/commit/9f7924eb79bf1e78816db7aaa4cf324471), it needs to be extended/modified into two parts each demonstrating that the end result of applying the CSS occurred (different background changes based on viewport/window sizes):
1. Render component within a given window size (e.g., < 1024) so that the 2 wrapped MyComponent has background yellow 
2. Change the window size (e.g., > 1024) so that the 2 wrapped MyComponent activate the @media query to set the background red

**Limitations of react-test-renderer framework**
The current react-test-renderer framework does not support viewport/window changes.  [Tests involving window size changes have been shown to work using Jest, Enzyme, and JSDom](https://stackoverflow.com/questions/45868042/figuring-out-how-to-mock-the-window-size-changing-for-a-react-component-test).  They [simulate window resize using JSDom by manually setting window.innerWidth](https://stackoverflow.com/questions/46221210/jest-enzyme-how-to-test-at-different-viewports/51997030#answer-51997030) and window.innerHeight.

After a window resize, JSDom needs to activate the @media query (setting some props and styles).  [The current version of JSDom still lists @media under "TODO"](https://github.com/jsdom/jsdom/blob/d6f8a97b5fb7709d6ad0215c1ae95fd4cab58489/lib/jsdom/level2/style.js#L29). Digging deeper, [this comment](https://github.com/testing-library/jest-dom/issues/113#issuecomment-496971128
) illustrates *why it is not possible to use JSDom, react-test-renderer or even Jest, Enzyme* for the 2-step test approach outlined above.

**Where does this Leave this Open Issue #1320 and PR #1343**
Here are some options to help unblock this PR for several projects that rely on this fix:

1. Need a different test approach than outlined above.  Suggestions are welcomed. Perhaps any guidance about doing [something similar to the hypothetical approach](https://github.com/testing-library/jest-dom/issues/113#issuecomment-496971128) for JSS OR
2. Consider [testing media queries through Cypress or similar framework](https://github.com/testing-library/jest-dom/issues/113#issuecomment-522231790) where stylesheet actually get rendered OR
3. Take this PR #1343 fix which was known to work prior to the [changes introduced by this commit 9f7924eb](https://github.com/cssinjs/jss/commit/9f7924eb79bf1e78816db7aaa4cf324471) and confirmed to work by @EZEDSEA recently. OR
4. Taking Option 3 will mean that [react-hot-loader feature supported by refactor of createUseStyles](https://github.com/cssinjs/jss/pull/1237) will need to be re-investigated.

**Meta Issue**
The need for tests is important particularly to verify this PR #1343 or the flaw with  [react-hot-loader implementation 9f7924eb](https://github.com/cssinjs/jss/commit/9f7924eb79bf1e78816db7aaa4cf324471).  If there is a testing strategy that works with react-test-renderer it would be good to articulate.  If it is not possible, what alternate testing framework could be used to verify?  Would the repo owners approve such an endeavor?

kof:
Thank you for the summary, I decided to look into the problem now. Here is a simplified reproduction I wish I had before

https://codesandbox.io/s/react-jss-playground-forked-sty9t?file=/index.js

kof:
Given this code:

```js
  wrapper: () => ({
    padding: 40,
    backgroundColor: "green",
    textAlign: "left",
    "@media (min-width: 1024px)": {
      backgroundColor: "red"
    }
  })
});

const Comp = () => {
  const classes = useStyles();
  return (
    <div className={classes.wrapper}>
      <h1>Hello React-JSS!</h1>
      <a href="http://cssinjs.org/react-jss" traget="_blank">
        See docs
      </a>
    </div>
  );
};

const App = () => (
  <React.Fragment>
    <Comp />
    <Comp />
  </React.Fragment>
);

render(<App />, document.getElementById("root"));
```
Registry contains correct CSS:

```
        .wrapper-0 {}
        .wrapper-d0-1 {
          padding: 40px;
          background: red;
          text-align: left;
        }
        @media (min-width: 1024px) {
          .wrapper-d0-1 {
            background-color: green;
          }
        }
          .wrapper-d1-2 {
            padding: 40px;
            background: red;
            text-align: left;
          }
        @media (min-width: 1024px) {
          .wrapper-d1-2 {
            background-color: green;
          }
        }
```

But the problem is that the second media query from a dynamic rule doesn't have the declaration inside when rendering in the browser:

EZEDSEA:
@kof What can we do to help?  You had mentioned unit tests didn't seem to test for this before.  The fix that is proposed initially seems to just suggest a rollback of code from v10.0.1.

kof:
@EZEDSEA I am searching for a solution that fixes es without reverting to the previous state. The PR that broke it in the first place was trying to fix hot reload, which is also important for many to work and has no tests whatsover.

It's super annoying for a maintainer to git 2 PRs one introduces changes that break one thing fixes another, next one reverts it and breaks the first one.

That is why without a test that is failing I am not going to merge anything at all, otherwise we will be stuck in this loop.

kof:
If you want to help, for now don't worry about the test, but try find a minimal change that fixes the problem without breaking hot reload.

kof:
So here is the real problem:

The problem is not in react-jss at all. It's the combination of jss-plugin-nested and jss-plugin-rule-value-function and overall plugins based architecture together with this particular interface.

```js
  wrapper: () => ({
    padding: 40,
    backgroundColor: "green",
    textAlign: "left",
    "@media (min-width: 1024px)": {
      backgroundColor: "red"
    }
  })
 ```
 
This is just hard to make work efficiently. That function returns style object that contains a nested media query. To make it work on each update  jss-plugin-rule-value-function has to add/update first level rule that includes a nested media query, then jss-plugin-nested needs to add/update that media rule along with it's child rules.
 
It essentially needs to transform the code above to this

```js
      const styles = {
        wrapper: () => ({
          padding: 40,
          background: 'red',
          textAlign: 'left'
        }),
        '@media (min-width: 1024px)': {
          wrapper: {
            backgroundColor: 'green'
          }
        }
      }
```

There are a number of challenges with this. 
1. It needs to avoid creating another media query if one already exists.
1. It needs to avoid creating a new child "wrapper" rule, but it needs to override it's styles in case they have changed
1. It needs to remove media queries that have been potentially created on previous update
1. It needs to remove  previous child rules inside media query which might have been created potentially and now don't exist

This is a lot of diffing and logic to support, the easiest way forward is to not support nested media queries inside of function rules at all, maybe even warn about it.

A similar thing that works right now is and it has none of these problems.

``js
      const styles = {
        wrapper: () => ({
          padding: 40,
          background: 'red',
          textAlign: 'left'
        }),
        '@media (min-width: 1024px)': {
          wrapper: () => ({
            backgroundColor: 'green'
          })
        }
      }
```

One property it is missing is that you can't generate the media condition based on props and for this we would need to find a separate solution. I could imagine a better interface for props based condition would be something like this:

```js

      const styles = {
        wrapper: () => ({
          padding: 40,
          background: 'red',
          textAlign: 'left'
        }),
        '@media': {
          condition: (props) =>  (min-width: 1024px)
          wrapper: (props) => ({
            backgroundColor: 'green'
          })
        }
      }
```

pybuche:
@kof 👋 Any update on this? I decided to switch from SASS to JSS on my project, but I now realize I can't create a "responsive system" to adapt the sizes of my components based on media queries. 

Basically I've started to craft something like this:

```typescript
mport _ from "lodash";
import { CSSProperties } from "react";

export const breakpoints = {
  xxs: 0,
  xs: 376,
  sm: 576,
  md: 768,
  lg: 992,
  xl: 1200,
  xxl: 1440,
} as const;

export type Breakpoint = keyof typeof breakpoints;

export type Responsive<T> = { [key in Breakpoint]?: T };

/**
 * Generates the media query key used in `createUseStyles`, to apply style on screens larger than `breakpoint`
 * See `responsiveHideUp` for an example
 * @param breakpoint
 */
export const responsiveUp = (breakpoint: Breakpoint) => {
  return `@media(min-width: ${breakpoints[breakpoint]}px)`;
};

/**
 * Generates the media query key used in `createUseStyles`, to apply style on screens smaller than `breakpoint`
 * See `responsiveHideDown` for an example
 * @param breakpoint
 */
export const responsiveDown = (breakpoint: Breakpoint) => {
  return `@media(max-width: ${breakpoints[breakpoint]}px)`;
};

/**
 * Style prop usable under `createUseStyles` to hide an elements on screens bigger than `breakpoint`
 * See murray's Frame implementation of `hideUp` for an example
 * @param breakpoint
 */
export const responsiveHideUp = (breakpoint: Breakpoint) => ({
  [responsiveUp(breakpoint)]: { display: "none" },
});

/**
 * Style prop usable under `createUseStyles` to hide an elements on screens small than `breakpoint`
 * See murray's Frame implementation of `hideDown` for an example
 * @param breakpoint
 */
export const responsiveHideDown = (breakpoint: Breakpoint) => ({
  [responsiveDown(breakpoint)]: { display: "none" },
});
```


to implement it in a design-systemed component like this:
```typescript
type Props = {
  hideUp?: Breakpoint;
  hideDown?: Breakpoint;
  // ... other props
}

const useStyles = createUseStyles<string, Props>({
  frameStyle: ({
    hideUp,
    hideDown,
    className,
  }) => {
    const otherStyleProps = {}; // color, background etc

    const hideUpProps = hideUp ? responsiveHideUp(hideUp) : {};
    const hideDownProps = hideDown ? responsiveHideDown(hideDown) : {};

    return {
      composes: [className],
      // ... other props
      ...hideUpProps,
      ...hideDownProps,
    };
  },
});
```

and then use it like this:

```typescript
<MyComponent hideUp="md" />
<MyComponent hideDown="xl" />
```


Do you think that's achievable? If we need to make a change to JSS, I can help, but I might need some pointers first! Thanks!

kof:
seems like we got a successor for the same bug https://github.com/cssinjs/jss/pull/1523

PR #1349, opened 2020-06-04
fix(react-jss): fix react-jss theme types

vismu:
# What would you like to add/fix?

The problem is that my app `Theme` interface didn't match to `unknown` theme type.
So I think `any` type would be better in this case.

### DIFF
--- a/packages/react-jss/src/index.d.ts
+++ b/packages/react-jss/src/index.d.ts
@@ -37,13 +37,13 @@ declare const JssContext: Context<{
   disableStylesGeneration: boolean
 }>
 
-interface WithStylesProps<S extends Styles | ((theme: unknown) => Styles)> {
-  classes: Classes<S extends ((theme: unknown) => Styles) ? keyof ReturnType<S> : keyof S>
+interface WithStylesProps<S extends Styles | ((theme: any) => Styles)> {
+  classes: Classes<S extends (theme: any) => Styles ? keyof ReturnType<S> : keyof S>
 }
 /**
  * @deprecated Please use `WithStylesProps` instead
  */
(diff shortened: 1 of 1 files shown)


vismu:
> Is this the best way to do it in ts? Is user able to define a strict theme? It sounds like it should be a generic but I only use flowtype, so I have no idea
> cc @moshest can you advise us here pls?

Of course we can add second generic for WithStylesProps to specify theme object in more strict way.
```ts
interface WithStylesProps<S extends Styles | ((theme: T) => Styles), T> {
  classes: Classes<S extends ((theme: T) => Styles) ? keyof ReturnType<S> : keyof S>
}
```
But, actually, we don't need this Theme generic to calculate our classes result. 
So it will be a little bit overhead and do the usage more annoying.
```ts
interface Props extends WithStylesProps<typeof styles, Theme> {}
```
vs
```ts
interface Props extends WithStylesProps<typeof styles> {}
```

kof:
Mb I misunderstand something, is user able to specify the theme type right now and let ts show this type everywhere it is used?

vismu:
In my app I have a Theme interface and use it while defining of dynamic styles for component.
```ts
const styles = (theme: Theme) => {...}
```

moshest:
I agree.

We should change theme type to `any`:

```ts
//                                                    ↧ this one
interface WithStylesProps<S extends Styles | ((theme: any) => Styles)> {
  classes: Classes<S extends ((theme: any) => Styles) ? keyof ReturnType<S> : keyof S>
}
```

kof:
Alright, merged!

vismu:
> Alright, merged!

Great! Can you tell me how frequently new versions of `react-jss` are published.

Issue #1255, opened 2020-01-06
[react-jss] Version 10.0.3 contains breaking change in Typescript definitions


pajasevi:
__Expected behavior:__
In version 10.0.0 and above, the TS definition clearly states that, if user wants to use theme in the function, Theme type must be supplied to the generic `createUseStyles` type as follows:
```ts
declare function createUseStyles<T, C extends string>(
  styles: (theme: T) => Record<C, Style | string>,
  options?: {
    index?: number
    name?: string
    theming?: Theming<T>
  } & StyleSheetFactoryOptions
): (data?: any) => Record<C, string>
```
Thus user could use types styles with provided theme as follows:
```ts
const useStyles = createUseStyles<Theme, "checkboxWrapper" | "input">((theme: Theme) => ({
	checkboxWrapper: {
		display: "inline-flex",
		alignItems: "center",
		fontSize: theme.utils.pxToRem(14),
	},
	input: {
		marginRight: theme.utils.pxToRem(6),
	}
}));
```
And the `useStyles` hook would return proper classNames.

__Describe the bug:__
In react-jss@10.0.3 a new typescript definitions were introduced which are breaking against previous behavior. The Theme argument no longer needs to be supplied to the generic which means that every usage of `createUseStyles` is now broken.
```ts
declare function createUseStyles<C extends string = string>(
  styles: ((theme: any) => Styles<C>),
  options?: CreateUseStylesOptions
): (data?: unknown) => Classes<C>
```

__Recommended fix__
Revert this behavior under 10.0.4 version and release a new major version with this breaking change, according to the semver.

__Codesandbox link:__ 
Please create a codesandbox.io with the issue. Make it as minimal as possible as this will help us find the bug quicker.

__Versions (please complete the following information):__
 - react-jss: 10.0.3
 - Browser [e.g. chrome, safari]: yes
 - OS [e.g. Windows, macOS]: yes
Feel free to add any additional versions which you may think are relevant to the bug.

pajasevi:
Long time, no answer. Are you even gonna address this issue?

HenriBeck:
@kof Do you think this is worth a completely new major version? This is only breaking change about types (Generics) (not actually running code) and I personally don't think it should be hard to migrate for most users.

yhaskell:
From what I saw in the code, that was specifically typing issue. js code still accepts second argument

pajasevi:
[...]

PR #1352, opened 2020-06-09
fix `createUseStyles` types

moshest:
Fixing TypeScript types for `createUseStyles` and `JssStyle`

## Todo

- [x] Add tests if possible - new test file under `packages/jss/tests/types/Styles.ts`
- [ ] Add documentation

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -4,16 +4,21 @@ import * as css from 'csstype'
 type FnValue<R> = R | ((data: any) => R)
 
 type NormalCssProperties = css.Properties<string | number>
-type CssProperties = {[K in keyof NormalCssProperties]: FnValue<NormalCssProperties[K] | JssValue>}
+type NormalCssValues<K> = K extends keyof NormalCssProperties
+  ? NormalCssProperties[K] | JssValue
+  : JssValue
 
-// Jss Style definitions
-type JssStyleP = {
-  [key: string]: FnValue<JssValue | JssStyleP>
+export type JssStyle = {
+  [K in keyof NormalCssProperties | string]:
+    | NormalCssValues<K>
+    | JssStyle
+    | ((data: any) => NormalCssValues<K> | JssStyle)
 }
 
-export type JssStyle = CssProperties & JssStyleP
-
-export type Styles<Name extends string | number | symbol = string> = Record<Name, JssStyle | string>
+export type Styles<Name extends string | number | symbol = string> = Record<
+  Name,
+  FnValue<JssStyle | string>
+>
 export type Classes<Name extends string | number | symbol = string> = Record<Name, string>
 export type Keyframes<Name extends string = string> = Record<Name, string>
 
--- a/packages/jss/tests/types/Styles.ts
+++ b/packages/jss/tests/types/Styles.ts
@@ -1,0 +1,56 @@
+import {Styles} from '../../src'
+
+interface Props {
+  flag?: boolean
+}
+
+const styles: Styles = {
+  basic: {
+    textAlign: 'center',
+    display: 'flex',
+    width: 500,
+    justifyContent: 'center'
+  },
+  property: {
+    textAlign: 'center',
+    display: 'flex',
+    width: '100%',
+    justifyContent: (props: Props) => (props.flag ? 'center' : 'left')
+  },
+  inner: {
+    textAlign: 'center',
+    display: 'flex',
+    width: '100%',
+
+    '&.foo': {
+      fontSize: 12
+    }
+  },
+  func: (props: Props) => ({
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: 'center',
+    alignItems: 'center',
+    width: '96px',
+    cursor: 'pointer',
+    position: 'relative',
+    pointerEvents: props.flag ? 'none' : null
+  }),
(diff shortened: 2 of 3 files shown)


moshest:
I don't think this test failure is related to my changes.. 

```
09 06 2020 21:25:16.741:ERROR [launcher]: Cannot load browser "BS_Safari Yosemite"!
440  Error: Password is required.
```

kof:
That's because we are not able to run browserstack tests for external contributors

kof:
Is this a breaking change for the user?

moshest:
Don't think so. Types should never trigger breaking changes.

I notice some new errors on my code if I wrote a function that return `undefined` instead of `null` (as `any !== JssValue`). Not sure it's a breaking change, fells more like a fix.

If that not a fix and JSS support also `undefined` values, we should update `JssValue` as well and add `undefined` to the list.

Other then that, it should accept basically anything, including custom css properties like: 

```ts
  root: {
   fontSize: 12,
   customFont: 'foo' // <- non-existing property
  }
});
```

kof:
> Types should never trigger breaking changes.

https://github.com/cssinjs/jss/issues/1255#issuecomment-573434369

kof:
If someone suddenly can't compile types after updating a minor change, how is this not breaking users code?

moshest:
I think that if the user did something wrong and the type check suddenly catch that on a version update it's a good thing. It's better to failed on compile time then on runtime.

As far as I know, those types are align with the current implementation. The only contradiction that I found is on the `JssValue` type and it's not directly related to this fix.

It's seems that JSS works with `undefined` values but it's not allowed on the TS types. So the question is, do we want to allow `undefined` as a valid `JssValue` or it's consider a bug?

kof:
Falsy values are supported for the rule function (function returning an object with properties/values)
https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-rule-value-function/src/index.js#L58

But  https://github.com/cssinjs/jss/blob/master/packages/jss/src/plugins/styleRule.js#L45 method styleRule.prop  can set and get in one method and so if the second argument is undefined, it will act as a  getter, so it turns out we have an inconsistency with function values https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-rule-value-function/src/index.js#L75 because if `fnValues[prop](data)` returns an undefined `styleRule.prop(prop, fnValues[prop](data), options)` will act as a getter. 

To make it consistent we need to support falsy values [here](https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-rule-value-function/src/index.js#L75)

kof:
> I think that if the user did something wrong and the type check suddenly catch that on a version update it's a good thing. It's better to failed on compile time then on runtime.

2 questions:
1. Did user had actually wrong code that will now be uncovered in this case?
2. Should upgrading a non-major version result in any code changes require by the user? In theory no, even if user's code was wrong already. It was tested the way it was written and considered OK. Now if user needs to change their code in a minor upgrade, they will have to go through verification process and I am not sure this is expected from a minor upgrade.

kof:
The thing is any change is potentially a breaking change. Semver is here to communicate to the user: "you will have to change your code and verify it works".

moshest:
I don't want to change functionality in this PR. If `undefined` is only supported on functions I can fix the types to support that only:

```ts
export type JssStyle = {
  [K in keyof NormalCssProperties | string]:
    | NormalCssValues<K>
    | JssStyle
    | ((data: any) => NormalCssValues<K> | JssStyle | undefined)
}
```

To answer your questions:

1. Yes, following [your comment](https://github.com/cssinjs/jss/pull/1352#issuecomment-641859069) it's seems that user did something wrong if he provide `undefined` to `styleRule.prop` as this PR will catch that.
2. Since we didn't change any functionally in this PR, i don't think we need to bump the version and let the user recheck his code. If we will get some errors from the user about filled type check that actually work on runtime we can release another fix to address that but I idea is that it shouldn't happen.

kof:
> I don't want to change functionality in this PR. If undefined is only supported on functions I can fix the types to support that only:

In this case right now we support falsy values for function rules and only a null for function values.

moshest:
> > I don't want to change functionality in this PR. If undefined is only supported on functions I can fix the types to support that only:
> 
> In this case right now we support falsy values for function rules and only a null for function values.

What's the different between "function rules" and "function values"? It's not clear to me..

kof:
https://cssinjs.org/jss-plugin-rule-value-function

kof:
Regarding breaking changes in types: https://twitter.com/FezVrasta/status/1270648268012339203?s=20 argument resonates with me

In theory, it should have been breaking change, but both ts and flow are a mess in terms of breaking changes so we shouldn't care either breaking users types in minor versions.

afzalsayed96:
I just upgraded a project from `10.0.1` to `10.4.0` for a bug fix and a ton of shit broke 😔

kof:
@afzalsayed96 sorry to hear that, you will have to report separate issues, also check out the changelog between the versions

afzalsayed96:
Any plans to support falsy values in function values? I'm widely using it something like below pattern everywhere in my codebase:

```JS
block: ({ height, width }) => ({
    height,
    width
  })
```

moshest:
@afzalsayed96 can you attach the typescript errors you get (preferably on a new issue, linked to this one)?

PR #1353, opened 2020-06-10
fix(react-jss/types): fix react default props types support

vismu:
Hello, dear colleagues!
I fixed an issue connected with react default props in typescript.
You can read more about processing of react default props in typescript here: 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx
In short, they calculate resulted props with help of `JSX. LibraryManagedAttributes` small function.

So, as you can see, in PR I used this helper to calculate component props types according official documentation.

Thanks for your attention!

## Todo

- [x] Add tests if possible
- [x] Add documentation

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,10 @@ Since you are interested in what happens next, in case, you work for a for-profi
 
 ---
 
+### Improvements
+
+- [react-jss] add properly react default props types calculation (https://github.com/cssinjs/jss/pull/1353)
+
 ## 10.2.0 (2020-6-3)
 
 ### Improvements
--- a/packages/react-jss/src/index.d.ts
+++ b/packages/react-jss/src/index.d.ts
@@ -66,21 +66,19 @@ declare function createUseStyles<Theme = DefaultTheme, C extends string = string
   options?: CreateUseStylesOptions<Theme>
 ): (data?: unknown) => Classes<C>
 
+type GetProps<C> = C extends ComponentType<infer P> ? P : never
+
 declare function withStyles<
   ClassNames extends string | number | symbol,
   S extends Styles<ClassNames> | ((theme: any) => Styles<ClassNames>)
 >(
   styles: S,
   options?: WithStylesOptions
-): <
-  Props extends {
-    classes: S extends (theme: any) => Styles<ClassNames>
-      ? Classes<keyof ReturnType<S>>
-      : Classes<ClassNames>
-  }
->(
-  comp: ComponentType<Props>
-) => ComponentType<Omit<Props, 'classes'> & {classes?: Partial<Props['classes']>}>
+): <C>(
+  comp: C
+) => ComponentType<
+  JSX.LibraryManagedAttributes<C, Omit<GetProps<C>, 'classes'> & Partial<WithStylesProps<S>>>
+>
 
 type Omit<T, K> = Pick<T, Exclude<keyof T, K>>
 
--- a/packages/react-jss/src/index.test.tsx
+++ b/packages/react-jss/src/index.test.tsx
@@ -1,0 +1,24 @@
+import React, {Component} from 'react'
+
+import WithStyles from '.'
+
+interface Props {
+  testProp: string
+}
+
+class TestComponent extends Component<Props> {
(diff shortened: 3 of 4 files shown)


vismu:
> Cool, one conflict to go and I am gonna merge it. Also Please check out other TS issues, mb there are simple fixes and I could get more stuff into that release

Conflict was resolved. 

Also I found a little issue that my test wasn't catch by `tsc` because of `tsx` extension. So I've added `tsx` files including in tsconfig.json. 

About other TS issues. I found only one more and my PR is here: https://github.com/cssinjs/jss/pull/1349

packages/react-jss/tests/defaultPropsWithStyles.tsx
```diff
@@ -0,0 +1,24 @@
+import React, {Component} from 'react'
```
kof, line 1: I wonder if we should keep all ts tests in one file, otherwise there will be a mixture of js and tsx tests in that folder and that's weird because tsx tests are not jest tests like in those js files.

I think ideally we should move everything typescript into some dir in each package and have there both interfaces and tests

kof: I realize this is not really part of the scope here, but with adding this file name it contributes to confusion

vismu: > I wonder if we should keep all ts tests in one file, otherwise there will be a mixture of js and tsx tests in that folder and that's weird because tsx tests are not jest tests like in those js files.
> 
> I think ideally we should move everything typescript into some dir in each package and have there both interfaces and tests


Oh, thats right. I can do a subfolder `types` just by example of https://github.com/cssinjs/jss/blob/master/packages/jss/tests/types/Styles.ts
Also If you want  I can change file name by your proposal.

kof: what about moving everything typescript into {packageName}/ts/ ... and inside for example
`packages/jss/ts/index.d.ts`, `packages/jss/ts/index.test.d.ts` or if tests are specific to some interface `packages/jss/ts/styles.test.ts` etc

vismu: Hmm, I think that such folder reorganising doesn't make sense. 
For me as for new contributor it is much clearer if all tests will be placed at `tests` folder including `ts` tests.

kof: Those ts tests are not the same kind of tests, in fact they are not tests, they are just code that uses types that we use because the code and tests itself is written without ts so nothing is using ts types.

kof: Also have you noticed that tests dir is only used for some type of tests, all unit tests are directly located next to the code e.g. JssProvider.js, JssProvider.test.js

vismu: > Also have you noticed that tests dir is only used for some type of tests, all unit tests are directly located next to the code e.g. JssProvider.js, JssProvider.test.js

Oh, I didn't noticed that. But, follow that logic, I should put my test near the withStyles.js file and call it withStyles.test.tsx.

kof: that is exactly the logic I am following, but the thing is this "tests" is testing the interface, so I consider the interfaces file to be "code" and that code using that interface - test

vismu: Nice, think I understood your point. So I will put my test near the `index.d.ts` and call it `index.test.tsx`. Are you ok with that?

kof: yeah, for now that would be ok unless you want to move all ts things into {packageName}/ts

vismu: Done. I think it's not good idea move index.d.ts to another folder, because it should be placed right near index.js file. Thats how typescript will associate interfaces with implementation.

kof:
merged!

PR #1355, opened 2020-06-16
[react-jss] TypeScript: Added support for innerRef prop

sorteam:
Can't pass innerRef prop

Add support for innerRef in types

## Todo

- [x] Add tests if possible
- [x] Add documentation


vismu:
Yep, actually, your last commit do right change, because `innerRef` prop is added only to HOC component. 
I will have one request.
Pls, add check of ability to pass `innerRef` prop to component `TestComponentWitStyles` at `index.test.tsx` file.

sorteam:
@moshest check this please..

moshest:
Can we have the tests use a real world example instead of `any`?

kof:
merged

PR #1362, opened 2020-07-07
Fix Flow typing for CreateGenerateId

joshclow:
Brings this into alignment with the Typescript annotation for this same function:
```
export type CreateGenerateId = (options?: CreateGenerateIdOptions) => GenerateId
```

I didn't open an issue, but can if that would be helpful for tracking

The TS and Flow types for GreateGenerateId disagree on the optionality of the options parameter. It *appears* to me that options is meant to be optional and thus the TS annotation is more correct. I've made the corresponding update to the flow type.


kof:
thanks!

PR #1376, opened 2020-08-02
fix: Made it possible for windows to run yarn build

RobertAron:
Running `yarn build` on windows machines causes errors due to the way external modules are calculated. This is the error message:

```
(plugin size-snapshot) Error: ModuleNotFoundError: Module not found: Error: Can't resolve './Jss.js' in '/'
```

I found another library that had the exact same problem. This PR mirrors the fix that they created.

https://github.com/doczjs/docz/issues/783

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -31,7 +31,8 @@ Object.keys(pkg.peerDependencies || {}).forEach(key => {
   }
 })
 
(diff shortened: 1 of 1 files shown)


kof:
thanks!

kof:
released in 10.4.0

PR #1378, opened 2020-08-03
fix #1360

### LINKED ISSUE #1360: Memory leak with dynamic styles
__Expected behavior:__
My APP must render very frequently by our realtime data.

__Describe the bug:__
My APP has memory leak on Firefox, after using material-ui. 
This is the same issue in material-ui https://github.com/mui-org/material-ui/issues/21528

__Codesandbox link:__ 
 [https://codesandbox.io/s/dazzling-visvesvaraya-eczqe?file=/src/Tab.js](https://codesandbox.io/s/dazzling-visvesvaraya-eczqe?file=/src/Tab.js) 

 - jss: "^10.0.3"
 - Browser [e.g. chrome, safari]:  firefox
 - OS [e.g. Windows, macOS]: macOS

Jokcy:
fix https://github.com/cssinjs/jss/issues/1360

This is early preview PR, just use to talk about whether this solution will work, or may it will cause other problem.

### DIFF
--- a/packages/jss-plugin-nested/src/index.js
+++ b/packages/jss-plugin-nested/src/index.js
@@ -94,9 +94,14 @@ export default function jssNested(): Plugin {
         // Replace all $refs.
         selector = selector.replace(refRegExp, replaceRef)
 
+        // remove same rule before add
+        if (container.getRule(selector)) {
+          container.deleteRule(selector)
+        }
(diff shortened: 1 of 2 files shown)


kof:
I think the solution should be to identify when we the same rules and not add them at all.

kof:
released in 10.4.0

PR #1382, opened 2020-08-12
Upgrade flow-bin to 0.131.0 and migrate types

kof:
## Todo

- [x] Upgrade theming package

```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/theming/src/create-with-theme.js:23:18

Cannot spread props because Flow cannot determine a type for props [1]. object type [2] is
inexact, so it may contain ref with a type that conflicts with ref's definition in props [1].
Try making object type [2] exact. [cannot-spread-inexact]

     node_modules/theming/src/create-with-theme.js
      20│             '[theming] Please use withTheme only with the ThemeProvider'
      21│           );
      22│
 [1]  23│           return <Component theme={theme} ref={ref} {...props} />;
      24│         }}
      25│       </context.Consumer>
      26│     ));

     /private/tmp/flow/flowlib_36a9420e/react.js
 [2] 159│ declare type React$ComponentType<-Config> = React$AbstractComponent<Config, mixed>;
```

### DIFF
--- a/packages/css-jss/src/createCss.test.js
+++ b/packages/css-jss/src/createCss.test.js
@@ -3,7 +3,7 @@ import expect from 'expect.js'
 import {stripIndent} from 'common-tags'
 import {create as createJss, type StyleSheet} from 'jss'
 import {createGenerateId} from '../../../tests/utils'
-import {create as createCss, type Css} from './index'
+import {create as createCss} from './index'
 import {type StyleArg} from './types'
 
 type CssTestType = {|
--- a/packages/css-jss/src/createCss.test.js
+++ b/packages/css-jss/src/createCss.test.js
@@ -10,7 +10,7 @@ describe('css-jss', () => {
 
   beforeEach(() => {
     const jss = createJss({createGenerateId})
-    css = createCss(jss)
+    css = ((createCss(jss): any): Function)
   })
 
   it('should accept a single style object argument', () => {
--- a/packages/css-jss/src/createCss.test.js
+++ b/packages/css-jss/src/createCss.test.js
@@ -1,16 +1,22 @@
 // @flow
 import expect from 'expect.js'
 import {stripIndent} from 'common-tags'
-import {create as createJss} from 'jss'
+import {create as createJss, type StyleSheet} from 'jss'
 import {createGenerateId} from '../../../tests/utils'
-import {create as createCss} from './index'
+import {create as createCss, type Css} from './index'
+import {type StyleArg} from './types'
+
+type CssTestType = {|
+  (...args: StyleArg[]): string,
+  getSheet: () => StyleSheet
+|}
 
 describe('css-jss', () => {
   let css
 
   beforeEach(() => {
     const jss = createJss({createGenerateId})
-    css = ((createCss(jss): any): Function)
+    css = (createCss(jss): CssTestType)
   })
 
   it('should accept a single style object argument', () => {
--- a/packages/jss/src/SheetsRegistry.js
+++ b/packages/jss/src/SheetsRegistry.js
@@ -56,7 +56,7 @@ export default class SheetsRegistry {
   /**
    * Convert all attached sheets to a CSS string.
    */
-  toString({attached, ...options}: {|attached?: boolean|} & ToCssOptions = {}): string {
+  toString({attached, ...options}: {|attached?: boolean, ...ToCssOptions|} = {}): string {
     let css = ''
     for (let i = 0; i < this.registry.length; i++) {
       const sheet = this.registry[i]
--- a/packages/jss/src/SheetsRegistry.js
+++ b/packages/jss/src/SheetsRegistry.js
@@ -56,7 +56,7 @@ export default class SheetsRegistry {
   /**
    * Convert all attached sheets to a CSS string.
    */
-  toString({attached, ...options}: {attached?: boolean, ...ToCssOptions} = {}): string {
+  toString({attached, ...options}: {|attached?: boolean|} & ToCssOptions = {}): string {
     let css = ''
     for (let i = 0; i < this.registry.length; i++) {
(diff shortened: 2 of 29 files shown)


packages/css-jss/src/createCss.test.js
```diff
@@ -10,7 +10,7 @@ describe('css-jss', () => {
 
   beforeEach(() => {
     const jss = createJss({createGenerateId})
-    css = createCss(jss)
+    css = ((createCss(jss): any): Function)
```
TrySound, line 13: Function is an alias for any. You can avoid double casting.

packages/jss/src/SheetsRegistry.js
```diff
@@ -56,7 +56,7 @@ export default class SheetsRegistry {
   /**
    * Convert all attached sheets to a CSS string.
    */
-  toString({attached, ...options}: {attached?: boolean, ...ToCssOptions} = {}): string {
+  toString({attached, ...options}: {|attached?: boolean|} & ToCssOptions = {}): string {
```
TrySound, line 59: Why not make ToCssOptions exact and revert to type spread?

packages/css-jss/src/createCss.test.js
kof: yeah, probably, was more intuitive to do that, the actual reason for any here is I have a static property on that function that is only exposed for tests and I haven't figured out how to describe that properly, since you are here, can you show me a type without casting for that css value?

packages/jss/src/types/jss.js
```diff
@@ -188,9 +188,9 @@ export type StyleSheetOptions = {|
   Renderer?: Class<Renderer> | null,
   insertionPoint?: InsertionPoint,
   jss: Jss
-|}
+}
```
TrySound, line 191: Use explicit inexact type syntax `{value: number, ...}`

packages/jss/src/SheetsRegistry.js
kof: try it, it will show up with other errors and this seemed to be the correct choice

packages/jss/src/types/jss.js
kof: why? it's by default inexact

packages/css-jss/src/createCss.test.js
TrySound: This should work.
```
type Fn = {|
  (arg: number) => void,
  staticOne: number
|}
```

kof: there is no such syntax with function as you showed

packages/jss/src/SheetsRegistry.js
TrySound: Inexact type is almost never correct choice. It's a hack for javascript dynamism.
This will solve the problem with exact type. `{...null}` is sealed object, can be shaped with explicit type annotation.
```
  options: ToCssOptions = {...null}

```

packages/css-jss/src/createCss.test.js
TrySound: Sorry, corrected

packages/jss/src/types/jss.js
TrySound: Not for long. Flow introduced explicit inexact syntax to eventually migrate to exact by default. Btw it can be enabled optionally now.

TrySound: There is also a lint rule
https://flow.org/en/docs/linting/rule-reference/#toc-implicit-inexact-object

packages/jss/src/SheetsRegistry.js
kof: this will add a bunch of stuff to the build I can not afford in a lib

TrySound: Only one case. Which stuff?

kof:
```
    _proto.toString = function toString(_temp) {
      var _ref = _temp === void 0 ? _extends({}, null) : _temp,
          attached = _ref.attached,
          options = _objectWithoutPropertiesLoose(_ref, ["attached"]);
```
vs

```
    _proto.toString = function toString(_temp) {
      var _ref = _temp === void 0 ? {} : _temp,
          attached = _ref.attached,
          options = _objectWithoutPropertiesLoose(_ref, ["attached"]);
```

TrySound: 🤷‍♂️

packages/jss/src/types/jss.js
kof: when they do it we will have to change it everywhere, I take it as a separate task

packages/jss/src/SheetsRegistry.js
TrySound: Cast empty object to `any` then.

kof: any other options without additional runtime overhead?

packages/css-jss/src/createCss.test.js
kof: worked

packages/jss/src/types/jss.js
TrySound: Is it necessary to make it inexact? It does not solve theme problem.

kof: sec, trying to fix other places which led to the errors ...

kof: changed it everywhere to be exact, had to dig through dozens of functions

kof:
Finally finished upgrading dependencies, thanks @TrySound for the CR!

kof:
released in 10.4.0

PR #1387, opened 2020-08-27
docs(csp): update style-loader config

Hongbo-Miao:
The doc is using old style-loader API.

Based on the new API, `attrs` has been changed to `attributes` https://webpack.js.org/loaders/style-loader/#attributes


kof:
thanks

Hongbo-Miao:
No problem : )

PR #1388, opened 2020-08-31
Fix prefix after value-function

vlazh:
Prefixing of value-function is not worked. So it fix it. But I'm not sure if I did it right.

## Todo

- [ ] Add tests if possible
- [ ] Add documentation


vlazh:
It seems like CI errors, not code.

kof:
Thanks for the PR, looks correct but I can't merge it unless we add a test!

PR #1391, opened 2020-09-09
Update Classname Example

Chartieer:
Adding a section explaining the name prop in the options object to add classname prefix on Module/Component level

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -155,6 +155,66 @@ The above code will compile to
 
 ```html
 <div id="root">
+  <button class="myButton-1-25">
+    <span class="myLabel-1-26">
+      Submit
+    </span>
+  </button>
+</div>
+```
+
+and
+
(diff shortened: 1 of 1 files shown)


docs/react-jss.md
```diff
@@ -153,6 +153,66 @@ const App = () => <Button fontStyle="italic">Submit</Button>
...
+  padding: 10px;
+}
+.myLabel-1-26 {
+  display: block;
+  color: red;
+  font-weight: bold;
+  font-style: italic;
+}
+```
+## Adding Component Prefix Option
```
kof, line 179: 1. I would probably write them lower case
2. headlines need to be added to the TOC above manually

kof (changes requested):
Thanks, one more change and we are good to go

Issue #1035, opened 2019-02-15
[RFC] Move to Typescript


HenriBeck:
With more OS projects now planning to move to Typescript (jest, Vue.js, etc.), I think it would be time to think about doing the same. Typescript has really matured and now with eslint and babel both supporting Typescript as well, it should be future proof.

Additionally, I think the tooling around Typescript (speed, autocompletion) etc. is far head of flow currently.

I would start with "rewriting" smaller packages (like `css-functions` and `theming`) in the beginning.
Also, I started writing the new docs with Typescript.

__Are you willing to implement it?__
Yes

TrySound:
Besides better tooling I saw a lot of unsoundness in TS which is "won't fix" when flow to gonna be and actually already safer than TS in some cases. For me it doesn't worth to switch.
I'm quite happy with flow in big project and open source.

HenriBeck:
Why would Typescript be less type-safe?

TrySound:
Here guys from flow team provides some examples
https://discord.gg/zGY9gW

For example TypeScript doesn't have exact types at all (https://github.com/Microsoft/TypeScript/issues/12936) when flow is gonna have them by default. Object spread is unsound (https://github.com/Microsoft/TypeScript/issues/15454) when flow is already more sound here and is gonna improve experience with exact by default.

![image](https://user-images.githubusercontent.com/5635476/52887392-949ad880-3188-11e9-89af-f1b10188dc2a.png)

TrySound:
Flow has some problems but flow team is quite active and open last few weeks. They are going to the right direction but their plan is time consuming. Now they have resources to work with community.

kof:
That's interesting, do you know how many people work on flow full time? I feel like flow with strict mode and better error descriptions would actually make the whole thing actually quite competitive. It's just a matter if companies are willing to invest enough to improve.

TrySound:
This is just now. There are more people there
![image](https://user-images.githubusercontent.com/5635476/52889438-fe1ee500-3190-11e9-9fac-0aa03025e462.png)

kof:
Fulltime contributors?

TrySound:
It's a chat with flow team (facebook guys). The link I put above.

HenriBeck:
I think flow error descriptions and reporting is quite better (especially when it comes to react props) than Typescripts.

What I now dislike the most about it is how slow the editor integration is and that types are an afterthought.

TrySound:
Which editor do you use?

[...]

PR #1402, opened 2020-10-05
Restore TypeScript support for Observable styles

appsforartists:

```typescript
const jss = createJSS(createDefaultJSSPreset());
jss.createStyleSheet(
  {
    myClassName: someObservable$,
  },
  {
    link: true,
  },
).attach();
```

should work in TypeScript, and [used to work with `@types/jss`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/de655960b603d6b47f7030674f084780c76e045f/types/jss/css.d.ts).  JSS now supplies its own types, which don't include `Observable` support.  This PR adds them back.

As @worudso alluded to in #1035, a better solution would only include the `Observable` type if users had installed the Observable plugin.  Otherwise, users could be frustrated if the typecheck passes for `createStyleSheet(someObservables)` but it fails at runtime.

However, that solution would require bigger and more thoughtful changes to the typings.  This PR is a quick fix that addresses the regression from `@types/jss`, while leaving the work of a refactor to a future PR.

## Todo

- [x] Add tests if possible
- [ ] ~Add documentation~

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 Since you are interested in what happens next, in case, you work for a for-profit company that benefits from using the project, please consider supporting it on https://opencollective.com/jss.
 
+### Bug fixes
+
+- [jss] Restores TypeScript support for Observable styles [1402](https://github.com/cssinjs/jss/pull/1402)
+
 ---
 
 ## 10.4.0 (2020-8-14)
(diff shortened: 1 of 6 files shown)


kof:
lgtm, thanks!

kof:
released https://github.com/cssinjs/jss/releases/tag/v10.5.0

PR #1403, opened 2020-10-07
feat(units): add `gap`

### LINKED ISSUE #1348: CSS Property `gap` not receiving a default unit
__Expected behavior:__
- CSS Property `gap` would receive a default unit of `px` for example
- Stretch goal, add `grid-gap` to maintain compatibility

__Describe the bug:__
A clear and concise description of what happened instead of the expected behavior.
- CSS Property `gap` not receiving a default unit, but is just a number making it invalid css

__Codesandbox link:__ 
Please create a [codesandbox.io](https://codesandbox.io/) with the issue. Make it as minimal as possible as this will help us find the bug quicker.
https://codesandbox.io/s/muddy-lake-h4k93?file=/src/App.tsx
`gridGap` is commented out to show the difference

 - jss: 10.1.1
 - Browser [e.g. chrome, safari]: chrome 81
 - OS [e.g. Windows, macOS]: mac

According to MDN, `grid-gap` is now being replaced with just `gap` in order to get ready for full support of `gap` for `flexbox`: https://developer.mozilla.org/en-US/docs/Web/CSS/gap
However, they also suggest that both properties be used in the mean time. 

For context, I made this issue originally in [Material-UI](https://github.com/mui-org/material-ui/issues/21235) and was directed this direction

I'm not sure where we would add code? A default value needs to be added to [jss-plugin-default-unit](https://github.com/cssinjs/jss/blob/d9cca31ad2d058aa0108fa6ec48fe872b0e7e886/packages/jss-plugin-default-unit/src/defaultUnits.js#L126), but for compatibility, where could we add a `grid-gap` property for every `gap` used? It isn't a vendor prefix, but css-vendor would tell if it was needed.

If I could be pointed in the right direction, I'd love to do a PR for this.
Thanks!

moshfeu:

resolve #1348

Add gap to defaultUnits

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/jss-plugin-default-unit/src/defaultUnits.js
+++ b/packages/jss-plugin-default-unit/src/defaultUnits.js
@@ -122,6 +122,7 @@ export default {
   // Some random properties
   'shape-margin': px,
   size: px,
+  gap: px,
 
   // Grid properties
   grid: px,


kof:
I would place it right after grid-gap, since the later was deprecated, also not sure if we still need to support grid-gap

moshfeu:
I wasn't sure because it also affect on `flex`, let me know.

kof:
merged, thanks

kof:
https://github.com/cssinjs/jss/releases/tag/v10.5.0

Issue #979, opened 2019-01-12
The tests are published to npm


oliviertassinari:
__Expected behavior:__
We can reduce the size of the npm packages by not publishing the tests.

__Describe the bug:__
Here is one example:
https://unpkg.com/jss-plugin-camel-case@10.0.0-alpha.7/src/index.test.js

__Versions (please complete the following information):__
 - jss: 10.0.0-alpha.7

kof:
We should definitely exclude .test.js from files in package.json

oliviertassinari:
If we had a `priority: low` flag. I would have added it. I think that it can be a good first issue flag for a potential new contributor :).

oliviertassinari:
I would encourage this solution:
**.npmignore**
```
/*
!/src/*.js
!/dist/*.js
*.test.js
```

kof:
What is wrong with complexity:low flag? I don't think the priority is good term for this. Priority is different depending on a consumer. Complexity of fixing something is relative to the project itself.

oliviertassinari:
Maybe replacing `complexity:low` with the convention: `good first issue`? 
Yes, complexity !== value

kof:
We went away from using .npmignore because it was much harder to exclude things than to select things which need to be included. We just forgot that we have .test.js files as we moved to a monorepo, because we moved in projects which use this structure.

kof:
We had a label like "good first issue" before and no single time people contributed to them, so I removed them because its just noise.

kof:
I consider all complexity:low issues a good first issue and its easy to find if one looks through issues.

kof:
NPM files supports glob pattern https://docs.npmjs.com/files/package.json#files so we can exclude those in it as well.

mdsadiq:
i think this issue is fixed already 😅

kof:
possible :)

PR #1406, opened 2020-10-10
perf: remove test files from the package

RobertAron:
https://github.com/cssinjs/jss/issues/979
Prevent the test files from being published to NPM.

I'm not super comfortable with how lerna works, but I believe this is what's needed.

## Todo

- [X] Add tests if possible (N/A)
- [X] Add changelog if users should know about the change (N/A)
- [X] Add documentation (N/A)

### DIFF
--- a/packages/css-jss/package.json
+++ b/packages/css-jss/package.json
@@ -18,7 +18,8 @@
   },
   "files": [
     "dist",
-    "src/**/!(*.test.*)",
+    "src",
+    "!*.test.*",
     "LICENSE"
   ],
   "keywords": [
--- a/packages/css-jss/package.json
+++ b/packages/css-jss/package.json
@@ -17,6 +17,7 @@
     "url": "https://github.com/cssinjs/jss/issues/new"
   },
   "files": [
+    "!*.test.js",
     "dist",
     "src",
     "LICENSE"
--- a/packages/css-jss/package.json
+++ b/packages/css-jss/package.json
@@ -19,8 +19,7 @@
   "files": [
     "dist",
     "src",
-    "!*.test.*",
-    "LICENSE"
+    "!*.test.*"
   ],
   "keywords": [
     "jss",
--- a/packages/css-jss/package.json
+++ b/packages/css-jss/package.json
@@ -17,9 +17,8 @@
     "url": "https://github.com/cssinjs/jss/issues/new"
   },
   "files": [
-    "!*.test.js",
     "dist",
-    "src",
+    "src/**/!(*.test).js",
     "LICENSE"
   ],
   "keywords": [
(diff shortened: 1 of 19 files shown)


packages/css-jss/package.json
```diff
@@ -17,6 +17,7 @@
     "url": "https://github.com/cssinjs/jss/issues/new"
   },
   "files": [
+    "!*.test.js",
```
kof, line 20: files is actually defining what should be in the package, we don't need to define what should not be in the package in addition

kof: oh you are talking about the tests which are next to the src files

kof: I wonder as well if that works, because it's a negative match, so it will probably just do nothing and move on to src and still match everything

RobertAron: I tried with npm pack and thought it was working but just double-checked and I'm not so sure anymore...

I'll fiddle with it a bit more.

kof: otherwise I only see .npmignore as an option

RobertAron: I was able to glob ignore specifically the files that end in .test.js that are in the src folders. I could setup .npmignore instead if you think that's a cleaner approach.

RobertAron: Pretty confident it's in a good state now. [Here](https://www.digitalocean.com/community/tools/glob?comments=true&glob=src%2F%2A%2A%2F%21%28%2A.test.%2A%29&matches=false&tests=src%2Findex.js&tests=src%2Findex.test.js&tests=src%2Findex.test.tsx&tests=src%2Fwhatever.md&tests=src%2Fhi%2Findex.test&tests=src%2Fhi%2Findex.test.other&tests=src%2Futils%2Ftest.js&tests=src%2Futils%2Ftest.test.js&tests=src%2Futils%2Fother.js) are the cases I was thinking about.

kof: yeah, that glob looks scary, at this point I think .npmignore is cleaner

RobertAron: I agree it's a little scary 😬

Apparently [files included with the “package.json#files” field cannot be excluded through .npmignore or .gitignore](https://docs.npmjs.com/files/package.json). I wasn't sure if you meant to pull the `files` out of the `package.json` so I kept looking for a better solution.

I found someone suggested doing what I was doing originally so I went back to looking into it. Apparently what I did works, but you have to do it in a specific order. Let me know if that's an ok solution.


Also that link I included above mentions that files called `LICENSE` should be included no matter what you do, so I think it could be taken out of the array. For now, I left it in just in case there was some reason for it to be there.

kof:
LiCENSE is included automatically, no need for listing it

kof:
checked, it works, thank you

PR #1408, opened 2020-10-15
correct umd build fixes #1329

### LINKED ISSUE #1329: Including react-jss via systemjs; ReferenceError: module is not defined.
Hi! ✋

__Expected behavior:__
When connect library to the project via systemjs doesn't throw a reference error.

__Describe the bug:__
It happens because of two places in the code. If I wrap them into `try catch` - everything works great.

__dist/react-jss.js:line: 484__
```
module.exports = shouldUseNative() ? Object.assign : function (target, source) {
  	...
};
```
Pay attention on __shouldUseNative__ method (line: 484):
Снимок экрана 2020-04-12 в 18 17 33

and dist/react-jss.js:line:1180
```
{
    module.exports = require('./cjs/react-is.development.js');
}
```

__GitHub link:__ 
https://github.com/rocetspace/modules
`npm i && npm start` -> localhost:9000

 - jss: 10.1.1
- react-jss: 10.1.1
 - Browser [e.g. chrome, safari]: all browsers
 - OS [e.g. Windows, macOS]: all OS

Снимок экрана 2020-04-12 в 15 14 31

sscaff1:
#1329 

The umd build wasn't an actual umd build. It failed to run in the context of an Microfrontend project (i.e. systemJS) or in the browser. This PR fixes that.

## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/css-jss/.size-snapshot.json
+++ b/packages/css-jss/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/css-jss.js": {
-    "bundled": 59599,
-    "minified": 21252,
-    "gzipped": 7233
+    "bundled": 3429,
+    "minified": 1203,
+    "gzipped": 687
   },
   "dist/css-jss.min.js": {
-    "bundled": 58524,
-    "minified": 20657,
-    "gzipped": 6968
+    "bundled": 3429,
+    "minified": 1203,
+    "gzipped": 687
   },
   "dist/css-jss.cjs.js": {
     "bundled": 2968,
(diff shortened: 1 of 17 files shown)


kof:
@TrySound wdyt?

TrySound:
There are two changes in current pr
1. convert all node_modules from commonjs to es modules
2. import commonjs versions of babel helpers which has to be also converted to es modules

The second case is rather pointless and does not solve anything.
The first case solves the issue with untranspiled "object-assign" package a dependency of prop-types.

kof:
@sscaff1 do you mind updating the PR according to @TrySound's feedback?

kof:
please reopen, not sure what to do here

Issue #1363, opened 2020-07-16
Default unit not applied when value is 0


mitchellstern:
__Expected behavior:__
Default units should be applied to all numerical values, not just non-zero values. For instance, `transitionDuration: 300` should give `transition-duration: 300ms` and `transitionDuration: 0` should give `transition-duration: 0ms`.

__Describe the bug:__
When the value is 0, no unit is being applied, resulting in invalid styles like `transition-duration: 0`. The plain property value `0` is not recognized by the browser where time values are expected, and should have a unit according to https://developer.mozilla.org/en-US/docs/Web/CSS/time.

I believe the line responsible for this is the following:

https://github.com/cssinjs/jss/blob/4094410d82dfdae772e1c09f0cd187cb48fa1cdc/packages/jss-plugin-default-unit/src/index.js#L27

__Codesandbox link:__ 
https://codesandbox.io/s/cranky-gould-yuyq5?file=/src/App.tsx

Note the invalid style being produced by JSS on the indicated element:

<img width="379" alt="Screen Shot 2020-07-16 at 10 02 16 AM" src="https://user-images.githubusercontent.com/3311523/87680611-a43a9a00-c74b-11ea-8ede-e8a34e2584fc.png">

__Versions (please complete the following information):__
 - jss: 10.3.0
 - Browser [e.g. chrome, safari]: Chrome
 - OS [e.g. Windows, macOS]: macOS

reifnotreef:
I'm not sure if this is a bug.

This has to do with default units. This happens with lineHeight as well, it has no default unit so if you say ```lineHeight: 8``` instead of ```lineHeight: '8px'``` you're gonna have a BAD time.

mitchellstern:
Since no default unit is given for `line-height` in https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-default-unit/src/defaultUnits.js, any value will get translated without a unit, so the library is functioning as intended here. (Incidentally, the documentation at https://developer.mozilla.org/en-US/docs/Web/CSS/line-height suggests using unitless numbers for line-height values when possible, so this seems like a reasonable default.)

The issue I'm mentioning here is that properties that *should* have a unit don't have one when a value of 0 is supplied.

To illustrate the problem, here are some examples from the current version of the library.

Line height (no unit expected):
✅ `lineHeight: 0` gets turned into `line-height: 0`
✅ `lineHeight: 8` gets turned into `line-height: 8`

Border width (default unit of `px` expected):
[...]

PR #1413, opened 2020-10-22
[default-unit] Fix missing default unit for 0ms and 0%

mitchellstern:

Fixes https://github.com/cssinjs/jss/issues/1363.

Currently, no default units are applied to a zero value. While the CSS spec allows this for length units like `px` (see https://www.w3.org/TR/css-values-3/#lengths), this causes incorrect CSS to be generated for other units like `ms`, which the browser then ignores.

An example of this bug is provided in the linked issue (https://github.com/cssinjs/jss/issues/1363):

Screen Shot 2020-07-16 at 10 02 16 AM

This PR makes two changes to the logic in `packages/jss-plugin-default-unit/src/index.js`:
* It prevents a value of `0` from being returned early in the initial falsy value check.
* It adds a default unit (if available) to all numeric values, except for the special case of `0px`.

While generating `0px` instead of `0` would also be acceptable when the default unit is `px`, it may deviate from what developers are used to, so we omit it in this PR.

In order to pass the pre-commit checks, this PR also fixes a `yarn format:ci` failure in `docs/react-jss.md`.

## Todo

- [X] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/jss-plugin-default-unit/src/index.js
+++ b/packages/jss-plugin-default-unit/src/index.js
@@ -24,7 +24,7 @@ const units = addCamelCasedVersion(defaultUnits)
  * Recursive deep style passing function
  */
 function iterate(prop: string, value: any, options: Options) {
-  if (!value && value !== 0) return value
+  if (value == null) return value
 
   if (Array.isArray(value)) {
     for (let i = 0; i < value.length; i++) {
--- a/packages/jss-plugin-default-unit/src/index.js
+++ b/packages/jss-plugin-default-unit/src/index.js
@@ -1,6 +1,6 @@
 // @flow
 import type {Plugin} from 'jss'
-import defaultUnits from './defaultUnits'
+import defaultUnits, {px} from './defaultUnits'
 
 export type Options = {[key: string]: string | ((val: number) => string)}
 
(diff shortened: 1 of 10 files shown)


mitchellstern:
@kof @HenriBeck All tests are passing for this PR on my local machine, but I'm not sure how to trigger a Travis CI job on GitHub. Is there anything else I should be doing here?

packages/jss-plugin-default-unit/src/index.js
```diff
@@ -24,7 +24,7 @@ const units = addCamelCasedVersion(defaultUnits)
  * Recursive deep style passing function
  */
 function iterate(prop: string, value: any, options: Options) {
-  if (!value) return value
+  if (!value && value !== 0) return value
```
kof, line 27: I would replace it with: `if (value == null)`

mitchellstern: Sure thing, this has been addressed in https://github.com/cssinjs/jss/pull/1413/commits/3efc90e440742085c34f25f6de4bd4f7f8f63641.

kof:
lgtm, thanks!

kof:
https://github.com/cssinjs/jss/releases/tag/v10.5.0

PR #1423, opened 2020-11-12
fix: jssStyle type

hosseinmd:
typescript can't work with   [K in keyof NormalCssProperties | string]

type definition
## Todo

- [ ] Add tests if possible
- [ ] Add documentation

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -15,13 +15,17 @@ type NormalCssValues<K> = K extends keyof NormalCssProperties
   ? NormalCssProperties[K] | JssValue
   : JssValue
 
-export type JssStyle = {
-  [K in keyof NormalCssProperties | string]:
-    | NormalCssValues<K>
-    | JssStyle
-    | Func<NormalCssValues<K> | JssStyle | undefined>
-    | Observable<NormalCssValues<K> | JssStyle | undefined>
-}
+export type JssStyle =
+  | {
+      [K in keyof NormalCssProperties]:
+        | NormalCssValues<K>
+        | JssStyle
+        | Func<NormalCssValues<K> | JssStyle | undefined>
+        | Observable<NormalCssValues<K> | JssStyle | undefined>
+    }
+  | {
+      [K: string]: JssStyle | Func<JssStyle> | Observable<JssStyle | undefined>
+    }
 
 export type Styles<Name extends string | number | symbol = string> = Record<
   Name,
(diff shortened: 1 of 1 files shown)


kof:
What is the actual error compiler was reporting? This fix seems fishy, need some ts experts to confirm cc @moshest

hosseinmd:
the bug is where intellisense no suggestions can't suggest CSS Properties.

hosseinmd:
Because we tell them k is CSS properties or string, which means any string.

kof:
oh I see, that makes sense

moshest:
Why we remove there the `Observable<NormalCssValues<K> | JssStyle | undefined>` part?

hosseinmd:
> Why we remove there the `Observable<NormalCssValues<K> | JssStyle | undefined>` part?

Hi, You are right, that is my mistake, I fix that.

packages/jss/src/index.d.ts
```diff
@@ -15,13 +15,17 @@ type NormalCssValues<K> = K extends keyof NormalCssProperties
...
+export type JssStyle =
+  | {
+      [K in keyof NormalCssProperties]:
+        | NormalCssValues<K>
+        | JssStyle
+        | Func<NormalCssValues<K> | JssStyle | undefined>
+        | Observable<NormalCssValues<K> | JssStyle | undefined>
+    }
+  | {
+      [K: string]: JssStyle | Func<JssStyle> | Observable<JssStyle | undefined>
```
kof, line 27: why is this value different than the first value, isn't it supposed to be the exact saame thing?

```
        | NormalCssValues<K>
        | JssStyle
        | Func<NormalCssValues<K> | JssStyle | undefined>
        | Observable<NormalCssValues<K> | JssStyle | undefined>
```

hosseinmd: Actually no, because the string is not a key of NormalCssValues.

hosseinmd: I just removed NormalCssValues

hosseinmd: NormalCssValues has no effect there

hosseinmd: But I missed jssValue

hosseinmd:
@kof please review the last commit

moshest:
lgtm

kof:
merged, thanks

hosseinmd:
Could you please publish master?

hosseinmd:
Please publish

kof:

PR #1425, opened 2020-11-13
Upgrade flow to 0.138.0

kof:


rollup.config.js
```diff
@@ -59,7 +59,20 @@ const commonjsOptions = {
   // https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports
   // https://github.com/reduxjs/react-redux/issues/643#issuecomment-285008041
   namedExports: {
-    react: ['Component', 'createContext']
+    react: [
```
kof, line 62: This is becoming annoying, especially if someone new tries to use a function from react and don't know about this whitelist.

I don't know exactly why, but it started failing in react-jss package and ssaying those are not exported, demanding to add them

Any ideas @TrySound

TrySound: https://github.com/cssinjs/jss/pull/1426

kof:
@TrySound seems like I got after the upgrade of rollup weird flow issues, seems like flow is not seeing the types from import any more, you might already know what that is

```
Importing a type from an untyped module makes it any and is not safe! Did you mean to add // @flow to the top of jss?
[untyped-type-import]

     1│ // @flow
     2│ /* eslint-disable no-use-before-define */
     3│ import warning from 'tiny-warning'
     4│ import type {Plugin} from 'jss'
     5│
     6│ const isObject = obj => obj && typeof obj === 'object' && !Array.isArray(obj)
     7│ const valueNs = `extendCurrValue${Date.now()}`
```

TrySound:
Fixed. Missed this.

kof:
Omg I would have died trying to fix it lol

TrySound:
.flow files were missing

PR #1427, opened 2020-11-14
make rule-value-function and global work together

### LINKED ISSUE #1335: Variable values in @global
__Is your feature request related to a problem? Please describe.__
I want to use variables values in @global
but if @global is a function then "Uncaught TypeError: Cannot set property 'fnStyle1587427462315' of null", if fields of @global are functions they are ignored
```
  '@global':  {
    '#sntch_block': {
      display: () => 'none',
    },
  },
});
```

__Describe the solution you'd like__
Variables values like already implemented but working inside @global

__Are you willing to implement it?__
Yes if it's simple to do (your code seems clear and easy to understand)

louisabraham:
Fix #1335 
## Implementation details

I attach an attribute `updateFun` to rules with type `'global'` that have a function as declation (i.e. rules that match both plugins).

I already had to modify `RuleList.js` to prevent the special case `rule.rules instanceof RuleList` from skipping the update, so I included the code that executes `updateFun` in `RuleList.js`. An important choice (that is easily modifiable) is that `plugins.onUpdate` will not be launched on the global rule that is updated with a function: **other plugins will not run on the updated properties**.

Another meaningful possibility would be to include the call to `updateFun` in the `onUpdate` method of the `rule-value-function`. This would allow us to call `plugins.onUpdate` from `RuleList.js`, so have other plugins run on the rule after update. I was not sure what interactions this could create with other plugins so I didn't handle it this way. 

It is my first contribution to a JS project, so please feel free to give me advice to improve my coding style :smiley:

## Example

```js
import jss, {SheetsRegistry} from 'jss'
import funplugin from 'jss-plugin-rule-value-function'
import globplugin from 'jss-plugin-global'

const jss_ = jss.default

jss_.use(funplugin.default(), globplugin.default())

const sheet = jss_.createStyleSheet({
  '@global': (data) => ({
    button: {
      width: 100,
      height: 100
    },
    '#sntch_block': {
      display: data.display
    }
  }),

  a: {
    display: 'inline',
    color: (data) => data.color
  },

  div: (data) => ({
    color: data.color
  })
})

const sheets = new SheetsRegistry()
sheets.add(sheet)
var str = sheets.toString()
console.log(str)

console.log('\n-----\n')

sheet.update({color: 'red', display: 'none'})

str = sheets.toString()
console.log(str)

console.log('\n-----\n')

sheet.update({color: 'blue', display: 'none'})

str = sheets.toString()
console.log(str)
```

gives 

```
.a-0-0-1 {
  display: inline;
}

-----

button {
  width: 100;
  height: 100;
}
#sntch_block {
  display: none;
}
.a-0-0-1 {
  display: inline;
  color: red;
}
.div-0-0-2 {
  color: red;
}

-----

button {
  width: 100;
  height: 100;
}
#sntch_block {
  display: none;
}
.a-0-0-1 {
  display: inline;
  color: blue;
}
.div-0-0-2 {
  color: blue;
}
```

### DIFF
--- a/packages/jss-plugin-rule-value-function/src/index.js
+++ b/packages/jss-plugin-rule-value-function/src/index.js
@@ -2,6 +2,7 @@
 import warning from 'tiny-warning'
 import {
   createRule,
+  RuleList,
   type Rule,
   type JssStyle,
   type RuleOptions,
@@ -22,7 +23,19 @@ export default function functionPlugin() {
     onCreateRule(name?: string, decl: JssStyle, options: RuleOptions): Rule | null {
       if (typeof decl !== 'function') return null
       const rule: StyleRuleWithRuleFunction = (createRule(name, {}, options): any)
-      rule[fnRuleNs] = decl
+      if (rule.type == 'global')
+        rule.updateFun = data => {
+          // compute styles
+          let styles = decl(data)
+          // Build a new RuleList that replaces the former one
+          rule.rules = new RuleList({
+            ...rule.options,
+            parent: rule
+          })
+          for (const selector in styles) rule.rules.add(selector, styles[selector])
+          rule.rules.process()
+        }
+      else rule[fnRuleNs] = decl
       return rule
     },
 
--- a/packages/jss/src/RuleList.js
+++ b/packages/jss/src/RuleList.js
@@ -200,8 +200,13 @@ export default class RuleList {
       sheet
     } = this.options
 
+    // updateFun is a workaround to use together
+    // jss-plugin-rule-value-function and jss-plugin-global
+    const hasUpdateFun = typeof rule.updateFun == 'function'
+
     // It is a rules container like for e.g. ConditionalRule.
-    if (rule.rules instanceof RuleList) {
+    // Ignore if there is an updateFun
+    if (!hasUpdateFun && rule.rules instanceof RuleList) {
       rule.rules.update(data, options)
       return
     }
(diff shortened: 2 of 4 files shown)


louisabraham:
I don't know how to add tests, but I think it would be a nice thing if someone has a bit of time!

kof:
See https://github.com/cssinjs/jss/issues/1335#issuecomment-727893215 regarding the syntax.

packages/jss/src/RuleList.js
```diff
@@ -200,16 +200,22 @@ export default class RuleList {
...
     } = this.options
 
+    // updateFun is a workaround to use together
+    // jss-plugin-rule-value-function and jss-plugin-global
+    const hasUpdateFun = typeof rule.updateFun == 'function'
+
     // It is a rules container like for e.g. ConditionalRule.
-    if (rule.rules instanceof RuleList) {
+    // Ignore if there is an updateFun
+    if (!hasUpdateFun && rule.rules instanceof RuleList) {
```
kof, line 209: breaks encapsulation between plugins and the core, core has no knowledge about style functions, so far we managed to run the plugin in a way that doesn't require this

kof: the idea is that the plugin reacts to onUpdate hook and calls the function

packages/jss-plugin-rule-value-function/src/index.js
```diff
@@ -22,7 +23,19 @@ export default function functionPlugin() {
     onCreateRule(name?: string, decl: JssStyle, options: RuleOptions): Rule | null {
       if (typeof decl !== 'function') return null
       const rule: StyleRuleWithRuleFunction = (createRule(name, {}, options): any)
-      rule[fnRuleNs] = decl
```
kof, line 25: this was how we hide the implementation detail of the interface that only this plugin knows how to handle

kof (changes requested):
Hey, thanks for the PR, we def. need to change the implementation

kof:
Would be great if someone added tests to rule-value-function package, so we can play with the implementation details and find a better way

PR #1433, opened 2020-12-15
Allow project to be in a directory that contains spaces

chapa:
Just ran into issue while trying to run `yarn build` and `yarn check:snapshots`, because `shell.exec` calls didn't expect project directory to contain spaces.


kof:
thanks

kof:

PR #1442, opened 2021-01-18
 Media query styles applied only to the first element in function

kof:

https://github.com/cssinjs/jss/issues/1320
https://github.com/cssinjs/jss/pull/1343


kof:
I have merged the test without the lib changes and it passes, which means it doesn't verify any fixes made by code changes.
Which means we are stuck at the testing phase. We need proper tests which verify what [those changes](https://github.com/cssinjs/jss/pull/1343) fix.

cc

PR #1446, opened 2021-01-23
Fix crash Chrome 88

### LINKED ISSUE #1445: margin: NaN crashes on Chrome 88
__Expected behavior:__

No crash

__Describe the bug:__

When rendering:

```jsx
import * as React from "react";
import { createUseStyles } from "react-jss";

const useStyles = createUseStyles((t) => ({
  container: {
    margin: NaN
  }
}));

export default function App() {
  useStyles();
  return <div />;
}
```

Chrome 88 crashes. `CSS.px(NaN)` isn't flying well.

> TypeError: Failed to execute 'px' on 'CSS': The provided double value is non-finite.

__Codesandbox link:__ 

https://codesandbox.io/s/react-jss-ts-forked-nl2rf?file=/src/App.tsx

 - jss: 10.5.0
 - Browser: Chrome >= 88
 
I have investigated the issue from https://github.com/mui-org/material-ui/issues/24519 and https://github.com/mui-org/material-ui/issues/24182. I'm not sure it's worth fixing here but I thought it would be interesting to report anyway.

It fails because `typeof NaN === 'number'`:

https://github.com/cssinjs/jss/blob/94c8f6c7e5eee7ce04c31ea2f7e318a5d5b2a082/packages/jss-plugin-default-unit/src/index.js#L43

The fix should be straightforward.

oliviertassinari:
Fixes #1445. The new test case fails without the fix (locally, not in the cloud as it's using an old version of Chrome).

I think that this bug will require a release sooner rather than later @kof.

### DIFF
--- a/packages/css-jss/.size-snapshot.json
+++ b/packages/css-jss/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "css-jss.js": {
-    "bundled": 61109,
-    "minified": 21924,
-    "gzipped": 7375
+    "bundled": 61133,
+    "minified": 21942,
+    "gzipped": 7382
   },
   "css-jss.min.js": {
-    "bundled": 60034,
-    "minified": 21301,
-    "gzipped": 7094
+    "bundled": 60058,
+    "minified": 21319,
+    "gzipped": 7102
   },
   "css-jss.cjs.js": {
     "bundled": 3034,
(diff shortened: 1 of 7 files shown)


kof:
lgtm, thanks

kof:

PR #1449, opened 2021-02-05
chore: edit globalThis.js to first check for `globalThis`

katelynsills:

N/A, but can create if that would be helpful. 

Now that [`globalThis` is generally available on all the latest platforms](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#browser_compatibility), `globalThis.js` can be improved by first checking for the presence of `globalThis`. 

My motivation: At Agoric, we have a secure version of JavaScript called SES. Under SES, `globalThis` exists, but `window`, `self`, and `Function('return this')()` are all undefined for security reasons. So, `jss` was failing for us, but by including `globalThis`, it will safely succeed even under our security requirements. 

## Todo

I would be happy to add a test, but I wasn't sure what to test and where to put it. It looks like there are no unit tests for `globalThis`. I'm testing locally that `jss` works for me under SES with this change.

- [ ] ~~Add documentation if it changes public API~~

### DIFF
--- a/packages/jss/.size-snapshot.json
+++ b/packages/jss/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "jss.js": {
-    "bundled": 63283,
-    "minified": 23316,
-    "gzipped": 7096
+    "bundled": 63344,
+    "minified": 23360,
+    "gzipped": 7098
   },
   "jss.min.js": {
-    "bundled": 61886,
-    "minified": 22548,
-    "gzipped": 6745
+    "bundled": 61947,
+    "minified": 22592,
+    "gzipped": 6748
   },
   "jss.cjs.js": {
-    "bundled": 58995,
-    "minified": 26159,
-    "gzipped": 7190
+    "bundled": 59056,
+    "minified": 26211,
+    "gzipped": 7200
   },
   "jss.esm.js": {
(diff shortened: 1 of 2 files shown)


kof:
What a great PR. I think testing this is not necessary, because to test this we would have to simulate that environment and we would be mostly testing the simulation, since there isn't much logic in there.

I think what we need is to add in inline comment exactly what you just said in the description, so that it's clear why it exists and we can merge it.

kof:
I just looked into the current version of core-js of this, https://github.com/zloirock/core-js/blob/master/packages/core-js/internals/global.js

katelynsills:
Ok, great! Would adding the following comment work:  `Now that `globalThis` is available on most platforms (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#browser_compatibility) we check for `globalThis` first.`?

kof:
That and also your use case with that execution environment

katelynsills:
> That and also your use case with that execution environment

That comment has been added!

kof:
merged, thank you, will be released with the next release at some point. I aggregate a bit.

kof:

PR #1454, opened 2021-02-17
make auto-complete of createUseStyles value work

### LINKED ISSUE #776: Check out how flow and typescript types from csstype can be used
@frenic did a generated types from MDN, now we should check it out and see what it needs to have a seemless integration for jss, react-jss, styled-jss

https://github.com/frenic/csstype

### LINKED ISSUE #806: styles typing (auto-completion, validation)

### LINKED ISSUE #1439: Types don't work on on `createUseStyles`
__Expected behavior:__
I expect the same type inference/autocompletion when using `createUseStyles` that I get when I use the `style` attribute in jsx. Here you can see the autocompletion offered on `<div style...`
image

__Describe the bug:__
There is no autocompletion when using `createUseStyles`
image

I'm a little bit confused because it appears as though a PR was merged to fix this behaviour but it doesn't seem to be working: https://github.com/cssinjs/jss/pull/1352

@moshest is the behavior you were aiming to fix with that PR? or am I misunderstanding something?

__Codesandbox link:__ 
https://codesandbox.io/s/react-jss-ts-13cqf

 - jss: 10.5.0
 - react-jss: 10.5.0
 - react: 17.0.0

### LINKED ISSUE #1452: Typescript definition for css properties
hi koff, your jss is awesome. I'm starting to leave the sass. writting mixins, functions, importing, etc are lot easier in JS.

But i prefer writting in typescript rather than js. It can avoid a lot of mistakes earlier.

Is there a TS definition for the css properties? rather than string, any, any[], any[][].
For an example is .fontWeight, only accept 'thin' | 'bold' | 'bolder' | 'thinner' | number,
no an arbitrary string allowed.

k-yle:

Closes #1439, Closes #1452, Closes #806, possibly fixes #776

Make auto-complete work for Typescript and/or VSCode users like in this screenshot:

image

This is **not** a breaking change. In the example above, folks can still type `display: "dhgjdkfhg"` and TS will accept it. This PR only makes auto-complete work. It deliberately doesn't add type-safety because that needs to wait for a semver-major release.

## Todo

- [ ] Add test that verifies the modified behaviour - **Not possible unfortunately** 
- [ ] Add documentation if it changes public API - not needed

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -50,8 +50,8 @@ export type CreateGenerateId = (options?: CreateGenerateIdOptions) => GenerateId
 export type GenerateId = (rule: Rule, sheet?: StyleSheet<string>) => string
 
 export type JssValue =
-  | string
-  | number
+  | (string & {})
+  | (number & {})
   | Array<string | number | Array<string | number> | '!important'>
   | null
   | false


moshest (approved):
Nice hack.

Worth mention this fix is related to https://github.com/microsoft/TypeScript/issues/29729

kof:
merged, thank you

kof:
Does this mean we can close that issue?

Issue #1455, opened 2021-02-17
Using Number.isNaN breaks JSS in IE11


GertSallaerts:
I wasn't able to find documentation on which browsers are officially supported but as IE11 used to work I'm guessing this might be a bug. 

https://github.com/cssinjs/jss/blob/105d854d4c0e3ea541f78cbd78cca583f4d83cf9/packages/jss-plugin-default-unit/src/index.js#L43

The above code was introduced in https://github.com/cssinjs/jss/pull/1446.

kof:
You are right, it should have been the global `isNaN` cc @oliviertassinari I wonder why it didn't break mui sooner

oliviertassinari:
Oh right, oops. To be honest, we don't care about IE 11 on Material-UI.

kof:
I wonder if we should drop it in JSS too ... not sure how to make that decision

oliviertassinari:
> not sure how to make that decision

We broke it for weeks and nobody noticed/complained. I would say it's a great signal.

kof:
@GertSallaerts how did you notice, in a product?

kof:
@oliviertassinari it might be that users upgrade only with a delay, then they don't test in IE11 every day, so they figure that out only in a few weeks that it's broken, not sure the delay in signal can be interpreted as "not used"

kof:
JSS core is probably used in a few different type of places than MUI as being lower level abstraction.

GertSallaerts:
Yes, we are updating our stack to React 17 at [ambassify.com](https://www.ambassify.com) and as a result also updated `react-jss` (which we use for our styling) to the latest version which is how I ran into this.

kof:
I am even more surprised because we run tests in IE 9, 10 and 11 and that PR had a test in it

GertSallaerts:
On IE11 specifically I'd suggest you drop it in the next major version and add a list of browsers you do officially support to the documentation, but that's just my two cents :sweat_smile:

kof:
Btw. we do have a list of supported browsers and it was linked in the docs, but now I don't see it being linked any more, we should be more explicit about it https://github.com/cssinjs/jss/blob/master/browsers.json

kof:
I think we should deprecate IE9-11 along with a scoped custom props support https://github.com/cssinjs/jss/issues/1015 because it makes sense for this feature to be in the core and we couldn't do it with IE11 support.

kof:
[...]

PR #1456, opened 2021-02-17
Reuse babel options and specify polyfills for karma explicitely

kof:
Fixes https://github.com/cssinjs/jss/issues/1455

I was looking into why we didn't catch that with tests, it turned out we load all polyfills in karma, but as I was looking I made the babel config reusable, it didn't change anything but still good to have.


babelOptions.js
```diff
@@ -0,0 +1,9 @@
+exports.getBabelOptions = ({useESModules}) => ({
+  presets: [['@babel/env', {loose: true}], '@babel/flow', '@babel/react'],
+  plugins: [
+    ['@babel/proposal-class-properties', {loose: true}],
+    ['@babel/transform-runtime', {useESModules}],
+    ['@babel/plugin-proposal-export-namespace-from'],
```
TrySound, line 6: Is this plugin actually used?

kof: I searched for `export * as` - nothing found

babelOptions.js
```diff
@@ -0,0 +1,9 @@
+exports.getBabelOptions = ({useESModules}) => ({
```
TrySound, line 1: Do you remember why we didn't use babel.config.js?

kof: no, but I would guess the useESModules option is dynamic, that's why?

TrySound: Yes, but it can be used only for rollup. There is no difference between inlined and imported helpers.

tests/utils.js
```diff
@@ -21,7 +21,14 @@ export function removeWhitespace(str) {
 }
 
 export function getRules(style) {
-  return [...style.sheet.cssRules]
+  const rulesArr = []
+  const {cssRules} = style.sheet
+  for (const key in cssRules) {
```
TrySound, line 26: Maybe better `for(;;)`?

babelOptions.js
kof: using the option for rollup only is fine, it's still better than having the entire babel config duplicted

babelOptions.js
kof: I guess we can remove it

babelOptions.js
TrySound: Btw, babel 7.13 does not require useESModules flag anymore. Right version is detected automatically.

kof: then we can remove the flag along with upgrading babel, probably better separately

tests/utils.js
kof: hah, for in failed in safari ... good catch

kof:

PR #1457, opened 2021-02-21
fix: type import from csstype

hosseinmd:
Typescript not working I changed import from `csstype` and now that working
## Todo

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -12,7 +12,7 @@ type Func<R> = ((data: any) => R)
 
 type NormalCssProperties = CSSProperties<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties
-  ? NormalCssProperties[K] | JssValue
+  ? NormalCssProperties[K]
   : JssValue
 
 export type JssStyle =
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -1,4 +1,4 @@
-import * as css from 'csstype'
+import {Properties as CSSProperties} from 'csstype'
 
 // Observable support is included as a plugin.  Including it here allows
 // TypeScript users to use Observables, which could be confusing if a user
@@ -10,7 +10,7 @@ import {Observable} from 'indefinite-observable'
 // TODO: Type data better, currently typed as any for allowing to override it
 type Func<R> = ((data: any) => R)
 
-type NormalCssProperties = css.Properties<string | number>
+type NormalCssProperties = CSSProperties<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties
   ? NormalCssProperties[K] | JssValue
   : JssValue


packages/jss/src/index.d.ts
```diff
@@ -10,9 +10,9 @@ import {Observable} from 'indefinite-observable'
 // TODO: Type data better, currently typed as any for allowing to override it
 type Func<R> = ((data: any) => R)
 
-type NormalCssProperties = css.Properties<string | number>
+type NormalCssProperties = CSSProperties<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties
-  ? NormalCssProperties[K] | JssValue
+  ? NormalCssProperties[K]
```
moshest, line 15: This seems as a breaking change.. why removing it?

hosseinmd: JssValue in there converting NormalCssValues to any.

I think that not needed because when the key is a key of NormalCssProperties we should respond type of that key

moshest: I agree but this can be a breaking change for some users. If that not required I prefer to remove it and maybe push it on a separate PR as breaking change.

hosseinmd: i have used jss in huge typescript project, there is nothing broke.

moshest: Still, if someone use a type that's not defined on the standard properties from some reason it will break. No sure if this consider a break or a fix, but I remember some complains in the past about this. `NormalCssProperties` is not perfect for all properties.

kof: Do I understand correctly that this change would break types if someone uses a property that doesn't exist yet in the NormalCssProperties list?

I think it might be unexpected for people if they use a perfectly valid property, but also I understand that if we don't make a non-existing property to an error, people won't see if they had a typo, right?

moshest: Yep, you got it right.

kof: Lets check out what others do, e.g. styled-components. It's an expectations question as well. I would prefer a strict version and ignore if type is wrong in some cases. This would make sure people contribute to the NormalCssProperties types and make sure everything is included, even if sometimes they will have to ignore the type checking for the missing property temporarily.

kof: If we decide to do that, we would be on a safe to include it in a major version

moshest: I think the same. We can merge this PR without this line though.

hosseinmd: > Do I understand correctly that this change would break types if someone uses a property that doesn't exist yet in the NormalCssProperties list?

No, if a property doesn't exist in NormalCssProperties then it will return JssValue.

hosseinmd: > I think the same. We can merge this PR without this line though.

This PR is because of this line

hosseinmd: This PR is a type fix, no break change.

kof: So the type assumes if the property is not in NormalCssProperties then it must be a custom property specific to jss and have a JssValue? 

Will that not throw if that property is not a JssValue?

kof: Looked at JssValue def, it is so flexible that it will probably not break anything:

```js
export type JssValue =
  | (string & {})
  | (number & {})
  | Array<string | number | Array<string | number> | '!important'>
  | null
  | false
 ```
 
 If we start refining the JssValue and that unknown property is not a JssValue, then it will throw, right?

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: Let's clarify this type, 
We have a NormalCssValues<K>, 
if K is own of the NormalCssProperties
then NormalCssValues = NormalCssProperties[K] 
else NormalCssValues = JssValue

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: throw = eslint error or typescript error

Is this mean of throw?

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: > then it will throw, right?

No

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: Because of JssStyle, but yes JssValue will be throw but that will cach from JssStyle

Comment (kof) [reply packages/jss/src/index.d.ts:15]: what if it's not a Jssvalue but is not in NormalCssProperties? It will probably not fail right now, because JssValue is flexible, but it still is kinda a wrong assumption, I think we lack an UnknownValue something like this

```
f K is own of the NormalCssProperties
then NormalCssValues = NormalCssProperties[K]
else NormalCssValues = JssValue | UnknownValue
```

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: So what is UnknownValue type?

Comment (kof) [reply packages/jss/src/index.d.ts:15]: We don't have it, it's what we should get when property is not in NormalCssProperties and it's not  a JssValue.
wdyt?

Comment (kof) [reply packages/jss/src/index.d.ts:15]: I just checked styled components, they do show an error when an unknown property is used

Screenshot 2021-02-23 at 10 52 16

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: What about override classes?

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]: How we can recognize that this is unknown property or is overridden class

Comment (kof) [reply packages/jss/src/index.d.ts:15]: not sure I understand, can you make an example?

Comment (hosseinmd) [reply packages/jss/src/index.d.ts:15]:
```js
{
  aClass: {
      width: '100%',
      '& .ant-select-selector': {
         border: 'none !important',
         paddingBottom: 8,
    },
  }
}
```

kof: what is `aClass`  there? In JSS this is called a rule name, it's not a class, it's a string that will be part of the generated class in dev mode

kof: and it's not part of NormalCssProperties, `width` is part of NormalCssProperties afaik, but not the parent rule name

hosseinmd:
```js
const useStyles= createUseStyles({
  aClass: {
      width: '100%',
      '& .ant-select-selector': {
         border: 'none !important',
         paddingBottom: 8,
    },
  }
})
```

kof: I might simply not understand what you are trying to say.

hosseinmd: > I just checked styled components, they do show an error when an unknown property is used
> 
> Screenshot 2021-02-23 at 10 52 16

What about '& .ant-select-selector'?

kof: we can probably use for this the new TS feature and identify nesting by "&"

hosseinmd: However, you can merge this PR, and we can continue to discuss in an issue.

kof: @moshest seems like we can merge it since because of flexible JssValue it should be breaking it right now, but please check it out, I have no code base in TS using it to do so.

moshest: I don't have time to test it right now but I'm ok with merging it. It should be fine for most users anyway..

kof:
merged! thank you

Issue #1459, opened 2021-03-01
[docs] Improve Help for People Doing a Lazy Migration for HOCs


ITenthusiasm:
### The Problem
From what I've seen [in the docs](https://cssinjs.org/react-jss/?v=v10.5.1), there doesn't seem to be much _straightforward_ assistance for those upgrading React JSS when it comes to HOCs. Some useful links are posted, but it would be helpful if the user didn't have to learn/read up on all the topics on their own.

### The Solution
Instead of redirecting the user to other articles, the docs could provide an example of creating a custom HOC in JS/TS that implements the new hooks API.

### Some Thoughts/Concerns
1. If a HOC example was provided that used hooks, it may not be easy to provide a [simple] example that's compatible with decorators. (If decorators aren't a focus/concern, then this isn't a problem.)
2. I wasn't sure if an example solution was excluded to avoid being opinionated. Thoughts or clarity on this would be helpful.

---

- [x] **I am willing to implement this**

kof:
If I understand correctly you are looking for an example how to create a HOC using the hook as a goal to keep using the existing components using HOC, right?

I think it's a great idea and we should add this when we decide to remove the HOC. Right now the message is to stop using them and use hooks interface.

You or anyone else are welcome to add such an example to the docs!

ITenthusiasm:
Yes, you are correct!

> You or anyone else are welcome to add such an example to the docs!

I'll try to look into this some time this weekend or next. I think the example shouldn't be too hard to create. Feel free to ping me again if I vanish after the next 2 weeks.

ITenthusiasm:
> I'll try to look into this some time this weekend or next. I think the example shouldn't be too hard to create. Feel free to ping me again if I vanish after the next 2 weeks.

Literally 3 weeks ago. But other things have left me preoccupied, obviously. With #1460 out of the way, and with the confusion of #1469 resolved, I'm going to start working on a draft for this. Hopefully it should be a pretty small change.

Issue #1431, opened 2020-12-03
Type props/data


lucasbordeau:
__Is your feature request related to a problem? Please describe.__

I can't manage to find any way to type the props/data parameter which is currently unknown : 

![image](https://user-images.githubusercontent.com/26528466/100999321-ef1f4780-355c-11eb-8b6f-26ac92edb6a8.png) 

__Describe the solution you'd like__

It would be nice to provide a type generic parameter on createUseStyle to type `data: unknown` (in fact I can see in the code that it is left as a TODO)

__Are you willing to implement it?__

Yes, I'll try to implement it !

ITenthusiasm:
Any updates on this? I'm hesitant to send a PR because I'm unsure what makes this particularly complicated/difficult.

Looking in [createUseStyles.js](https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/createUseStyles.js), I'd imagine another generic type parameter for `Props` could get added to the `createUseStyles` function. Then the `data` parameter for `useStyles` could be given the correct type.

There are only 4 instances of `data` in that file. One is for the parameter definition, another is for a React hooks dependency. `addDynamicRules` and `updateDynamicRules` from [sheets.js](https://github.com/cssinjs/jss/blob/ca8269f730bf9e57fdfa423d036b177e49307cdc/packages/react-jss/src/utils/sheets.js#L108) are the only 2 functions using `data`, and they use `any` for its type. In terms of making sure `useStyles` enforces receiving the correct argument, the change would be straightforward.

I guess making sure the CSS properties based on props are correctly typed would be potentially more complicated, though.


**EDIT**: Looking at this now, it looks like these files are using flow. Would the changes for TypeScript to work out smoothly only be in the `index.d.ts` files?

ITenthusiasm:
To keep track of things for myself (or whoever ends up implementing this), I'll continue to add what I find. If anyone else has anything to add to the discussion or notices something that's wrong, lmk.

Moving forward with the assumption that the flow files will need to get updated:

Moving down the list, `addDynamicRules` and `updateDynamicRules` would need to become generic functions. They would base the generic type on the type of `data`.

`sheet.updateOne` seems to be the only function in that file that relies on `data`. So I guess the rabbit hole only needs to go deeper there.

ITenthusiasm:
[...]

Issue #1273, opened 2020-01-30
Using dynamic styles in React JSS with Typescript


milotoor:
__Expected behavior:__
A dynamic style that uses React props should be able to get the type of the props for use in the style rule.

__Describe the bug:__
Consider the following basic component:

```typescript
type CardProps = {
  padding: number
};

const Card: React.FC<CardProps> = ({ padding }) => {
  return (
    <div style={{ padding }}>
      I am a card.
    </div>
  );
};
```

Using JSS this can be refactored:

```typescript
type CardProps = {
  padding: number
};

const useStyles = createUseStyles({
  card: props => ({
    padding: props.padding
  })
});

const Card: React.FC<CardProps> = () => {
  const classes = useStyles();
  return (
    <div  className={classes.card}>
      I am a card.
    </div>
  );
};
```

However, in strict mode Typescript will complain that `Parameter 'props' implicitly has an 'any' type`, because it's unable to infer the `props` argument's type. This can be fixed by explicitly providing the type:

```typescript
const useStyles = createUseStyles({
  card: (props: CardProps) => ({
    padding: props.padding
  })
});
```

But this won't scale well if there are other dynamic styles:

```typescript
const useStyles = createUseStyles({
  card: (props: CardProps) => ({
    padding: props.padding
  }),
  otherStyleRule: (props: CardProps) => ({
    padding: props.padding * 2
  })
});
```

It would be great if there were a way to inform `createUseStyles` what the `props` type will be. For example:

```typescript
const useStyles = createUseStyles<CardProps>({
  card: props => ({
    padding: props.padding
  }),
  otherStyleRule: props => ({
    padding: props.padding * 2
  })
});
```

Perhaps there is already a way to do this? The [Typescript section](https://cssinjs.org/react-jss/?v=v10.0.4#typescript) of the v10.0.4 docs is empty, so perhaps the resolution to my issue is to fill out that section. I'm happy to write a bit for the docs, though I need to understand how the types function first!

__Versions (please complete the following information):__
 - jss: 10.0.4
 - Browser [e.g. chrome, safari]: n/a
 - OS [e.g. Windows, macOS]: macOS
Feel free to add any additional versions which you may think are relevant to the bug.

maximprokopchuk:
Also it would be great to have `CardProps` from example above as expected type of useStyles function param (`props`):
`const styles = useStyles(props)`
[...]

Issue #1351, opened 2020-06-09
Support CSS types on createUseStyles


moshest:
This feature is related to the rollback made for issue https://github.com/cssinjs/jss/issues/1255

The fix was to rollback to previous version and now every IDEA autocomplete is broken when writing `createUseStyles`.

Rollback change: https://github.com/cssinjs/jss/pull/1270

Any plan to fix that soon?

kof:
Feel free to fix it, I will merge it but then you will have to deal with follow up problems. If you are willing to do that 👍 

Alternatively, I can move out TS types off core back to definitely typed and it can be released independently.

moshest:
Can you elaborate more about those problems?

What's wrong with the implementation previously on #1270?

kof:
I can't, it's just that TS types are not in a good shape and every change seems to result in new issues and I don't have time to maintain it.

moshest:
OK, I will take a look later on.

moshest:
released a PR here: https://github.com/cssinjs/jss/pull/1352

PR #1460, opened 2021-03-06
Improve Types for Hooks

ITenthusiasm:
## What Would You Like to Fix/Add?
This PR aims to ensure that the `createUseStyles` hook has accurate TS intellisense for `props` and `theme`s. The details of the work can be found in the commit message. Note that as a result of cascading effects, this PR also improves the types for the deprecated `withStyles` HOC. This should satisfy #1273 and #1431.

These are TypeScript changes **only**, as only these changes will be visible to the TypeScript users. These are the minimal changes necessary to improve the user experience for TS devs. Flow is separate and unrelated to the user, and I'm not so familiar with it, so it's simplest to add flow changes in a separate commit (if that's acceptable). (If someone else familiar with Flow could apply the changes, that would be helpful. I've already supplied the TS types, so it should be easy.)

## Goals
- The types should be accurate for **both** kinds of arguments supplied to `createUseStyles`/`withStyles` (ie. the object argument and the function argument).
- The type definitions should ban the declaration of [React-JSS-related] types that have already been defined in the surrounding scope. This prevents confusion that could be caused from unnecessary shadowing (which many eslint configs would oppose anyway). It also discourages users from attempting to define nested functions (which JSS seems to warn against already).
  - As an example: A function of `props` and `theme` that returns a style object will not give type inference to nested functions.
- The generic types should be ordered according to what users are most likely to need or use. For instance, `Props` is typically more necessary than `Theme`.
- A `withStyles` "lazy migration" in TS should be "sufficiently easy" ("not too complicated") to write given these changes. If not, these new types may need to be revisited. (This requires closely examining how well the HOC is defined, not just how well these new types are defined.)
  - On that note, I intend to send out a PR for this in the future, pending this PR. See #1459.

## Caveats
- Supplying **any** generic type parameters causes all defaults to be used by TS. This means you risk losing the implicit definitions of class names whenever you specify `props`/`data`.
  - This is a TS problem, and it will hopefully get resolved in the near future. See microsoft/TypeScript#26349.
- No breaking changes for JS. But for TS (mainly React), if anyone is using the old `createUseStyles<Theme>` syntax, they will have to update their types. This is unfortunate, but not hard. And since TS has better intellisense when `Name` is not specified, this shouldn't be a common concern.

## Unknowns / Concerns
- **Resolved**: Use default `any`. ~~In `jss/src/index.d.ts`, should all _default_ usages of `JssStyle` be updated to use `JssStyle<any>` in the context of these proposed changes?~~

## Extra Notes
- `createStyleSheet` uses `any` for the `Data`/`Props` type because the data can be dynamically changed at any point. `any` is safest.

## Todo
- [x] Add tests that verify the modified behavior
- [x] Add documentation if it changes the public API
  - N/A (no TS documentation currently exists...yet)

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -7,35 +7,36 @@ import {Properties as CSSProperties} from 'csstype'
 // TODO: refactor to only include Observable types if plugin is installed.
 import {Observable} from 'indefinite-observable'
 
-// TODO: Type data better, currently typed as any for allowing to override it
-type Func<R> = ((data: any) => R)
+type Func<P, T, R> = T extends undefined ? ((data: P) => R) : ((data: P & {theme: T}) => R)
 
 type NormalCssProperties = CSSProperties<string | number>
-type NormalCssValues<K> = K extends keyof NormalCssProperties
-  ? NormalCssProperties[K]
-  : JssValue
+type NormalCssValues<K> = K extends keyof NormalCssProperties ? NormalCssProperties[K] : JssValue
 
-export type JssStyle =
+export type JssStyle<Props = any, Theme = undefined> =
   | {
       [K in keyof NormalCssProperties]:
         | NormalCssValues<K>
-        | JssStyle
-        | Func<NormalCssValues<K> | JssStyle | undefined>
+        | JssStyle<Props, Theme>
+        | Func<Props, Theme, NormalCssValues<K> | JssStyle<undefined, undefined> | undefined>
         | Observable<NormalCssValues<K> | JssStyle | undefined>
     }
   | {
       [K: string]:
         | JssValue
-        | JssStyle
-        | Func<JssValue | JssStyle | undefined>
+        | JssStyle<Props, Theme>
+        | Func<Props, Theme, JssValue | JssStyle<undefined, undefined> | undefined>
         | Observable<JssValue | JssStyle | undefined>
     }
 
-export type Styles<Name extends string | number | symbol = string> = Record<
+export type Styles<
+  Name extends string | number | symbol = string,
+  Props = unknown,
+  Theme = undefined
+> = Record<
   Name,
-  | JssStyle
+  | JssStyle<Props, Theme>
   | string
-  | Func<JssStyle | string | null | undefined>
+  | Func<Props, Theme, JssStyle<undefined, undefined> | string | null | undefined>
   | Observable<JssStyle | string | null | undefined>
 >
 export type Classes<Name extends string | number | symbol = string> = Record<Name, string>
@@ -215,7 +216,10 @@ export interface StyleSheet<RuleName extends string | number | symbol = string |
    * Create and add rules.
    * Will render also after Style Sheet was rendered the first time.
    */
-  addRules(styles: Partial<Styles<RuleName>>, options?: Partial<RuleOptions>): Rule[]
+  addRules(
+    styles: Partial<Styles<RuleName, any, undefined>>,
+    options?: Partial<RuleOptions>
+  ): Rule[]
   /**
    * Get a rule by name.
    */
@@ -250,7 +254,7 @@ export interface JssOptions {
 
 export interface Jss {
   createStyleSheet<Name extends string | number | symbol>(
-    styles: Partial<Styles<Name>>,
+    styles: Partial<Styles<Name, any, undefined>>,
     options?: StyleSheetFactoryOptions
   ): StyleSheet<Name>
   removeStyleSheet(sheet: StyleSheet): this
--- a/packages/react-jss/src/index.d.ts
+++ b/packages/react-jss/src/index.d.ts
@@ -37,17 +37,21 @@ declare const JssContext: Context<{
   disableStylesGeneration: boolean
 }>
 
-type ClassesForStyles<S extends Styles | ((theme: any) => Styles)> = Classes<
-  S extends (theme: any) => Styles ? keyof ReturnType<S> : keyof S
->
+type ClassesForStyles<
+  S extends Styles<any, any, any> | ((theme: any) => Styles<any, any, undefined>)
+> = Classes<S extends (theme: any) => Styles<any, any, undefined> ? keyof ReturnType<S> : keyof S>
 
-interface WithStylesProps<S extends Styles | ((theme: any) => Styles)> {
+interface WithStylesProps<
+  S extends Styles<any, any, any> | ((theme: any) => Styles<any, any, undefined>)
+> {
   classes: ClassesForStyles<S>
 }
 /**
  * @deprecated Please use `WithStylesProps` instead
  */
-type WithStyles<S extends Styles | ((theme: any) => Styles)> = WithStylesProps<S>
+type WithStyles<
+  S extends Styles<any, any, any> | ((theme: any) => Styles<any, any, undefined>)
+> = WithStylesProps<S>
 
 declare global {
   namespace Jss {
@@ -72,18 +76,17 @@ interface CreateUseStylesOptions<Theme = DefaultTheme> extends BaseOptions<Theme
   name?: string
 }
 
-declare function createUseStyles<Theme = DefaultTheme, C extends string = string>(
-  styles: Styles<C> | ((theme: Theme) => Styles<C>),
+declare function createUseStyles<C extends string = string, Props = unknown, Theme = DefaultTheme>(
+  styles: Styles<C, Props, Theme> | ((theme: Theme) => Styles<C, Props, undefined>),
   options?: CreateUseStylesOptions<Theme>
-): (data?: unknown) => Classes<C>
+): (data?: Props & {theme?: Theme}) => Classes<C>
 
 type GetProps<C> = C extends ComponentType<infer P> ? P : never
 
-declare function withStyles<
-  ClassNames extends string | number | symbol,
-  S extends Styles<ClassNames> | ((theme: any) => Styles<ClassNames>)
->(
-  styles: S,
+declare function withStyles<ClassNames extends string | number | symbol, Props, Theme>(
+  styles:
+    | Styles<ClassNames, Props, Theme>
+    | ((theme: Theme) => Styles<ClassNames, Props, undefined>),
   options?: WithStylesOptions
 ): <C>(
   comp: C
@@ -91,7 +94,7 @@ declare function withStyles<
   JSX.LibraryManagedAttributes<
     C,
     Omit<GetProps<C>, 'classes'> & {
-      classes?: Partial<ClassesForStyles<S>>
+      classes?: Partial<ClassesForStyles<typeof styles>>
       innerRef?: RefObject<any> | ((instance: any) => void)
     }
   >
--- a/packages/react-jss/tests/types/createUseStyles.ts
+++ b/packages/react-jss/tests/types/createUseStyles.ts
@@ -1,0 +1,238 @@
+import {createUseStyles} from '../../src'
+
+type DefaultTheme = Jss.Theme
+
+interface MyProps {
+  property: string
+}
+
+interface MyTheme {
+  color: 'red'
+}
+
+const expectedCustomProps = {property: ''}
+const expectedDefaultTheme: DefaultTheme = {defaultFontSize: 0, themeColour: ''}
+const expectedCustomTheme: MyTheme = {color: 'red'}
+
+/* -------------------- THEME ARGUMENT -------------------- */
+// Regular, static styles work fine
+const themeArg1 = createUseStyles(theme => ({
+  someClassName: '',
+  anotherClassName: {
+    fontWeight: 'bold'
+  }
+}))
+const themeArg1ClassesPass = themeArg1()
+
+// Theme type assumed to be the default
+// Nested theme declaration banned
+// @ts-expect-error
+const themeArg2 = createUseStyles(theme => ({
+  themeNotAllowed: ({theme: innerTheme}) => ({
+    fontWeight: 'bold'
+  })
+}))
+// @ts-expect-error
+const themeArg2ClassesFail = themeArg2({theme: {}})
+// @ts-expect-error
+const themeArg2ClassesFail2 = themeArg2({theme: expectedCustomTheme})
+const themeArg2ClassesPass = themeArg2({theme: expectedDefaultTheme})
+
+// Props declaration is allowed
+const themeArg3 = createUseStyles<string, MyProps>(theme => ({
+  onlyPropsAllowed: ({...props}) => ({
+    fontWeight: 'bold'
+  })
+}))
+// @ts-expect-error
+const themeArg3ClassesFail = themeArg3({property: 0})
+// @ts-expect-error
+const themeArg3ClassesFail2 = themeArg3({...expectedCustomProps, theme: expectedCustomTheme})
+const themeArg3ClassesPass = themeArg3(expectedCustomProps)
+const themeArg3ClassesPass2 = themeArg3({...expectedCustomProps, theme: expectedDefaultTheme})
+
+// Nested props declaration banned
+const themeArg4 = createUseStyles<string, MyProps>(theme => ({
+  onlyPropsAllowed: ({...props}) => ({
+    fontWeight: 'bold',
+    // @ts-expect-error
+    propsNotAllowed: ({...innerProps}) => ''
+  })
+}))
+
+// Supplied theme type is acknowledged
+const themeArg5 = createUseStyles<string, unknown, MyTheme>(theme => ({}))
+// @ts-expect-error
+const themeArg5ClassesFail = themeArg5({theme: {}})
+// @ts-expect-error
+const themeArg5ClassesFail2 = themeArg5({theme: expectedDefaultTheme})
+const themeArg5ClassesPass = themeArg5({theme: expectedCustomTheme})
+
+// Custom theme can be determined from argument
+const themeArg6 = createUseStyles((theme: MyTheme) => ({
+  someClassName: {
+    fontWeight: 'bold'
+  }
+}))
+// @ts-expect-error
+const themeArg6ClassesFail = themeArg6({theme: {}})
+// @ts-expect-error
+const themeArg6ClassesFail2 = themeArg6({theme: expectedDefaultTheme})
+const themeArg6ClassesPass = themeArg6({theme: expectedCustomTheme})
+
+// Props can be determined implicitly
+const themeArg7 = createUseStyles(theme => ({
+  checkbox: ({property}: MyProps) => ({
+    borderColor: property
+  })
+}))
+
+// @ts-expect-error invalid props
+const themeArg7ClassesFail = themeArg7({colour: 'green'})
+// @ts-expect-error extraneous props
+const themeArg7ClassesFail2 = themeArg7({...expectedCustomProps, someUnTypedProp: 1})
+const themeArg7ClassesPass = themeArg7(expectedCustomProps)
+
+// Classes check
+const themeArgClasses7String: string = themeArg7ClassesPass.checkbox
+// @ts-expect-error invalid className
+themeArg7ClassesPass.doesntExist
+
+/* -------------------- NO THEME ARGUMENT -------------------- */
+// Regular, static styles work fine
+const noThemeArg1 = createUseStyles({
+  someClassName: '',
+  anotherClassName: {
+    fontWeight: 'bold'
+  }
+})
+const noThemeArg1ClassesPass = noThemeArg1()
+
+// Theme declaration is allowed, but not nested theme declaration
+// Theme type assumed to be the default
+const noThemeArg2 = createUseStyles({
+  themeOnly: ({theme}) => ({
+    fontWeight: 'bold',
+    // @ts-expect-error
+    themeNotAllowed: ({theme: innerTheme}) => '',
+    '& > *': {
+      color: 'red',
+      // @ts-expect-error
+      themeNotAllowed: ({theme: innerMostTheme}) => ''
+    }
+  })
+})
+// @ts-expect-error
+const noThemeArg2ClassesFail = noThemeArg2({theme: {}})
+// @ts-expect-error
+const noThemeArg2ClassesFail2 = noThemeArg2({theme: expectedCustomTheme})
+const noThemeArg2ClassesPass = noThemeArg2({theme: expectedDefaultTheme})
+
+// Props declaration is allowed, but not nested props declaration
+const noThemeArg3 = createUseStyles<string, MyProps>({
+  propsAndTheme: ({property, theme}) => ({
+    fontWeight: 'bold',
+    // @ts-expect-error
+    nothingAllowed: ({property: innerProperty}) => '',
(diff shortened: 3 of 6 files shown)


ITenthusiasm:
Just saw the CI fail due to some peculiar browser issue. But I didn't touch any JavaScript. :thinking: Is this something to be worried about?

ITenthusiasm:
From [Ryan's Comment](https://github.com/microsoft/TypeScript/issues/39008#issuecomment-642074983) on microsoft/TypeScript#39008, it unfortunately seems like `<Name, Props, Theme>`, may actually be a more practical ordering of the type parameters, even though it isn't necessarily in order of "relevance". :pensive:

Such a change would also remove one of the caveats I mentioned while getting things more ready for what microsoft/TypeScript#26349 would _hopefully_ bring to the table soon.

Thoughts? I could apply the change if desired.

packages/jss/src/index.d.ts
```diff
@@ -7,35 +7,38 @@ import {Properties as CSSProperties} from 'csstype'
 // TODO: refactor to only include Observable types if plugin is installed.
 import {Observable} from 'indefinite-observable'
 
-// TODO: Type data better, currently typed as any for allowing to override it
-type Func<R> = ((data: any) => R)
+type Func<P, T, R> = T extends undefined ? ((data: P) => R) : ((data: P & {theme: T}) => R)
```
moshest, line 10: maybe ThemeFunc is a better name then.

packages/jss/src/index.d.ts
```diff
@@ -7,35 +7,38 @@ import {Properties as CSSProperties} from 'csstype'
...
 
-type NormalCssProperties = CSSProperties<string | number>
+type NormalCssProperties = css.Properties<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties
-  ? NormalCssProperties[K]
+  ? NormalCssProperties[K] | JssValue
   : JssValue
 
-export type JssStyle =
+export type JssStyle<Props = unknown, Theme = undefined> =
```
moshest, line 17: Looks like you use this a lot with `any` i think the default should be `any` not `unknown`. What's the behavior if it's `unknown`? Will it break types somewhere?

moshest (approved):
Overall I liked that a lot. Great job!

ITenthusiasm:
I'm a little preoccupied right now. But just in case, really quickly, I wanted to call attention to:

- In `jss/src/index.d.ts`, should all default usages of `JssStyle` be updated to use `JssStyle<any>` in the context of these proposed changes?
- `Styles<Props, Theme, Name>` ordering vs. `Styles<Name, Props, Theme>` ordering.

Some thoughts on those would be helpful before merging (if everything gets approved). I know moshest commented on the default type for `JssStyle`. If the default is changed accordingly, then the first item in the list would automatically be resolved.

packages/jss/src/index.d.ts
ITenthusiasm: Yeah the name here is a hard one. The `theme` type isn't always guaranteed to be provided, so I don't know if `ThemeFunc` would work. If a prefix would add clarity, maybe something like `DataFunc` or `StylesFunc`? Just some ideas. Other options could exist too.

packages/jss/src/index.d.ts
ITenthusiasm: I don't think the types would break. `unknown` is considered the "type-safe" version of `any` because it forces the end user to specify types. And using `any` is typically discouraged. So my brain has defaulted to using `unknown` in place of `any` in consequence.

From the user's perspective, `unknown` will force the user to define types explicitly. So if they want to destructure or use `props`, they'll need to provide the type as a generic (which is a good thing). Additionally, many TS linters will complain if you're using something that implicitly has the `any` type, so in some cases `unknown` may as well be used.

However, looking at `JssStyle`, I think I may agree that it would actually be better to make `Props` default to `any`. As far as hooks/HOCs are concerned, users will only be interfacing with `Styles`. So `Styles` is probably the only type that needs to default to `unknown`.

kof: I guess those who didn't specify the theme type will not be able to upgrade because this will require them. How do other TS projects handle such changes, as breaking change and major version increment?

ITenthusiasm:
Rebased onto master.

packages/jss/src/index.d.ts
ITenthusiasm: @moshest Changed the default to `any`. Mind giving the file a second glance to make sure things still look normal?

packages/jss/src/index.d.ts
```diff
@@ -50,8 +53,8 @@ export type CreateGenerateId = (options?: CreateGenerateIdOptions) => GenerateId
 export type GenerateId = (rule: Rule, sheet?: StyleSheet<string>) => string
 
 export type JssValue =
-  | (string & {})
-  | (number & {})
+  | string
+  | number
```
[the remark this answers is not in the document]
ITenthusiasm: Yeah. Before I started working with a forked repo, I was editing things from a `node_modules` folder. I guess it had the older code when I copied things over. Thanks for pointing that out. I'll change it.

packages/react-jss/src/index.d.ts
```diff
@@ -72,26 +76,25 @@ interface CreateUseStylesOptions<Theme = DefaultTheme> extends BaseOptions<Theme
   name?: string
 }
 
-declare function createUseStyles<Theme = DefaultTheme, C extends string = string>(
-  styles: Styles<C> | ((theme: Theme) => Styles<C>),
+declare function createUseStyles<Props = unknown, Theme = DefaultTheme, C extends string = string>(
```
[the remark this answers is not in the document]
ITenthusiasm: Yeah, this one is a bit more difficult and is why I asked the question about ordering the generic type parameters.

Having `props` as the third argument requires people to "reach over" `Theme` and explicitly define `ClassNames` in order to provide any kind of prop info. I'm not sure how intuitive it is.

If people use theme and class names significantly more than props, then putting props last would make more sense. (This is my first time really getting into the TS side of React JSS, so I don't know what's commonly used and what isn't.) I'm trying to think of how intuitive things will be moving forward too. If it's something unintuitive, I imagine it'll be stuck there for quite some time.

packages/react-jss/tests/types/createUseStyles.ts
```diff
@@ -0,0 +1,202 @@
...
+      innerSingleValue: ({property, theme}) => '',
+      secondNest: ({property, theme}) => ({
+        backgroundColor: 'blue',
+        // @ts-expect-error
+        nothingAllowed: ({theme: innerTheme, ...innerProps}) => ''
+      })
+    }
+  }
+})
+
```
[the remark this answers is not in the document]
ITenthusiasm: Yeah that's a good idea. I'll add it to both sections for good measure.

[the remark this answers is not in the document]
ITenthusiasm: Updated this in my recent commit (`Requested Changes Part 1`). Please unresolve if the change is unsatisfactory.

packages/react-jss/src/index.d.ts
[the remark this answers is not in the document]
k-yle: Agreed, this is a tricky situation until microsoft/TypeScript#26349 is merged. There are different approaches to using JSS - we use theme very often but rarely use props so I think my opinion here is biased. 

My suggestion would be to make `props` the third argument so that this doesn't break anything, but I think that's for @kof to decide.

ITenthusiasm:
Just for the record, I intend to clean up the commit history with a local `rebase` once everything is finalized. So once all the conversations are resolved -- if the PR comes out to be acceptable -- please don't merge the PR until I've reduced it to 1 single commit.

packages/jss/src/index.d.ts
moshest: seems fine now!

packages/react-jss/src/index.d.ts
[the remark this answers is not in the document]
ITenthusiasm: Just so it's visible when @kof responds:

Looking to the future, I think I'd prefer `<Name, Props, Theme>` (or `<Name, Theme, Props>`, depending on which one users employ more frequently). For users, I think that would be significantly more intuitive. Note that if possible, I'd prefer for `Styles`, `withStyles`, and `createUseStyles` to have the same type parameter order. This would improve consistency and minimize confusion.

The overarching question(s) is/are: What is the most intuitive ordering of the generic parameters that minimizes friction? Will many complain about an unintuitive ordering of parameters in the future, especially after `Props` finally gets type support? How difficult will introducing something more intuitive get in the future? Is it worth a breaking change?

In my (limited) opinion, I feel that long-term, something more intuitive is worth the change. This is riding on the assumption that the breaking change will not be too difficult for users to fix, and that it will be very difficult (if not impossible) to introduce a more intuitive structure in the future if we neglect it now. This also rides on the reality that after microsoft/TypeScript#26349 is merged, we'll be virtually unable to take practical advantage of it if `Names` is not the first type parameter. (Reaching over everything to put a `_` on the last parameter will make things ugly for the user.)

In the mean time, for hooks users who _only_ use theme, people can get `Theme` typing implicitly by using the function version of `createUseStyles`, especially with the help of #1453. So they wouldn't have to "reach over" `Name` or `Props`. (Note also that `Props` lacks a "non-hackish" equivalent if it is placed last.) Honestly, since for current users `Name` loses implicit typing when `Theme` is supplied, I'd wager that most people would prefer to move to the implicit types anyway. (That's an _assumption_ though. And I don't know how big migrating pains would be.)

Migration docs can always be added as well. Whatever the decision, I'm happy to change the ordering of the generic type params again once a conclusion is reached.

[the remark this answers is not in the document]
kof: So the question is if people are gonna have styles with props only vs props and theme vs. theme only more often? I think I would put theme last, because conditional styles is probably a more popular case than theme based once.

I think people should really put theming into the ui library components which would be reused everywhere, but application level styles should rarely need to access theme.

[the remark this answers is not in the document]
k-yle: @ITenthusiasm - LGTM, provided @kof is happy with this being a breaking change. 

I think we'll probably need some sort of upgrade guide for typescript users to explain what all the new errors mean (due to the change in argument order)

k-yle (approved):

packages/react-jss/src/index.d.ts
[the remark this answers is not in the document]
ITenthusiasm: Agreed! I was thinking of creating some kind of RFC issue related to the docs for that after this got merged.

ITenthusiasm:
Since everything got approved sooner than I expected, I reduced my commits today (instead of waiting until tomorrow).

These are not breaking changes for JS. However, they are breaking changes as far as TS is concerned due to the new type definitions (particularly for `createUseStyles`).

The PR is approved and the history is reduced to a single commit. @kof, if these changes are acceptable, then this PR is ready to merge. Otherwise, if there are any changes you or anyone else would like me to add, I'd be happy to include them.

kof:
Is it acceptable to release this under minor version increase?

ITenthusiasm:
> Is it acceptable to release this under minor version increase?

@kof, The short answer is that I'm not sure. But in my _opinion_, a minor version increase is probably fine. I mainly feel this way since the JS API hasn't changed, and this is an improvement on the TS types, where features were previously missing.

> How do other TS projects handle such changes, as breaking change and major version increment?

I tried to do some research on this but didn't find any "objectively right" answers. I saw #1255, where some users were upset about breaking type changes. So...some people would likely get upset with a minor increase. But I also saw your link to [Ryan's comment](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25677#issuecomment-388215959) on breaking type changes at the end of that issue. Ryan has additional comments that may be related to the matter [here](https://github.com/microsoft/TypeScript/issues/14116#issuecomment-280592571) and [here](https://github.com/microsoft/TypeScript/issues/14116#issuecomment-292581018). Additionally, other libraries like CRA seem to release minor versions with breaking changes if the changes aren't expected to be too significant for most users.

I also noticed your comments on #1351. If the types truly haven't stabilized yet, then it may be practical to forego bumping the major version since complaints/issues related to types would be found often. Technically speaking, if the types were back on DT, users would encounter breaking changes without bumping major versions anyway. (Btw, I think the types are better housed _here_ than on DT.)

If there aren't any large incoming JS changes, then bumping the major version only for this TS change (especially if the definitions haven't _fully_ stabilized yet) feels a bit weird. I'm sure there'll be people who disagree. But after observing a few discussions -- and just thinking about it practically -- I find an overly religious following of semver unhelpful. (I do like semver though. And generally speaking, it's practical and desirable.)

Personally, I support going with a minor release. If this does happen to be released as a minor change, I'll take the heat of the complaints. The decision is up to you though. (Others are free to chime in.)

Warnings about breaking changes in a changelog, joined with a useful migration guide, would probably facilitate a minor release. It would hopefully reduce complaints too.

packages/jss/src/index.d.ts
```diff
@@ -7,35 +7,38 @@ import {Properties as CSSProperties} from 'csstype'
...
-        | JssStyle
-        | Func<JssValue | JssStyle | undefined>
+        | JssStyle<Props, Theme>
+        | Func<Props, Theme, JssValue | JssStyle<undefined, undefined> | undefined>
         | Observable<JssValue | JssStyle | undefined>
     }
 
-export type Styles<Name extends string | number | symbol = string> = Record<
+export type Styles<
+  Name extends string | number | symbol = string,
```
hosseinmd, line 34: Is Name necessary?

hosseinmd: I think this should not be generic, we could add it to Record directly.

ITenthusiasm: Hey @hosseinmd!

Yes, `Name` is necessary. Without it, users would be unable to guarantee intellisense on class names.

Additionally, `Styles` must remain a generic type in order for users to be able to specify `Props` and `Theme`.

ITenthusiasm:
If anyone's interested in an opinion besides my own on the versioning, I saw this comment from a moderator in the Testing Library Discord.

> I'd say it's a `patch` if the typing before was wrong. If you narrow the parameter types along the way, it should be `major`.

I couldn't find this person's GitHub, so I don't know who they were. But it's another thought!

If strictly, religiously following semver is the goal of this project, then a major version would likely make more sense.

But again, given the volatility of the types, that may cause unnecessarily frequent major bumps (unless the types stabilize some day).

packages/jss/src/index.d.ts
ITenthusiasm: I think I misunderstood part of your question. I'll investigate when I get some time.

hosseinmd: I think you are right

ITenthusiasm:
No rush on the PR, but I'm curious to know what the intentions with this PR will be? My team would benefit from being able to employ the new type changes (as would others). And I don't feel comfortable starting on the work for the lazy migration HOC docs (TypeScript) until I know this is finalized.

I'm just trying to get an idea of the timeline.

kof:
@ITenthusiasm I am finishing https://github.com/cssinjs/jss/pull/1456 and then gonna merge this one and make a release with all recent changes as a minor

packages/jss/src/index.d.ts
```diff
@@ -7,35 +7,38 @@ import {Properties as CSSProperties} from 'csstype'
...
 
-// TODO: Type data better, currently typed as any for allowing to override it
-type Func<R> = ((data: any) => R)
+type Func<P, T, R> = T extends undefined ? ((data: P) => R) : ((data: P & {theme: T}) => R)
 
-type NormalCssProperties = CSSProperties<string | number>
+type NormalCssProperties = css.Properties<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties
-  ? NormalCssProperties[K]
+  ? NormalCssProperties[K] | JssValue
```
hosseinmd, line 14: This is a bug i resolved this in few PR ago

hosseinmd: https://github.com/cssinjs/jss/pull/1457

ITenthusiasm: @hosseinmd I appreciate your scrutiny. I mentioned in [another comment](https://github.com/cssinjs/jss/pull/1460#discussion_r590701242) that some previous changes got lost after I moved my `node_modules` changes to the forked repository. It seems this was only a problem for `packages/jss/src/index.d.ts`. I re-observed the changes and tried to make sure that only the changes related to _this PR_ were kept. Could you re-examine the diff again and tell me if everything looks good?

ITenthusiasm: Note that I cannot stop `prettier` from doing its formatting work, as in the case of `line 13` in the added changes in that file.

ITenthusiasm:
There was a line or 2 where some previous changes were accidentally reverted. My recent `amend` to the commit fixed those.

packages/jss/src/index.d.ts
ITenthusiasm: Just to make it easier to keep track of everything, I'm gonna resolve this for now. (If that's dissatisfactory, feel free to unresolve this conversation.)

Renaming `Func` isn't necessarily related to this PR and would create some noise for anyone reviewing the changes in the future. So I'll leave the naming improvement to the future.

packages/jss/src/index.d.ts
hosseinmd: Looks good to me, thank

kof:
Can you please add to changelog whatever we have done here? eventually with hints what needs to be done in case of errors

ITenthusiasm:
I can do that. Would you want me to update the changelog in this PR or in a separate one?

kof:
Both works, if you want to create a separate - feel free

ITenthusiasm:
Okay cool. I'll probably create a separate PR in that case. I'll try to add changelog updates today or tomorrow.

kof:
Merged  🎉 Thank you everyone!

hosseinmd:
I suggest to use commit lint to generate change Log automatically

kof:
Usually it leads to poor changelogs but we can discuss this in a separate issue, there are projects which do that, we can look at them.

It will require a lot of scrutiny in commit messages, where I am just doing squash-merge of contributed PRs and most of the time just write the changelogs, because it would be a mess otherwise. Most of the time it's hard enough to get people to do good code changes, only exceptional contributors write good commit messages or changelogs

hosseinmd:
i created an issue https://github.com/cssinjs/jss/issues/1465

kof:

hosseinmd:
Order of generic should be changed.

Current is
```ts
createUseStyles<C extends string = string, Props = unknown, Theme = Jss.Theme>
```

should change to 
```ts
createUseStyles<Props = unknown, Theme = Jss.Theme,C extends string = string>
```

ITenthusiasm:
Hi @hosseinmd. Originally, this PR had the order you suggested, but it was changed to the current ordering because it plays most nicely with TypeScript.

Due to [the nature of TypeScript's type inference](https://github.com/microsoft/TypeScript/issues/43119), putting `C` last actually causes users to _lose_ type info. More specifically, if _any_ generic type parameters are specified, then TypeScript immediately assumes the defaults for the remaining, unspecified type parameters. So for instance, under your suggested order, this

```typescript
const useStyles = createUseStyles<MyProps>({ /* ... */ });
const classes = useStyles(props)
```

would cause `classes` to be typed as `Record<string, string>` instead of `Record<RULE_NAMES, string>`. With this knowledge, there are 3 main problems:

1. Under `createUseStyles<Props, Theme, C>`, anyone who _only_ wanted to use `Props` will now be _forced_ to supply a type for `Theme` so that they can finally supply a type for `C` (to get a proper `classes` type). This would result in a lot of code looking something like `createUseStyles<Props, never, C>`, which is unappealing and inefficient. Unfortunately, users will _always_ have to specify `C` if they want autocomplete on `classes`, so `C` actually takes the highest priority.
2. Under `createUseStyles<Props, Theme, C>`, we would not be able to take maximum advantage of microsoft/TypeScript#26349 if/when it gets merged. With `createUseStyles<C, Props, Theme>`, users would be able to do `createUseStyles<_, MyProps>` to maintain an accurate `classes` type. It's not flawless, but it's the best arrangement available that would reduce rule name duplication.
3. Under `createUseStyles<Props, Theme, C>`, we would very likely get more issues reporting "bugs" in React-JSS that are not React-JSS bugs at all but are in fact misapplications of TypeScript. The TypeScript limitation I described to you earlier is not readily apparent to the average TS user, nor does it seem to be a frequently encountered problem. Consequently, when users see that `classes` is properly typed when `Props` isn't specified, and that `classes` _loses_ types when `Props` is specified, many will (understandably) conclude that the issue is in React-JSS. A similar problem occurred in #1331, where a TS issue was mistaken for a React-JSS issue. Of course, we should design around user experience, not simply minimizing "bug" reports. But since we get both with `<C, Props, Theme>`, this is a more valuable order.

hosseinmd:
I don't agree with you at all. C generically defined from inserted object to createUseStyle, and usually we don't need to specify C

ITenthusiasm:
Friend, you're free to disagree. But if you closely read the PR, the comments, and the issues I mentioned in my last comment, you would understand that TypeScript will not work correctly if C isn't specified. Your response tells me that you did not closely examine the examples I gave nor the issues I linked. I can't help you beyond the linked issues and examples I provided (additional examples exist in the linked issues).

Trust me, I ran through tons of headaches trying to figure this thing out. The current order is due to a limitation in TypeScript.

hosseinmd:
> TypeScript will not work correctly if C isn't specified

This is not true.

hosseinmd:
i have an issue with your type definition my application typescript is broken, and I have to define whole C and props, but we don't need to define C because that could be recognized from inserted object.

hosseinmd:
I have to back old version

ITenthusiasm:
@hosseinmd I'm unwilling to continue the conversation until you observe the issues I linked and the evidence I provided.

I'm the one who created the PR, which -- as I mentioned -- started off as `createUseStyles<Props, Theme, C>`. Consequently, I'm also the one who _saw_ `classes` lose type inference when running the extensive tests that I wrote. Though `C` does not lose inference in _all_ situations, it does lose inference once any generic type parameters are provided. This is what led me to create one of the TS issues I linked (which is now closed due to the `_` sigil PR they have open). This is also what led me to place `C` at the front before merging.

> My application typescript is broken

Because `Props` was not a previously provided type parameter, a break would have occurred anyway under your suggestion. So this comment confuses me slightly. If you were never supplying `Props` (or `Theme`), your application shouldn't be broken right now. If you're concerned because you wanted an easy way to supply props, you can see #1467. It provides examples for implicit and explicit typing. If you're concerned because of issues with `Theme`, see the [migration suggestions in the changelog](https://github.com/cssinjs/jss/blob/master/changelog.md#breaking-changes). If you'd be helped by defining a global `Theme` type, see #1453.

hosseinmd:
I created a PR please review that work well

MaximeCheramy:
I've updated react-jss and I'm now facing this issue: https://github.com/cssinjs/jss/issues/1479

PR #1461, opened 2021-03-06
Improve Contributing Docs (Flow)

ITenthusiasm:
This is a PR to improve the contribution docs by explaining how to setup VS Code with Flow support. (It also adds `.vscode` to the `.gitignore` so that different users can maintain their own settings).

I burned an unnecessary amount of time darting around trying to figure out how to get Flow support. Theoretically, users could just do their own research. But if the docs add a quick comment about how to get started, it will help contributors get to work faster (and reduce the number of people who give up because the mere setup seems too difficult).

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -40,3 +40,17 @@ Using karma (real browsers will get launched).
 ```bash
 yarn test
 ```
+
+## IDEs
+
+If you are using VS Code, you'll need to do 2 things to get Flow type support:
+
+1. Install the [Flow Language Support extension](https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode).
+2. If necessary, update your _local_ settings to disable JS file validation:
+
+```JSON
+// <PROJECT_ROOT>/.vscode/settings.json
+{
+  "javascript.validate.enable": false
+}
+```
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage
 lerna-debug.log
 yarn-error.log
 .idea
+.vscode


ITenthusiasm:
Since there are no actual code changes here, I'm pretty sure the CI isn't me here either. :thinking:

contributing.md
```diff
@@ -40,3 +40,17 @@ Using karma (real browsers will get launched).
 ```bash
 yarn test
 ```
+
+## IDEs
```
kof, line 44: This is about adding general flowtype to vscode ... how is this related to JSS project?

I think we should add documentation that specifically helps with JSS related problems

kof: One doesn't even need ide integration to contribute to JSS, one can run flow over console ...

kof:
> I burned an unnecessary amount of time darting around trying to figure out how to get Flow support. 

What specific problem did you have? Why wasn't it possible to just run the checker in the console and be done?

kof:
but anyway, gonna merge it in the hope I won't have to support more cases like this

ITenthusiasm:
Hey! Running late here since the PR already got merged.

Perhaps I could've specified more clearly, but this wasn't related to the library as much as it was related to helping contributors. Flow runs just fine in the console. But in the IDE, the errors were blaring up my eyes, and sometimes the types didn't appear rightly (as a result of TS complaining in VS Code in flow JS files).

People could theoretically just use the console. But having the help of the IDE to speed up type checking/verification helps a lot. The time I burned was on getting the IDE to behave correctly so I could observe the flow files just fine. The rest of the IDE support (ie. for TS) comes automatically because there's already a TS config and built-in support.

If you decide to remove the commit later I understand. I was a little unsure myself. But I figured I'd put it out there just in case because I didn't want anyone else to give up because of IDE issues. I know a lot of frontend devs use VS Code.

kof:
Yeah, I got it, thought initially as well this shouldn't be here but then - whatever helps if it doesn't hurt.

PR #1464, opened 2021-03-11
Type: Resolve plugins additional properties

hosseinmd:

In this PR I added a generic type for Styles in jss for additional properties.
Export compose and extend packages added properties, merged them in default plugin package

## Todo

### DIFF
--- a/packages/jss-plugin-compose/src/index.d.ts
+++ b/packages/jss-plugin-compose/src/index.d.ts
@@ -1,3 +1,7 @@
 import {Plugin} from 'jss'
 
+export type ComposeProperties = {
+  composes: (string | string[])[]
+}
+
 export default function jssPluginSyntaxCompose(): Plugin
--- a/packages/jss-plugin-extend/src/index.d.ts
+++ b/packages/jss-plugin-extend/src/index.d.ts
@@ -1,3 +1,7 @@
 import {Plugin} from 'jss'
 
+export type extendProperties = {
+  extends: string[]
+}
+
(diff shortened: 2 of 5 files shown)


packages/jss-plugin-compose/src/index.d.ts
```diff
@@ -1,3 +1,7 @@
 import {Plugin} from 'jss'
 
+export type ComposeProperties = {
+  composes: (string | string[])[]
```
kof, line 4: - possible values are a string or an array of strings 
- ideally if a string starts with "$", it should match with top level rule names (afaik it's possible now with TS)

packages/jss-plugin-extend/src/index.d.ts
```diff
@@ -1,3 +1,7 @@
 import {Plugin} from 'jss'
 
+export type ExtendProperties = {
+  extends: string[]
```
kof, line 4: - it's "extend"
- it's much more complex than that, please check the docs https://cssinjs.org/jss-plugin-extend?v=v10.5.1

kof:
- [ ] we should def. have ts tests for those things

packages/jss-plugin-compose/src/index.d.ts
hosseinmd: Template Literal Types is a good idea, but it is not working in this case because of string type.

For example:
```
`$${ClassName}' | string
```

This return always string.

kof:
we got merge conflicts

kof:
Can someone from @cssinjs/typescript team please give it a look?

PR #1467, opened 2021-03-14
Add Documentation for Using Hooks with TypeScript

ITenthusiasm:
This is a PR to add documentation for using hooks with TypeScript. I didn't expect 10.6.0 to get released this fast. But since it's out there, I think getting these docs out as soon as possible is optimal. It will help people who've been waiting for better TS support, and it will also give better ideas on how to migrate older code.

Tests are added under a `docs.tsx` file. If the examples are simple, then any TS code that gets added to the documentation should probably have something like this. Implicit types are complicated, so it's important to know they work as expected before suggesting the change to other users.

### DIFF
--- a/docs/react-jss.md
+++ b/docs/react-jss.md
@@ -30,7 +30,7 @@ Try it out in the [playground](https://codesandbox.io/s/j3l06yyqpw).
 - [Custom setup](#custom-setup)
 - [Multi-tree setup](#multi-tree-setup)
 - [Injection order](#injection-order)
-- [TypeScript](#typescript)
+- [Usage with TypeScript](#usage-with-typescript)
 
 ## Install
 
@@ -585,6 +585,6 @@ const Button = () => {
 }
 ```
 
-## TypeScript
+## Usage with TypeScript
 
-TODO hooks support
+For help using TypeScript with React-JSS, go [here](react-jss-ts.md).
--- a/docs/react-jss-ts.md
+++ b/docs/react-jss-ts.md
@@ -1,0 +1,120 @@
+# Using TypeScript with React-JSS
+
+React-JSS enables you to supply types for your `ruleNames`, `props`/`data`, and `theme` both explicitly and implicitly.
+
+## Explicit Typing
+
+```tsx
+// Define Component
+type RuleNames = 'myButton' | 'myLabel'
+
+interface ButtonProps {
+  children?: React.ReactNode
+  spacing?: number
+  fontWeight?: string
+  labelColor?: string
+}
+
+interface CustomTheme {
+  background: string
+}
+
+const useStyles = createUseStyles<RuleNames, ButtonProps, CustomTheme>({
+  myButton: {
+    padding: ({spacing}) => spacing || 10
+  },
+  myLabel: ({theme, ...props}) => ({
+    display: 'block',
+    color: props.labelColor || 'red',
+    fontWeight: props.fontWeight || 'bold',
+    backgroundColor: theme.background || 'gray'
+  })
+})
+
+function Button({children, ...props}: ButtonProps): React.ReactElement {
+  const theme = useTheme<CustomTheme>()
+  const classes = useStyles({...props, theme})
+
+  return (
+    <button className={classes.myButton}>
+      <span className={classes.myLabel}>{children}</span>
+    </button>
+  )
+}
+
+// Create App
+const theme = {background: 'black'}
+
+function App(): React.ReactElement {
+  return (
+    <ThemeProvider theme={theme}>
+      <Button fontWeight="bold">Submit</Button>
+    </ThemeProvider>
+  )
+}
+```
+
+You only need to supply as much type information as you need. For instance, if you aren't using a `Theme`, then you don't need to supply a type for it. You'll notice that if you want to supply types for `Props` or `Theme`, you'll have to supply the type for `RuleNames` first. This means that if you want autocomplete for the `classes` that come from `useStyles`, you'll need to explicitly provide the strings you'll be using. This restriction is due to a limitation in TypeScript [that will hopefully get resolved soon](https://github.com/microsoft/TypeScript/pull/26349). However, you can see the next section for a workaround.
+
+## Implicit Typing
+
+If you want to provide `Props` and/or `Theme` types without having to supply the `RuleNames` type, you can make TypeScript infer the types implicitly. The primary benefit of doing this is that you'll get autocomplete on `classes` _in addition to_ your props/theme without having to duplicate your rule names. There are two approaches:
+
+Using the functional variant...
+
+```tsx
+const useStyles = createUseStyles((theme: CustomTheme) => ({
+  myButton: {
+    padding: (props: ButtonProps) => props.spacing || 10
+  },
+  myLabel: props => ({
+    display: 'block',
+    color: props.labelColor || 'red',
+    fontWeight: props.fontWeight || 'bold',
+    backgroundColor: theme.background || 'gray'
+  })
+}))
+```
+
+Or using the object variant...
+
+```tsx
+const useStyles = createUseStyles({
+  myLabel: ({theme, ...props}: ButtonProps & {theme: CustomTheme}) => ({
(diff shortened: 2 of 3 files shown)


ITenthusiasm:
Although this is urgent, please do closely observe the documentation. We should make sure the docs are at least _decent_ before publishing any changes.

ITenthusiasm:
Added some amend commits. The first one was to make sure the code examples in the `docs.tsx` test were properly updated in the documentation. The second commit simplified some phrasing.

ITenthusiasm:
After this gets merged, it should be simple enough to tackle #1459 in a separate PR

kof:
Maybe it makes sense to have a page for usage with typescript, not just for hooks api, but everything related to ts and react-jss

ITenthusiasm:
> Maybe it makes sense to have a page for usage with typescript, not just for hooks api, but everything related to ts and react-jss and move it to a separate md file

That could definitely be useful. I'd be happy to help with everything related to the TS + ReactJSS portion. But I'm unfamiliar with the other packages, so I wouldn't be able to say much more there.

With respect to React-JSS, since HOCs are deprecated, it would be a little strange to have TS examples with them. The only TS example I'll have in that area is the TS example of creating a lazy migration HOC. (Technically speaking, it may be worth considering exporting that lazy migration HOC.) I don't know if there's much else TS to show for React-JSS.

ITenthusiasm:
Any thoughts about the content in the PR so far? Should it be changed at all?

kof:
> With respect to React-JSS, since HOCs are deprecated, it would be a little strange to have TS examples with them.

Totally, I would only document APIs we want people to use. I thought about ThemeProvider, JssProvider, global Theme and probably some other things, I wold just go over the main document for react-jss and see what needs TS docs/examples

And I would move this to a react-jss-ts.md file and link it from multiple places where people would look for them as well as we need to add it to navigation structure on the documentation site

kof:
Once ready mabe ping `@cssinjs/typescript` for a review

ITenthusiasm:
@cssinjs/typescript Any thoughts on the additions to the TypeScript section?

ITenthusiasm:
@k-yle, it doesn't look like you're in `@cssinjs/typescript`, but you gave some really helpful feedback on #1460. Would appreciate your comments here as well.

ITenthusiasm:
Also I missed this comment:

> Totally, I would only document APIs we want people to use. I thought about ThemeProvider, JssProvider, global Theme and probably some other things, I wold just go over the main document for react-jss and see what needs TS docs/examples

And I would move this to a react-jss-ts.md file and link it from multiple places where people would look for them as well as we need to add it to navigation structure on the documentation site

I'll look into this when I have the time.

docs/react-jss.md
```diff
@@ -587,4 +587,101 @@ const Button = () => {
...
+    </ThemeProvider>
+  )
+}
+```
+
+You only need to supply as much type information as you need. For instance, if you aren't using a `Theme`, then you don't need to supply a type for it. You'll notice that if you want to supply types for `Props` or `Theme`, you'll have to supply the type for `RuleNames` first. This means that if you want autocomplete for the `classes` that come from `useStyles`, you'll need to explicitly provide the strings you'll be using. This restriction is due to a limitation in TypeScript [that will hopefully get resolved soon](https://github.com/microsoft/TypeScript/pull/26349).
+
+### Implicit Typing
+
+If you want to provide `Props` and/or `Theme` types without having to supply the `RuleNames` type, you can make TypeScript infer the types implicitly:
```
k-yle, line 648: Another option that may be worth documenting:

If `CustomTheme` is the same across your whole application, TypeScript and JavaScript users can create a file called `global.d.ts` anywhere in their project. It is not imported by anything; TypeScript and/or VSCode's TSServer finds it automatically. It could like this:

```ts
declare global {
  namespace Jss {
    export interface Theme {
      themeColour: string;
    }
  }
}

export {};
```

The benefit is that you don't need to specify `CustomTheme` when you call `createUseStyles`. There is a bit more info in #1453

docs/react-jss.md
```diff
@@ -587,4 +587,101 @@ const Button = () => {
...
+function App(): React.ReactElement {
+  return (
+    <ThemeProvider theme={theme}>
+      <Button fontWeight="bold">Submit</Button>
+    </ThemeProvider>
+  )
+}
+```
+
+You only need to supply as much type information as you need. For instance, if you aren't using a `Theme`, then you don't need to supply a type for it. You'll notice that if you want to supply types for `Props` or `Theme`, you'll have to supply the type for `RuleNames` first. This means that if you want autocomplete for the `classes` that come from `useStyles`, you'll need to explicitly provide the strings you'll be using. This restriction is due to a limitation in TypeScript [that will hopefully get resolved soon](https://github.com/microsoft/TypeScript/pull/26349).
```
k-yle, line 644: I would consider extending the last sentence to say _"or, see the next section for a workaround"_. 

Just a thought, in case some people stop reading at the end of that section. There seems to be quite a misunderstanding about this point (https://github.com/cssinjs/jss/pull/1460#issuecomment-799478204, #1469)

docs/react-jss.md
ITenthusiasm: Yeah I was thinking about that, but I wasn't sure how to introduce it. What you provided gave me a good amount of inspiration though. Thanks!

docs/react-jss.md
ITenthusiasm: Yeah that's a good point. I didn't expect this to be such a large source of confusion. I tried to add some additional words like you said for clarity.

My recent commit has changes that address both of your suggestions. If you have free time, let me know what you think.

ITenthusiasm:
> Totally, I would only document APIs we want people to use. I thought about ThemeProvider, JssProvider, global Theme and probably some other things, I wold just go over the main document for react-jss and see what needs TS docs/examples

@kof It would definitely be nice to have TS documentation for those other guys. Unfortunately, I don't have much familiarity with them quite yet -- particularly `ThemeProvider` and `JssProvider`. (If they're just React components, then they hopefully wouldn't need much documentation, though -- if any.) At the same time, some information on how to use the hooks would certainly be helpful to have soon. It would help people who are starting to use TS, as well as anyone who needs to perform a migration or who wants to use the new changes.

After the existing documentation changes get approved, it could be good to launch out what we have now. Or is that too hasty? Thoughts?

> And I would move this to a react-jss-ts.md file and link it from multiple places where people would look for them as well as we need to add it to navigation structure on the documentation site

Are you imagining a navbar that looks something like this?

React JSS
 - API
 - TypeScript

I'd be fine with extracting out the current changes into their own file if that's desirable. However, I don't know how to make nested links appear in the navbar on the website.

docs/react-jss.md
k-yle: Looks great :+1:

docs/react-jss.md
```diff
@@ -587,4 +587,121 @@ const Button = () => {
 
 ## TypeScript
 
-TODO hooks support
```
kof, line 590: Lets move changes to react-jss-ts.md? So that we have space for all TS specific things and link that file from react-jss.md and other places where we find the right fit. Good places to link is super important for discoverability!

kof:
We can merge what we have now and add more later

ITenthusiasm:
@kof I moved the TS info to its own page as requested and squashed all the commits. Unless there are any other desired changes, this PR is ready to merge.

As I mentioned previously, I don't know how to make the TypeScript portion appear as a sub link under `React-JSS` (similar to how `API` and `Styles Syntax` are sub links under `JSS Core`). That would likely be good to have though.

kof:
Maybe for now it would be good to link from the current md pages to the TS document?

kof:
What about creating a "## Using with TypeScript" section in react-jss.md and a sentence and a link. Section because this way it would be much more visible than a link within a block of text anywhere

ITenthusiasm:
Pushed those new changes to `react-jss.md`. You can let me know if they're sufficient or not.

kof:
merged, thank you!

kof:

PR #1469, opened 2021-03-15
fix: resovle createUseStyle generics order

hosseinmd:

https://github.com/cssinjs/jss/pull/1460

Currently, for add props types we have to add classes name first.

I changed order of generic types of createUseStyle
Now we don't have to add classes name, because classes names recognizing from inserted object.

## Todo

## Changelog

Please summarize the changes in a way that makes sense inside the changelog. Feel free to add migration tips or examples if necessary.

### DIFF
--- a/packages/react-jss/src/index.d.ts
+++ b/packages/react-jss/src/index.d.ts
@@ -76,7 +76,7 @@ interface CreateUseStylesOptions<Theme = DefaultTheme> extends BaseOptions<Theme
   name?: string
 }
 
-declare function createUseStyles<C extends string = string, Props = unknown, Theme = DefaultTheme>(
+declare function createUseStyles<Props = unknown, Theme = DefaultTheme, C extends string = string>(
   styles: Styles<C, Props, Theme> | ((theme: Theme) => Styles<C, Props, undefined>),
   options?: CreateUseStylesOptions<Theme>
 ): (data?: Props & {theme?: Theme}) => Classes<C>
--- a/packages/react-jss/tests/types/createUseStyles.ts
+++ b/packages/react-jss/tests/types/createUseStyles.ts
@@ -39,7 +39,7 @@ const themeArg2ClassesFail2 = themeArg2({theme: expectedCustomTheme})
 const themeArg2ClassesPass = themeArg2({theme: expectedDefaultTheme})
 
 // Props declaration is allowed
-const themeArg3 = createUseStyles<string, MyProps>(theme => ({
+const themeArg3 = createUseStyles<MyProps>(theme => ({
   onlyPropsAllowed: ({...props}) => ({
     fontWeight: 'bold'
   })
@@ -52,7 +52,7 @@ const themeArg3ClassesPass = themeArg3(expectedCustomProps)
 const themeArg3ClassesPass2 = themeArg3({...expectedCustomProps, theme: expectedDefaultTheme})
 
 // Nested props declaration banned
-const themeArg4 = createUseStyles<string, MyProps>(theme => ({
+const themeArg4 = createUseStyles<MyProps>(theme => ({
   onlyPropsAllowed: ({...props}) => ({
     fontWeight: 'bold',
     // @ts-expect-error
@@ -61,7 +61,7 @@ const themeArg4 = createUseStyles<string, MyProps>(theme => ({
 }))
 
 // Supplied theme type is acknowledged
-const themeArg5 = createUseStyles<string, unknown, MyTheme>(theme => ({}))
+const themeArg5 = createUseStyles<unknown, MyTheme>(theme => ({}))
 // @ts-expect-error
 const themeArg5ClassesFail = themeArg5({theme: {}})
 // @ts-expect-error
@@ -129,7 +129,7 @@ const noThemeArg2ClassesFail2 = noThemeArg2({theme: expectedCustomTheme})
 const noThemeArg2ClassesPass = noThemeArg2({theme: expectedDefaultTheme})
 
 // Props declaration is allowed, but not nested props declaration
-const noThemeArg3 = createUseStyles<string, MyProps>({
+const noThemeArg3 = createUseStyles<MyProps>({
   propsAndTheme: ({property, theme}) => ({
     fontWeight: 'bold',
     // @ts-expect-error
@@ -149,7 +149,7 @@ const noThemeArg3ClassesPass = noThemeArg3(expectedCustomProps)
 const noThemeArg3ClassesPass2 = noThemeArg3({...expectedCustomProps, theme: expectedDefaultTheme})
 
 // Props and Theme types are properly acknowledged when supplied
-const noThemeArg4 = createUseStyles<string, MyProps, MyTheme>({
+const noThemeArg4 = createUseStyles<MyProps, MyTheme>({
   propsAndTheme: ({property, theme}) => ({
     fontWeight: 'bold',
     // @ts-expect-error
@@ -169,7 +169,7 @@ const noThemeArg4ClassesPass = noThemeArg4(expectedCustomProps)
 const noThemeArg4ClassesPass2 = noThemeArg4({...expectedCustomProps, theme: expectedCustomTheme})
 
 // Nested declarations are banned (single nest test)
-const noThemeArg5 = createUseStyles<string, MyProps, MyTheme>({
+const noThemeArg5 = createUseStyles<MyProps, MyTheme>({
   singleNest: {
     fontWeight: 'bold',
     singleValue: ({property, theme}) => '',
@@ -182,7 +182,7 @@ const noThemeArg5 = createUseStyles<string, MyProps, MyTheme>({
 })
 
 // Nested declarations are banned (double nest test)
-const noThemeArg6 = createUseStyles<string, MyProps, MyTheme>({
+const noThemeArg6 = createUseStyles<MyProps, MyTheme>({
   doubleNest: {
     fontWeight: 'bold',
     singleValue: ({property, theme}) => '',
@@ -199,7 +199,7 @@ const noThemeArg6 = createUseStyles<string, MyProps, MyTheme>({
 })
 
 // Nested declarations are banned (triple nest test)
-const noThemeArg7 = createUseStyles<string, MyProps, MyTheme>({
+const noThemeArg7 = createUseStyles<MyProps, MyTheme>({
   tripleNest: {
     fontWeight: 'bold',
     singleValue: ({property, theme}) => '',


ITenthusiasm:
@kof Please see the entire discussion that happened from [this comment](https://github.com/cssinjs/jss/pull/1460#issuecomment-799388866) onward in the linked PR.

If you decide you want go with these changes, that's fine. But from what I understand, I strongly recommend leaving the code as is.

hosseinmd:
That hard to define props, we don't have to define classes names.

kof:
Lets discuss this here, because the linked discussion is on a closed PR and this one seems like a good place to figure it out:

Do I understand correctly that you see different compiler errors from the same types? If yes what can be the reason, different TS versions?

ITenthusiasm:
The complaint doesn't seem to be about getting compiler errors for the same types. The complaint seems to be that hosseinmd doesn't like having to specify `RuleNames` first in order for him to specify `Props`.

kof:
@hosseinmd do you say the RuleNames type is inferred correctly for you and allows you to get the errors when using classes without manually specifying the RuleNames?

kof:
I am actually wondering if it's possible to not have to specify RuleNames, because technically they are already there in the object and TS could use those keys and map them to classes

ITenthusiasm:
> I am actually wondering if it's possible to not have to specify RuleNames, because technically they are already there in the object and TS could use those keys and map them to classes

@kof Unfortunately, that's not possible. I found this out the hard way while working on #1460, and k-yle seems to understand this problem as well (see #1453). I'll try to go through the logic of why... _but it's a lot of logic_. Note that I'm going to try to _condense_ this logic, and there is much more that can be said about why we're stuck in the situation of needing to define `RuleNames`.

### We Need `Record<Keys, Values>`

In order for users to get the correct types on their objects (properly typed `data` functions, CSS properties, etc.), we have to type the values of the object the user defines. But in order for us to type the values of the object, we also have to type the _properties_ of the object themselves. This necessitates something like `Record<Keys, Values>` or `{ [Key in string]: Value }`.

This issue is why React-JSS has the `Styles` type in the first place. It's basically a simplified version of the necessary `Record<Keys, Values>` type that users need to get intellisense on their style objects.

### We Need a Type for `Keys` in `Record<Keys, Values>`
_But the need to define keys in the first place is the problem_. Let's say we have some general object value type, like `JssStyles`. If we define an object as `Record<string, JssStyles>`, then TS will never provide intellisense on the object properties. We _explicitly_ told TypeScript, "You only need to enforce that the properties are strings." So this basically tells TypeScript, "You can use _any_ string you want as a property".

### Generic Types Help Us Get around Defining a Type for `Keys`
In React-JSS styles, we get around this by using generic types. This is because TypeScript will try to infer the generic types if we don't provide any type arguments. So if we define a new type `Styles<Names> = Record<Names, JssStyles>`, we can "force" TypeScript to infer the object properties. So for instance, doing this:

```typescript
const myStyles: Styles = {
  ruleName: { /* ... */ };
};
```

Will force TypeScript to infer that `myStyles` must be of type `Record<"ruleName", JssStyles>` because we didn't provide any generic type parameters. This is what many React-JSS users have been relying on. But here lies the problem.

### TypeScript _does not_ Infer _any_ Types When You Supply Generic Type Info
This is the thing that trips everyone up. And this is the key issue that k-yle, myself, and other developers have come to recognize. We're relying on the fact that TypeScript infers `RuleNames`. _But TypeScript stops inferring **any** types once the user supplies even a single generic type argument_. Because this is a very niche, fairly uncommon scenario, a lot of people can't wrap their heads around this problem until they run into it themselves. But it's true.

Microsoft has _tons_ of users complaining about this in confusion. You can see it my microsoft/TypeScript#43119 and in someone else's microsoft/TypeScript#39008. And [Ryan (member of Microsoft) makes it very clear](https://github.com/microsoft/TypeScript/issues/39008#issuecomment-642074983) that once you supply generic type information, the remaining generic type parameters are forced to their defaults.

### The Conclusion
All of this means that if we use a system like `<Props = unknown, Theme = DefaultTheme, Names extends string = string>`, users will try to specify `Props`, but will then inevitably lose type information on `Names`. More accurately, `Names` will be _not_ be inferred as `LIST_OF_STRINGS`, but will be _defaulted_ to `string`.

We could try to remove `Names` so that TypeScript can't default us to `Record<string, JssStyles>`, but then TypeScript would have no way to infer the `Keys` because we've removed that information. As Ryan said, we're in this situation because of a poor design in TypeScript, but that's just what we're stuck with.

This is why `<Names, Props, Theme>` is the _only_ possible order. Users will always have to specify `Names` once they supply _any_ generic type arguments. So it's better to make them do that up front instead of forcing them to reach over other irrelevant types to define (eg. `Theme`) to supply info for `Names`. More importantly, once microsoft/TypeScript#26349 gets merged, users will be able to do `const useStyles = createUseStyles<_, Props>`, which is much less painful.

In the meantime, I've tried to add docs to show people how to force TypeScript to infer  **all** types. This is why I recommended the function argument for `createUseStyles`. Speaking from a standpoint of TypeScript, it is the clearest, least confusing approach, though the object variant is still somewhat manageable.

ITenthusiasm:
I should note that once microsoft/TypeScript#26349 gets merged, I'm **hoping** we can get away with something like this:

```typescript
declare function createUseStyles<Props = unknown, Theme = DefaultTheme>(
  styles: Styles<_, Props, Theme> | ((theme: Theme) => Styles<_, Props, undefined>),
  options?: CreateUseStylesOptions<Theme>
): (data?: Props & {theme?: Theme}) => ClassesForStyles<typeof styles>
```

But that's mildly hackish and it's not readily apparent if that would be "legal" once the new changes get merged. Additionally, if we applied this change, users would never be able to specify strict rules on `RuleNames`. So even if this becomes possible in the future, we'll be forced with having to decide between:

1. Enabling users to get intellisense for `classes` in the easiest possible way or
2. Leaving `RuleNames` as something users will have to force TypeScript to infer through the `_` sigil. (Thus, the user would still have full control over `RuleNames` type enforcements.)

Regardless of the outcome for `createUseStyles`, any people using the `Styles` type for any reason would still have to provide `RuleNames` or the `_` sigil explicitly.

k-yle:
Adding to @ITenthusiasm's explanation: here's [an example in the typescript playground](https://www.typescriptlang.org/play?ts=4.2.3#code/FAEwpgxgNghgTmABAMwK4DsIBcCWB7dRCBGLMAVQGcwBlLATyjEoEYAeAYUTAA8z0QlRJSxwc6AOaIAvMNHiJAGkQAFOHgAOQ2RgDW6PAHd0ygCoALMAFsksgCJhkMVFCwXrYAHwAKYIjmMzABciHSBlJzKapqUZpY2nogAPoje3ljxYCHuNgCUMolhTBEcUepayhjgyOJgIJ65in6ImrgElAD8IRwkZFS0DMUA8hpt6BE5XsC5Id4gpDBdquVCAGSIAN4ZHkuTAL750okcsJTUJZ7AoJCwCCgY2PiExGCkFNRFzABMbNFaMog9AZjHEPACHE4XG5MsouLx+II5GJJACRMiJD5miJwiFPiUyjFQQlkql0plsplDoVBsxIstCYCBI5avVGs1Wk9Ot1eu8BuERmMJplLjNUvMsIsQn81pttjZdpkDgVECcYGdaRxLsAGBokNKARtEIYcCAMiF0KgrAAjMBwRB7ADcVwA9M7ENtqO7DHhEPAkNthDAbAA6d2WFpwcB2vDIMNICRgdC2nAQIRWGC6JAGRAgHDIZC2xMQMCh4CuxDmPAAN1tLRrdoARKgPjTWA3fQJEE2W+Evg3lAYyBWjHH6L67la8Bku6r1RE0QpPA3gBB2lhAT3iiwAS83v08ewF5ICVofBsDk7V+N182+cUvjuefvWz9pWeL1cywAqL9+L+IHAJAMCcwCgEdLAQP9ECgmDEH-WD4Lg6CkIQ5DEPQtDMNQ7CUNw9DnSuHUkAhZxXEmA0iDwMDUDgEIGwQEB2xSBsJAQRN20dQj6F1RAADEHl+MxlAAJUSWRTG4PhE0RKpmSTEBEA6UlxRgKUqUQYT8lmOYFilRB1i2clEFMJUjg03Irl4DQ8DgdciMQAApM4ADUYCgVBbGaFJvCPKQDIOLzUgta1a38iz-BSABBOA4Bgeg2F8klgptO0opiuKEvkFEUmS21EhSAByABCHArGs2yYHQLACsuCLEAtKAoECpwoGoSyeHKuzuKQJzKE+X4VgBSr6CJWxGWqFkxMCjZmn8fwAG0AGkAMITN6BjRAADkbPTKAODOP5bVwZgAF0glmua6u2uBdv2yhXPc2lFtqy65pSXr+ulUaXtekl+MwAaGUmZRrtuly3I8iJnpJD6aTYWSankyomURup8vGuS0Yuy6UiGK1qDgKsYCtJg2FBty7oeyG2Gh96zk+EkEcmi69mm7HECWkJfLO9m6t6qmwF5mH6bhr7jOFIWUn+iBAYqcWPGUfmIaQOm+rhpmkYx1H6kZlHmd+uq8YJomSbANglce4W1cCXWJvkn77Xazr3W60JWzYZpNqDJB4WkoREpyy0UpJSh6GtKjUSypRmn1HR0H0IwTGaci47tupgDEjTIBskAPf8L2bCaPmRcCWXYnlhJAt8wLpbL0bFZL0mNbqZG051lIA-qlwoFtzH6kCo3bRN0nYZtjuo6S7ve+1y4tXEMg4CcYsVSfTdmAFTk2BTxASKhSZEnPYArJsrqeNnc4ye9ySEX9ifA5C1LhDDyce9kXzM+E7PI0vwukUXK4gA) that shows that the order of the generics makes no difference

kof:
To understand the problem I am trying to reproduce it on a simple playground, here is what I got so far, but I am guessing it's not representing the actual problem. @ITenthusiasm what do I need to add there to make it show the problem? this will be maybe a good example to explain for future readers  https://www.typescriptlang.org/play?jsx=0&ts=4.2.3#code/MYewdgzgLgBFIGEA2BDCECmEYF4YB4BlKATySwD4AKDKALhgG8YBtAJQFdyA5FAWwwwAlmBgBrDCRAAzGMTJYAug2gAnEQHMYAXwCUuCjFoBuALAAoC6EiwIIAVAAWm3AXnkI1aAogN3WfRxDRgsYGGtocKRfGBQwEldGbVDw8EjgVHQsV3hkNEwIIlIPagyIXRTpEFUYKgjYCQSRGG8PfRDzMLCM-KwWRsVXAHI7B2cwDSGU5M6YVVoOVVEerIgLGas02BWC11HacY0qDrCUBgAiAEZzgBoUgCMGIYAmKfM9CytMguMgA

kof:
Updated the link

hosseinmd:
> I am actually wondering if it's possible to not have to specify RuleNames, because technically they are already there in the object and TS could use those keys and map them to classes

Yes, it is possible, if we move ruleNames to last generic param, so we just need to define props and ruleNames will be recognized

hosseinmd:
> @hosseinmd do you say the RuleNames type is inferred correctly for you and allows you to get the errors when using classes without manually specifying the RuleNames?

Yes

kof:
What version of TS are you all testing with? Just to make sure this is not the source of our differences

hosseinmd:
> What version of TS are you all testing with? Just to make sure this is not the source of our difference

I testing now, give me time.

hosseinmd:
@ITenthusiasm is right, RuleNames not inferred

The benefit of this PR is
```js 
//change this line 
const themeArg3 = createUseStyles<string, MyProps>
// to this line 
const themeArg3 = createUseStyles<MyProps>
```

hosseinmd:
Sorry this PR is wrong.

ITenthusiasm:
> but I am guessing it's not representing the actual problem

@kof Yeah that's not quite the same as the actual problem. There are a few reasons why, but I don't think diving into those reasons would advance the React-JSS discussion much more, and I don't want to write another chunk of text. @k-yle added a really good TypeScript playground to reflect the issue (thanks). I don't think a more relevant example can be added than his.

If someone wants to see a _basic_ example of the issue I was describing, microsoft/TypeScript#43119 is a sufficient example. If someone wants to see an example of this issue _in the context of React-JSS_, then [k-yle's TS playground](https://www.typescriptlang.org/play?ts=4.2.3#code/FAEwpgxgNghgTmABAMwK4DsIBcCWB7dRCBGLMAVQGcwBlLATyjEoEYAeAYUTAA8z0QlRJSxwc6AOaIAvMNHiJAGkQAFOHgAOQ2RgDW6PAHd0ygCoALMAFsksgCJhkMVFCwXrYAHwAKYIjmMzABciHSBlJzKapqUZpY2nogAPoje3ljxYCHuNgCUMolhTBEcUepayhjgyOJgIJ65in6ImrgElAD8IRwkZFS0DMUA8hpt6BE5XsC5Id4gpDBdquVCAGSIAN4ZHkuTAL750okcsJTUJZ7AoJCwCCgY2PiExGCkFNRFzABMbNFaMog9AZjHEPACHE4XG5MsouLx+II5GJJACRMiJD5miJwiFPiUyjFQQlkql0plsplDoVBsxIstCYCBI5avVGs1Wk9Ot1eu8BuERmMJplLjNUvMsIsQn81pttjZdpkDgVECcYGdaRxLsAGBokNKARtEIYcCAMiF0KgrAAjMBwRB7ADcVwA9M7ENtqO7DHhEPAkNthDAbAA6d2WFpwcB2vDIMNICRgdC2nAQIRWGC6JAGRAgHDIZC2xMQMCh4CuxDmPAAN1tLRrdoARKgPjTWA3fQJEE2W+Evg3lAYyBWjHH6L67la8Bku6r1RE0QpPA3gBB2lhAT3iiwAS83v08ewF5ICVofBsDk7V+N182+cUvjuefvWz9pWeL1cywAqL9+L+IHAJAMCcwCgEdLAQP9ECgmDEH-WD4Lg6CkIQ5DEPQtDMNQ7CUNw9DnSuHUkAhZxXEmA0iDwMDUDgEIGwQEB2xSBsJAQRN20dQj6F1RAADEHl+MxlAAJUSWRTG4PhE0RKpmSTEBEA6UlxRgKUqUQYT8lmOYFilRB1i2clEFMJUjg03Irl4DQ8DgdciMQAApM4ADUYCgVBbGaFJvCPKQDIOLzUgta1a38iz-BSABBOA4Bgeg2F8klgptO0opiuKEvkFEUmS21EhSAByABCHArGs2yYHQLACsuCLEAtKAoECpwoGoSyeHKuzuKQJzKE+X4VgBSr6CJWxGWqFkxMCjZmn8fwAG0AGkAMITN6BjRAADkbPTKAODOP5bVwZgAF0glmua6u2uBdv2yhXPc2lFtqy65pSXr+ulUaXtekl+MwAaGUmZRrtuly3I8iJnpJD6aTYWSankyomURup8vGuS0Yuy6UiGK1qDgKsYCtJg2FBty7oeyG2Gh96zk+EkEcmi69mm7HECWkJfLO9m6t6qmwF5mH6bhr7jOFIWUn+iBAYqcWPGUfmIaQOm+rhpmkYx1H6kZlHmd+uq8YJomSbANglce4W1cCXWJvkn77Xazr3W60JWzYZpNqDJB4WkoREpyy0UpJSh6GtKjUSypRmn1HR0H0IwTGaci47tupgDEjTIBskAPf8L2bCaPmRcCWXYnlhJAt8wLpbL0bFZL0mNbqZG051lIA-qlwoFtzH6kCo3bRN0nYZtjuo6S7ve+1y4tXEMg4CcYsVSfTdmAFTk2BTxASKhSZEnPYArJsrqeNnc4ye9ySEX9ifA5C1LhDDyce9kXzM+E7PI0vwukUXK4gA) is a sufficient example.

> this will be maybe a good example to explain for future readers

If we're talking about documenting things for future readers, then honestly, [my first comment on the matter](https://github.com/cssinjs/jss/pull/1469#issuecomment-803378791) and [k-yle's](https://github.com/cssinjs/jss/pull/1469#issuecomment-803459737) are enough. From this discussion and PR alone, there clearly isn't a "simple" or "easy" way to explain the problem in 2 sentences. And a random isolated example isn't going to make sense without the background explanation.

What's also clear is that many people will likely be unwilling to read something as large as a [condensed explanation](https://github.com/cssinjs/jss/pull/1469#issuecomment-803378791) anyway, especially when they're running with (understandable) presumptions about how TS should operate. And if that's the case, no amount of talking or isolated examples will help them. People will just have to run into the problem themselves and bang their heads against the wall like I had to do. They'll learn the truth one way or another by playing with the source code or [k-yle's playground](https://www.typescriptlang.org/play?ts=4.2.3#code/FAEwpgxgNghgTmABAMwK4DsIBcCWB7dRCBGLMAVQGcwBlLATyjEoEYAeAYUTAA8z0QlRJSxwc6AOaIAvMNHiJAGkQAFOHgAOQ2RgDW6PAHd0ygCoALMAFsksgCJhkMVFCwXrYAHwAKYIjmMzABciHSBlJzKapqUZpY2nogAPoje3ljxYCHuNgCUMolhTBEcUepayhjgyOJgIJ65in6ImrgElAD8IRwkZFS0DMUA8hpt6BE5XsC5Id4gpDBdquVCAGSIAN4ZHkuTAL750okcsJTUJZ7AoJCwCCgY2PiExGCkFNRFzABMbNFaMog9AZjHEPACHE4XG5MsouLx+II5GJJACRMiJD5miJwiFPiUyjFQQlkql0plsplDoVBsxIstCYCBI5avVGs1Wk9Ot1eu8BuERmMJplLjNUvMsIsQn81pttjZdpkDgVECcYGdaRxLsAGBokNKARtEIYcCAMiF0KgrAAjMBwRB7ADcVwA9M7ENtqO7DHhEPAkNthDAbAA6d2WFpwcB2vDIMNICRgdC2nAQIRWGC6JAGRAgHDIZC2xMQMCh4CuxDmPAAN1tLRrdoARKgPjTWA3fQJEE2W+Evg3lAYyBWjHH6L67la8Bku6r1RE0QpPA3gBB2lhAT3iiwAS83v08ewF5ICVofBsDk7V+N182+cUvjuefvWz9pWeL1cywAqL9+L+IHAJAMCcwCgEdLAQP9ECgmDEH-WD4Lg6CkIQ5DEPQtDMNQ7CUNw9DnSuHUkAhZxXEmA0iDwMDUDgEIGwQEB2xSBsJAQRN20dQj6F1RAADEHl+MxlAAJUSWRTG4PhE0RKpmSTEBEA6UlxRgKUqUQYT8lmOYFilRB1i2clEFMJUjg03Irl4DQ8DgdciMQAApM4ADUYCgVBbGaFJvCPKQDIOLzUgta1a38iz-BSABBOA4Bgeg2F8klgptO0opiuKEvkFEUmS21EhSAByABCHArGs2yYHQLACsuCLEAtKAoECpwoGoSyeHKuzuKQJzKE+X4VgBSr6CJWxGWqFkxMCjZmn8fwAG0AGkAMITN6BjRAADkbPTKAODOP5bVwZgAF0glmua6u2uBdv2yhXPc2lFtqy65pSXr+ulUaXtekl+MwAaGUmZRrtuly3I8iJnpJD6aTYWSankyomURup8vGuS0Yuy6UiGK1qDgKsYCtJg2FBty7oeyG2Gh96zk+EkEcmi69mm7HECWkJfLO9m6t6qmwF5mH6bhr7jOFIWUn+iBAYqcWPGUfmIaQOm+rhpmkYx1H6kZlHmd+uq8YJomSbANglce4W1cCXWJvkn77Xazr3W60JWzYZpNqDJB4WkoREpyy0UpJSh6GtKjUSypRmn1HR0H0IwTGaci47tupgDEjTIBskAPf8L2bCaPmRcCWXYnlhJAt8wLpbL0bFZL0mNbqZG051lIA-qlwoFtzH6kCo3bRN0nYZtjuo6S7ve+1y4tXEMg4CcYsVSfTdmAFTk2BTxASKhSZEnPYArJsrqeNnc4ye9ySEX9ifA5C1LhDDyce9kXzM+E7PI0vwukUXK4gA).

I can't help anyone further than what I've said. And at this point I've given the same explanations and set of references at least 2-3 times between this PR and #1460.

kof:
@ITenthusiasm yeah, totally understandable, it must be frustrating, consider this as teaching us

I haven't used typescript yet (at the job still flowtype), will soon start using it eventually though.

I played a bit more and found this workaround as alternative to typing out the list of rule names:

https://www.typescriptlang.org/play?jsx=0&ts=4.2.3#code/C4TwDgpgBAyqA2EBKBXRUC8UCGA7EA9AUhAMYD2ATgCYA8A5BfFfQDRQDOwlAlrgOYA+ALAAoMRVxcowcgGF42DhwgdMUWnBCIOggBSlFy1QC4oAbygBtVIgBy2ALbQ+UANYQQ5AGawEqgF0zLl4BKABfAEpMQShDJRUOAG4xCXIpYDjyRzAeRC1EW2gsWgAFSnIwXT0ubWQ0CDMC+sR2MAqqs3LKtQAfHHxIpv8imIsxKCgeXz16DmyIbrB6Kdwodp7o8wnJ9Y6OADp55yWU8VFJ8J3KCGAUSjXawoaxK-PJaW8Kx2aOABV5EZEuoyvt2L99E9TH46hw2vsuvtohhYtsLlkMnEgdC8CB1OY3pNvFQoAZ0tJKA0HM5Vpx-BwtjtJh9MlDRlgoRwrJT7E4IAEmViEqpuVS+QF1BQcnkIM0iqCepCRg14ZsdoSoEQoH9ePx+BBKKtvAaILhSBBrrd7mtZAphRxNPT9PFjAzXql3uTWfT8TtsGY0bsssxKGZ6DdqPR1awdgAjAOCphUMP8G6mqPot5vMSgSBQJZqLAkCg0BjHRYdNhQXAoRyxg0iT2YjZVfHlpZmACM2ab0hdwKwX2yvwBdtdCqq7FzEB8dNhSthqqqkQ9-dUB2wSSgq+xh1IW6AA

kof:
closing this PR as we figured this out

kof:
Holy s. this thing falls apart as soon as I try declaring styles as a separate object, Props doesn't get used any more for function values

PR #1470, opened 2021-03-18
cover undef possibility on getHead

JorgeTmrz:
    TypeError: Cannot read property 'appendChild' of undefined

## Changelog
I just added a condiitonal access operator becuase it was failing on my jest test suites
at DomRenderer.js (Line: 249)

image

### DIFF
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -246,7 +246,7 @@ function insertStyle(style: HTMLElement, options: PriorityOptions) {
     return
   }
 
-  getHead().appendChild(style)
+  getHead()?.appendChild(style)
 }
 
 /**


kof:
Something is wrong with your setup, jest should provide JSDOM with a head element ready

JorgeTmrz:
Thanks for the fast response, but i think that's the best solution bcz when I handle it with the conditional access operator all my tests suites run properly, I'm working with react.

image

kof:
There are more problems you are not aware of with this. If you don't render to JSDOM, you can disable DOMRenderer

JorgeTmrz:
So I must be wrong, sorry for wasting your time bro, one last thing, how can I disable the DOMRenderer?

kof:
https://cssinjs.org/jss-api?v=v10.6.0#setup-jss-instance

Renderer: null

kof:
I am assuming you want to skip css rendering and only test components.

JorgeTmrz:
Thanks, have a nice day, greetings from Dominican Republic dude! 🇩🇴

PR #1477, opened 2021-03-23
Add sideEffects: false hint

kof:
Webpack needs it to know there are no side effects
https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free and bundlephobia relies on it too


packages/css-jss/package.json
```diff
@@ -40,5 +40,6 @@
     "@babel/runtime": "^7.3.1",
     "jss": "10.6.0",
     "jss-preset-default": "10.6.0"
-  }
+  },
+  "sideEffects": false
```
TrySound, line 44: I would put it near entry points

TrySound (approved):

kof:

PR #1478, opened 2021-03-23
Remove gitter from the readmes, we are using discussions now

kof:


kof:
I am wondering if we need to link to the discussions anywhere else, because, it's quite obvious when it's enabled

hosseinmd (approved):

PR #1487, opened 2021-04-02
fix(jss): Fix the dynamic expanded style update bug.

### LINKED ISSUE #1051: [react-jss]: Dynamic expanded styles are not updated correctly
__Expected behavior:__
Dynamic styles should correctly override the previously applied styles.

__Describe the bug:__
Dynamic styles did not correctly override the previously applied styles.

__Codesandbox link:__ 
https://codesandbox.io/s/742q2630yj ( Click "All Borders" button to toggle between two sets of styles )

 - jss: 10.0.0-alpha.12
 - Browser [e.g. chrome, safari]: chrome
 - OS [e.g. Windows, macOS]: Windows

Wind4:
## Corresponding Issue(s): 

Fixes #1051

- Adjust the style processing order, the invalid props will remove before add/update.

https://github.com/cssinjs/jss/blob/6f151d497383f557fb69ee8bc2825373e75d63fe/packages/jss/src/RuleList.js#L219-L238

Currently, the props was update/add before remove, for example:

Before styles:
```
{
  "margin-top": "1px",
}
```

Updated styles 
```
{
  "margin": "2px",
}
```
The jss processing order is: 
> setProperty('margin', '1px');
> removeProperty('margin-top');

The problem is `setProperty('margin')` behavior will set `margin-top` `margin-left` `margin-right` `margin-bottom` to same values,  after the `removeProperty` remove `margin-top` property, the finally style will lost `margin-top`.

- Fix `style[prop]` remove bug

https://github.com/cssinjs/jss/blob/6f151d497383f557fb69ee8bc2825373e75d63fe/packages/jss/src/plugins/styleRule.js#L59-L65

Because `this.style` was updated by *RuleList*, so `const isDefined = name in this.style ` always be `false`. the `remove` flag was same to.

## Todo

### DIFF
--- a/packages/jss/src/RuleList.js
+++ b/packages/jss/src/RuleList.js
@@ -216,17 +216,6 @@ export default class RuleList {
       // We need to run the plugins in case new `style` relies on syntax plugins.
       plugins.onProcessStyle(styleRule.style, styleRule, sheet)
 
-      // Update and add props.
-      for (const prop in styleRule.style) {
-        const nextValue = styleRule.style[prop]
-        const prevValue = style[prop]
-        // We need to use `force: true` because `rule.style` has been updated during onUpdate hook, so `rule.prop()` will not update the CSSOM rule.
-        // We do this comparison to avoid unneeded `rule.prop()` calls, since we have the old `style` object here.
-        if (nextValue !== prevValue) {
-          styleRule.prop(prop, nextValue, forceUpdateOptions)
-        }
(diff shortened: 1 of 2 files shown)


kof:
This looks great actually, we just need a test in jss package to verify this behavior, otherwise seems like a good PR!

kof:
Hi @Wind4, why did you close it?

PR #1496, opened 2021-04-22
simplify findHigherSheet in dom renderer

behnammodi:
Used `find` instead of `for` to simplify and reduce the size

### DIFF
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -130,17 +130,14 @@ const getHead = memoize((): HTMLElement => (document.querySelector('head'): any)
  * Find attached sheet with an index higher than the passed one.
  */
 function findHigherSheet(registry: Array<StyleSheet>, options: PriorityOptions): StyleSheet | null {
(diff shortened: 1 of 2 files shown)


packages/jss/src/DomRenderer.js
```diff
@@ -130,17 +130,14 @@ const getHead = memoize((): HTMLElement => (document.querySelector('head'): any)
...
-    if (
-      sheet.attached &&
-      sheet.options.index > options.index &&
-      sheet.options.insertionPoint === options.insertionPoint
-    ) {
-      return sheet
-    }
-  }
-  return null
+  return (
```
kof, line 133: this is mostly a cosmetic change that is also slower (on a micro level), we optimize everything in this lib to get the smallest bit of performance.

behnammodi: So you mean this PR is not a good effect on JSS?

behnammodi: I agree `for` is faster than `find`

kof: thank you

PR #1499, opened 2021-04-22
Implement css rule type

behnammodi:
Simplified some type by union type

### DIFF
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -33,7 +33,9 @@ const memoize = <Value>(fn: () => Value): (() => Value) => {
   }
 }
 
-type GetPropertyValue = (HTMLElementWithStyleMap | CSSStyleRule | CSSKeyframeRule, string) => string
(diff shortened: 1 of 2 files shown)


packages/jss/src/DomRenderer.js
```diff
@@ -33,7 +33,9 @@ const memoize = <Value>(fn: () => Value): (() => Value) => {
   }
 }
 
-type GetPropertyValue = (HTMLElementWithStyleMap | CSSStyleRule | CSSKeyframeRule, string) => string
+type CSSRule = HTMLElementWithStyleMap | CSSStyleRule | CSSKeyframeRule
```
kof, line 36: this is semantically incorrect, because HTMLElementWithStyleMap is not a CSS Rule, that's why it wasn't done this way

kof: I think trying to call it something is not worth the effort

behnammodi: Yes this is a mistake, I will to close this PR

Issue #1462, opened 2021-03-08
Atomic generator 


behnammodi:
We generate a lot of common style somethings like this:

```css
.a1-0-1-105 {
  display: none;
}

.a2-0-1-106 {
  display: none;
}
```

And style sheet size is increasing in a large project

We can use an atomic style generator, Some of the libraries did it like: 

https://github.com/robinweser/fela

And

https://github.com/styletron/styletron

Are you willing to implement it? Yes

kof:
We discussed this thing a bit, there are certain downsides coming with atomic approach, but I encourage you to create an AtomicDomRenderer that would render out the rules in an atomic way (one class per property).

behnammodi:
@kof Thank you, I will try to do that, but maybe somewhere I need your help

behnammodi:
@kof Hi man, I investigate the entire project to create an Atomic generator, I think we need a new class of `BaseStyleRule` for that.

That is right?

kof:
I was hoping its possible to do by just suppling a different DomRenderer implementation, without modifying antything else, renderer can be passed to Jss as an option. 

Or alternatively, which is eventually a simpler option, but would not work with vanilla JSS and not with any other integrations and would require to fork react-jss - creating a new rule per property from createUseStyles.

behnammodi:
I will investigate more about DomRenderer.

behnammodi:
I created a PR about that https://github.com/cssinjs/jss/pull/1501

behnammodi:
I hard work on DomRenderer, so I have to say it not possible to have Atomic Generator just by changing DomRender

I think this future is very important because the new approach of CSS in JS is about Atomic generation. I could find a new repo about that: https://github.com/johanholmerin/style9

and a video of Facebook on youtube about CSS in JS: https://www.youtube.com/watch?v=9JZHodNR184

kof:
Can you show me what you did and what was impossible? 

Regarding atomic - it's absolutely note new. It's been originally in react-native-web, then they migrated to partially using regular rules and partially atomic. There is fela and a bunch of others which do atomic. There is a number of downsides on this approach and it should be up to the user what they choose. What I am saying is that it's not a new trend but still a good thing to have as an option.

behnammodi:
@kof I said it not possible by `DomRenderer`

We need to compute on `Rule` before giving it to `DomRenderer`. I create a flow for that, so we have to have a process like this:

[...]

PR #1500, opened 2021-04-22
Ship memoize to utils

behnammodi:
`memoize` can use in other `Renderer` class such as `AtomicDomRenderer`, so I shipped to utils for use in future

### DIFF
--- a/packages/jss/src/utils/memoize.js
+++ b/packages/jss/src/utils/memoize.js
@@ -1,0 +1,12 @@
+/**
+ * Cache the value from the first time a function is called.
+ */
+const memoize = <Value>(fn: () => Value): (() => Value) => {
+  let value
+  return () => {
+    if (!value) value = fn()
+    return value
+  }
+}
+
+export default memoize
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -3,6 +3,7 @@ import warning from 'tiny-warning'
 import StyleSheet from './StyleSheet'
 import sheets from './sheets'
 import toCssValue from './utils/toCssValue'
+import memoize from './utils/memoize'
 import type {
   CSSStyleRule,
   CSSMediaRule,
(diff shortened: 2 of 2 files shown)


packages/jss/src/utils/memoize.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * Cache the value from the first time a function is called.
+ */
+const memoize = <Value>(fn: () => Value): (() => Value) => {
```
kof, line 4: 1. just exporting it here is not making it public, you would have to export it from index
2. I think this function should stay internal, it's 3 lines of code and making it public is going to enforce maintenance on this and backwards compatibility

kof: maybe you were thinking to add atomic dom renderer to the core lib? I think it should for now stay a standalone package.

behnammodi: Woo, yes I wanted to add `AtomicDomRenderer` sibling of `DomRenderer`

kof: The problem with this is that it will increase a bundle size while being optional.

kof: I mean in case someone  can't treeshake or for example bundlephobia is gonna show higher bundle size. I am actually planing on removing some stuff or moving it to separate package when it's not always used.

behnammodi: I can help you for removing some stuff to reduce the package's size

behnammodi: But now, I don't have any idea for separate stuff from `jss` package but I try to find

kof: I recommend to focus on creating the AtomicDomRenderer without doing any changes to jss itself, copy what you need first, then we will see if something needs to be done on jss side.

behnammodi: Ok

behnammodi: I will close this PR and continue the discussion on https://github.com/cssinjs/jss/issues/1462

PR #1503, opened 2021-04-24
use .has instead of .get

behnammodi:
I think this was missed


kof:
lgtm, thanks

behnammodi:
@kof You’re welcome, can I join to jss as a member?

kof:
Totally!

kof:

PR #1509, opened 2021-05-05
Move flow types to .flow typedef files

kof:
#1035


packages/jss/src/PluginsRegistry.js.flow
```diff
@@ -0,0 +1,41 @@
...
+  onCreateRule: Array<OnCreateRule>,
+  onProcessRule: Array<OnProcessRule>,
+  onProcessStyle: Array<OnProcessStyle>,
+  onProcessSheet: Array<OnProcessSheet>,
+  onChangeValue: Array<OnChangeValue>,
+  onUpdate: Array<OnUpdate>
+}
+
+declare export default class PluginsRegistry {
+  plugins: {|
```
HenriBeck, line 29: Would we even need to declare this as I wouldn't consider this to be a "public" API and shouldn't really be accessed from the outside?

HenriBeck (approved):

packages/jss/src/PluginsRegistry.js.flow
kof: yeah, was my question as well, since typings are only used from outside,  I am not sure internally used types provide any additional type safety, unless used from outside, but thats not going to happen

HenriBeck: I think it should be removed as long as we only want to use these types for the outside and not for internal type checking

kof: Just tried, can't remove completely all private fields from .flow, it's still gonna be a mixture of public apis and internal once, otherwise it won't compile

kof: Because some fields are accessed by jss core itself, from other modules, for actual public api we would need to differentiate that and the access from public apis

kof:
I finally finished moving all flow types in all packages

kof:
merged now we can add ts types

kof:
released in https://github.com/cssinjs/jss/releases/tag/v10.8.0

PR #1523, opened 2021-06-23
Fix media query issue

mjeanroy:
## Corresponding Issue(s): 

#1320

The goal is to fix an issue regarding media query rules when defined in a nested & dynamic rule.

## Todo

## Expectations on Changes

/

## Changelog

I already gave a detailed explanation [here](https://github.com/cssinjs/jss/issues/1320#issuecomment-866899968).

Please note that this PR is just a basic fix, it's probably not clean but I open this PR to get your feedback on this 🙏

### DIFF
--- a/packages/jss-plugin-nested/src/index.js
+++ b/packages/jss-plugin-nested/src/index.js
@@ -98,13 +98,16 @@ export default function jssNested(): Plugin {
         container.addRule(selector, style[prop], {...options, selector})
       } else if (isNestedConditional) {
         // Place conditional right after the parent rule to ensure right ordering.
-        container
-          .addRule(prop, {}, options)
-          // Flow expects more options but they aren't required
-          // And flow doesn't know this will always be a StyleRule which has the addRule method
-          // $FlowFixMe[incompatible-use]
-          // $FlowFixMe[prop-missing]
-          .addRule(styleRule.key, style[prop], {selector: styleRule.selector})
+        const queue = container.prepareQueue()
+        const addedRule = container.createRule(prop, {}, options)
+
+        // Flow expects more options but they aren't required
+        // And flow doesn't know this will always be a StyleRule which has the addRule method
+        // $FlowFixMe[incompatible-use]
+        // $FlowFixMe[prop-missing]
+        addedRule.addRule(styleRule.key, style[prop], {selector: styleRule.selector})
+
+        container.deployRule(addedRule, queue)
       }
 
       delete style[prop]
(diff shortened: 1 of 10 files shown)


packages/jss-plugin-nested/src/index.js
```diff
@@ -98,13 +98,16 @@ export default function jssNested(): Plugin {
...
       } else if (isNestedConditional) {
         // Place conditional right after the parent rule to ensure right ordering.
-        container
-          .addRule(prop, {}, options)
-          // Flow expects more options but they aren't required
-          // And flow doesn't know this will always be a StyleRule which has the addRule method
-          // $FlowFixMe[incompatible-use]
-          // $FlowFixMe[prop-missing]
-          .addRule(styleRule.key, style[prop], {selector: styleRule.selector})
+        const queue = container.prepareQueue()
```
mjeanroy, line 101: I think that this code exposes too much internal API, it probably needs to be changed.

kof: yeah, definitely, need to find a cleaner way

kof:
I am a bit surprised that this could be a fix for that linked bug. Can you please create a failing test? A test would be super valuable here to make sure we are fixing the same right thing

packages/jss-plugin-nested/src/index.js
kof: The logic with the queue is a mess :)

kof: It wasn't super lean before, but now with queue from the instance and from an argument in deployRule, it became even worse

packages/jss-plugin-nested/src/index.js
```diff
@@ -98,13 +98,16 @@ export default function jssNested(): Plugin {
...
+        const addedRule = container.createRule(prop, {}, options)
+
+        // Flow expects more options but they aren't required
+        // And flow doesn't know this will always be a StyleRule which has the addRule method
+        // $FlowFixMe[incompatible-use]
+        // $FlowFixMe[prop-missing]
+        addedRule.addRule(styleRule.key, style[prop], {selector: styleRule.selector})
```
kof, line 108: Essentially all this is about ability to call an .addRule on a rule that implements ContainerRule without rendering anything (only adding to a queue) and then deploying the container rule as a next step. Is that correct?

I think it would be much easier to find a better API that allows this if we had tests to verify the behavior, because  right now trying things out without a test is way too hard.

kof:
I think you are on a good path, we need tests and we need a much cleaner api.

hkar19:
may we have patch package for this commit please?

caitlinwhiteley:
Hi, are there any updates on the status of this PR?

PR #1526, opened 2021-06-30
enhancement: jss-syntax-documentation-for-high-specificity

Steffan-Ennis:
Document more uses for `&` when dealing with high specificity selectors

## Changelog

Added more documentation for use cases when using `&`.

### DIFF
--- a/docs/jss-syntax.md
+++ b/docs/jss-syntax.md
@@ -338,6 +338,31 @@ Compiles to:
 }
 ```
 
+## Specific selectors
+
+When extending third party libraries with high secificity selector it's often necessary to have also have a high specificity.
+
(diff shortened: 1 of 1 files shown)


kof:
Nesting is part of jss-plugin-nested package .... so we need to move that to that package ...

If you are wondering if that will be discovered - it's problematic, we need to change the site in a way that allows people scroll through all syntax extensions at once from one page.

Steffan-Ennis:
Ok good to know.

kof:
Ok I am moving it myself ...

kof:
merged, thanks

Steffan-Ennis:
thanks

PR #1538, opened 2021-08-01
update for jss-plugin-extend

heyymarco:
Hi, i've made bug-fix for the `jss-plugin-extend`.

On the scenario like this: 
[See the sandbox](https://codesandbox.io/s/jss-plugin-extend-fix-l5phj)
```
const btnGrand = {
  color: 'red',
  background: 'pink',
};

const btnParent = {
  // extend using array
  extend: [
    btnGrand,
  ],
};

const btnChild = {
  // extend using object
  extend: btnParent,
};

export default {
  'btnChild': btnChild,
};
```
the `btnChild` extend `btnParent` as an object, and  `btnParent` extend `btnGrand` as an array.  
The plugin fails to create correct css, because on the original file `/packages/jss-plugin-extend/src/index.js`, line 42:
```
extend(style.extend.extend, rule, sheet, newStyle)
continue
```
It assumes that the grandpa's type is a literal object, but actually its type might be string, literal object, string array, literal object array, or mixed type array.

Because I need the JSS to work correctly (I'm developing a css framework based on your JSS lib),  
so I created my own **plugin-extend**.  
I wrote it using TypeScript, I'm not familiar with Flow, sorry.

To easily handle 5 kinds of extend types: `object`, `string`, `object[]`, `string[]`, `(object|string)[]`,  
the trick is convert the extend to array if not an array, otherwise keep the array, then loop it:
```
    // if `extend` is an `Array` => loop it
    // otherwise => convert to single `Array` => loop it
    for (const singleExtend of (Array.isArray(extend) ? extend : [extend])) {
```
so, i just handle only for the `object` & `string` inside the loop.

I also updated the TS definition for the `JssStyle`, so it accept `extend` property correctly:
```
// upgrade `JssStyle` definition:
type Optional<T>        = T|null|undefined
type ExtendableObject   = JssStyle|string // extend using a JssStyle object or using a rule name
type SingleExtend       = Optional<ExtendableObject>
type Extend             = SingleExtend|SingleExtend[]
type Style              = JssStyle & { extend?: Optional<Extend> } // add `extend` prop into `JssStyle`
// export the upgraded `JssStyle`:
export type { Style, Style as ExtendableStyle, Style as JssStyle }
```

### DIFF
--- a/packages/jss-plugin-extend/src/jss-plugin-extend.tsx
+++ b/packages/jss-plugin-extend/src/jss-plugin-extend.tsx
@@ -1,0 +1,172 @@
+// jss   (builds css  using javascript):
+import type {
+    Plugin,
+    JssStyle,
+
+    Rule,
+    StyleSheet,
+}                           from 'jss'           // base technology of our nodestrap components
+
+// others:
+import warning              from 'tiny-warning'
+
+
+
+// utilities:
+type LiteralObject      = { [key: string]: any }
+const isLiteralObject   = (object: any): object is LiteralObject => object && (typeof(object) === 'object') && !Array.isArray(object);
+
+// upgrade `JssStyle` definition:
+type Optional<T>        = T|null|undefined
+type ExtendableObject   = JssStyle|string // extend using a JssStyle object or using a rule name
+type SingleExtend       = Optional<ExtendableObject>
+type Extend             = SingleExtend|SingleExtend[]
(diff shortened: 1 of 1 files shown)


kof:
Hey, thanks for the fix. To make it mergeable, we need a few things:
1. Add a test
2. Reduce the number of changes to a bare minimum.
3. Keep the current formatting and commenting style.

PR #1543, opened 2021-08-18
fix(react-jss): fix order of definition variable

behnammodi:
We declare a variable that is not used in some cases so I changed the order


kof:
lgtm, thanks

Issue #1458, opened 2021-03-01
Jss generates the same class name for dynamic rendering


heyymarco:
__Expected behavior:__
the class name suffix **should be unique** (by the counter increment).
![same](https://user-images.githubusercontent.com/31385284/109466308-eecd1480-7a97-11eb-927c-738c1954a6a4.png)

__Describe the bug:__
I created 2 simple react component, called **ChildComponent** & **ParentComponent**.
The ParentComponent renders the ChildComponent dynamically based on the states or props or both.
When the parent render the child, the child class name is exactly the same as the parent, thus make collision - the parent style overwrites the child.

__Reproduction:__ 

see this sandbox:
https://codesandbox.io/s/create-react-app-forked-go82n?file=/src/App.jsx

then check the checkbox to show the component,
then open dev tool, find the generated css.

__Versions (please complete the following information):__
 - react-jss: 10.5.1
 - Browser: chrome
 - OS : windows

Feel free to add any additional versions which you may think are relevant to the bug.

__Managing expectations:__

Maintainers will not be fixing the problem you have unless they have it too, if you want it to get fixed:

1. Submit a PR with a failing test
2. Discuss a solution
3. Implement it

You can also do the first step only and wait for someone else to work on a fix. Anything is much better than nothing.

heyymarco:
have you fix the bug? I think the bug around the `react-jss` => `getSheetIndex()`. It rely the increment _each time_ a new sheet generated. The index is remain the same if the components are statically rendered, but for dynamically rendered the index possibility different depend on the number of rendered unique components.

quarryman:
Same issue here. 
Does JSS team has plans to fix this?

Only workaround  we see is to pass Component.displayName to createUseStyles that is not minified on production build and makes classname unique

` const useStyles = createUseStyles(styles, { name: Component.displayName });`

kof:
Please provide a failing test and feel free to prototype a fix!

lfabl:
> Same issue here. Does JSS team has plans to fix this as this probably affects everyone with more or less standard JSS setup
> 
> Only workaround we see is to pass Component.displayName to createUseStyles that is not minified on production build and makes classname unique
> 
> ` const useStyles = createUseStyles(styles, { name: Component.displayName });`

Great!. It is work correctly. But i think it is not best practise method. It must provide to global context solution.

PR #1546, opened 2021-08-20
perf(react-jss): keep all result of executed function in map

behnammodi:
Now `memoize` function keep all result of a function that improves performance


packages/react-jss/src/utils/memoizeOne.js
```diff
@@ -6,28 +6,18 @@ type ComputeFn = (...args: Args) => Result
...
-      for (let i = 0; i < args.length; i++) {
-        if (args[i] !== lastArgs[i]) {
-          isSame = false
-        }
-      }
-
-      if (isSame) {
-        return lastResult
-      }
+    const key: string = JSON.stringify(args)
```
kof, line 12: this is very questionable, you don't know if the data is serializable

jvanderen1: Instead of serializing this, can you use some hashing function (such as [object-hash](https://www.npmjs.com/package/object-hash))?

behnammodi: Yes, but we used it where we sent an array

behnammodi: Also, all tests passed

behnammodi: > Instead of serializing this, can you use some hashing function (such as [object-hash](https://www.npmjs.com/package/object-hash))?

Why do we have to add a 10KB package for that?

behnammodi: > this is very questionable, you don't know if the data is serializable

Yes, but we know where we used it

jvanderen1: You guys are already using Terser in your rollup.config, so it's probably not a big deal.

behnammodi: > You guys are already using Terser in your rollup.config, so it's probably not a big deal.

I know, but is that less than 14 bytes? also `JSON.stringify` is a native method, it is faster than other solution

behnammodi: Aim of `object-hash` is different is not good solution for that

behnammodi: @kof What do you think about `object-hash`? I disagree

kof: I just checked, we only use memoizeOne to memoize classes object, that means
1. we know it is flat
2. we know it is serializable 

Sorry for confusion, I haven't looked up before where it isused as I wrote that message

kof:
My next question is where does the performance benefit come from, what makes it faster and by how much?

packages/react-jss/src/utils/memoizeOne.js
behnammodi: No problems, so I will resolve this conversation

kof: btw. the reason I got confused is that I though we use it in other places too, so if we are to deploy this assumptions, we have to rename this function to something that reflects them

behnammodi: Exactly, we need a new PR for that

kof: haha no, if you modify what function does, you also have to rename it in the same PR, but there is no problem here.

behnammodi:
> My next question is where does the performance benefit come from, what makes it faster and by how much?

So my response is: the previous version of this module just save the previous result, so if we call it by new parameters, the previous result will clear and new result save on there, but in the new version, we keep all result so we have access all previous result event we call function with new parameters

packages/react-jss/src/utils/memoizeOne.js
behnammodi: Why are you laghing? :) ok, I will rename this file to `memoize`, are you agree?

kof:
> > My next question is where does the performance benefit come from, what makes it faster and by how much?
> 
> So my response is: the previous version of this module just save the previous result, so if we call it by new parameters, the previous result will clear and new result save on there, but in the new version, we keep all result so we have access all previous result even we call the function with new parameters

Sure, but that means you have the situation that you frequently get the previous classes object and I don't see why this should be a regular scenario

behnammodi:
> > > My next question is where does the performance benefit come from, what makes it faster and by how much?
> > 
> > 
> > So my response is: the previous version of this module just save the previous result, so if we call it by new parameters, the previous result will clear and new result save on there, but in the new version, we keep all result so we have access all previous result even we call the function with new parameters
> 
> Sure, but that means you have the situation that you frequently get the previous classes object and I don't see why this should be a regular scenario

So your mean is `sheetClasses` and `props.classes` never change?

```js
const classes = mergeClassesProp(sheetClasses, props.classes)
```

kof:
> So your mean is sheetClasses and props.classes never change?

No. I mean when they change, they change in a new way every time, they don't go back to the previous version.

behnammodi:
> > So your mean is sheetClasses and props.classes never change?
> 
> No. I mean when they change, they change in a new way every time, they don't go back to the previous version.

I'm not sure, imagine user switch between some classes by a toggle action

kof:
sure, but even if this happens, we have to rerender the user component anyways, because this can only happen if user passes  a new classes object. So all you would be optimizing in this case is the actual mergeClasses function, right?

behnammodi:
> sure, but even if this happens, we have to rerender the user component anyways, because this can only happen if user passes a new classes object. So all you would be optimizing in this case is the actual mergeClasses function, right?

Exactly

kof:
Ok so if we are optimizing mergeClasses function itself, I don't see why ... its super fast and your memoization would keep in memory all versions of classes objects ever created through the entire application lifecycle

kof:
The way you did it will never actually clean those objects from the memory, its a memory leak.

behnammodi:
No, cache inside `memoize` function, is not globally

behnammodi:
We use memory until the component is live

kof:
> We use memory until the component is live

mergeClassesProp is the reference to the memoization function that holds the cache.
mergeClassesProp is created every time we call withStyles(styles)(Component), so if you have 100 of these calls, mergeClassesProp will persist 100 times and each mergeClassesProp will keep references to every classes object that is created during application life time

since on SSR module is imported once - this is going to grow the cache until the server restarts

behnammodi:
> > We use memory until the component is live
> 
> mergeClassesProp is the reference to the memoization function that holds the cache.
> mergeClassesProp is created every time we call withStyles(styles)(Component), so if you have 100 of these calls, mergeClassesProp will persist 100 times and each mergeClassesProp will keep references to every classes object that is created during application life time
> 
> since on SSR module is imported once - this is going to grow the cache until the server restarts

I got it, what about this change? 

```jsx
  const mergeClassesProp = useMemo(
      () =>
        memoize(
          (sheetClasses, classesProp): Classes =>
            classesProp ? mergeClasses(sheetClasses, classesProp) : sheetClasses
        ),
      []
    )
```

and moved it to `WithStyles` component

kof:
Its possible to memoize per instance or even per styles, the question is still what we are optimizing, mergeClasses is very fast  and to hit the optimization you need specific scenario, where I don't see how this would be a huge benefit, since the component would be updating anyways

behnammodi:
> Its possible to memoize per instance or even per styles, the question is still what we are optimizing, mergeClasses is very fast and to hit the optimization you need specific scenario, where I don't see how this would be a huge benefit, since the component would be updating anyways

Could you please see this [benchmark](https://perf.link/#eyJpZCI6IjdqdDdvbDNnZmRiIiwidGl0bGUiOiJKU1MiLCJiZWZvcmUiOiJjb25zdCBtZW1vaXplMSA9IChmbikgPT4ge1xuICBjb25zdCBjYWNoZSA9IG5ldyBNYXAoKVxuXG4gIHJldHVybiAoLi4uYXJncykgPT4ge1xuICAgIGNvbnN0IGtleSA9IEpTT04uc3RyaW5naWZ5KGFyZ3MpXG4gICAgaWYgKGNhY2hlLmhhcyhrZXkpKSB7XG4gICAgICByZXR1cm4gY2FjaGUuZ2V0KGtleSlcbiAgICB9XG5cbiAgICBjb25zdCByZXN1bHQgPSBmbiguLi5hcmdzKVxuICAgIGNhY2hlLnNldChrZXksIHJlc3VsdClcblxuICAgIHJldHVybiByZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZW1vaXplMiA9IChmbikgPT4ge1xuICBsZXQgbGFzdEFyZ3NcbiAgbGV0IGxhc3RSZXN1bHRcblxuICByZXR1cm4gKC4uLmFyZ3MpID0%2BIHtcbiAgICBpZiAoQXJyYXkuaXNBcnJheShsYXN0QXJncykgJiYgYXJncy5sZW5ndGggPT09IGxhc3RBcmdzLmxlbmd0aCkge1xuICAgICAgbGV0IGlzU2FtZSA9IHRydWVcblxuICAgICAgZm9yIChsZXQgaSA9IDA7IGkgPCBhcmdzLmxlbmd0aDsgaSsrKSB7XG4gICAgICAgIGlmIChhcmdzW2ldICE9PSBsYXN0QXJnc1tpXSkge1xuICAgICAgICAgIGlzU2FtZSA9IGZhbHNlXG4gICAgICAgIH1cbiAgICAgIH1cblxuICAgICAgaWYgKGlzU2FtZSkge1xuICAgICAgICByZXR1cm4gbGFzdFJlc3VsdFxuICAgICAgfVxuICAgIH1cblxuICAgIGxhc3RBcmdzID0gYXJnc1xuICAgIGxhc3RSZXN1bHQgPSBmbiguLi5hcmdzKVxuXG4gICAgcmV0dXJuIGxhc3RSZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZXJnZUNsYXNzZXMgPSAoYmFzZUNsYXNzZXMsIGFkZGl0aW9uYWxDbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGNvbWJpbmVkQ2xhc3NlcyA9IHsuLi5iYXNlQ2xhc3Nlc31cblxuICBmb3IgKGNvbnN0IG5hbWUgaW4gYWRkaXRpb25hbENsYXNzZXMpIHtcbiAgICBjb21iaW5lZENsYXNzZXNbbmFtZV0gPVxuICAgICAgbmFtZSBpbiBjb21iaW5lZENsYXNzZXNcbiAgICAgICAgPyBgJHtjb21iaW5lZENsYXNzZXNbbmFtZV19ICR7YWRkaXRpb25hbENsYXNzZXNbbmFtZV19YFxuICAgICAgICA6IGFkZGl0aW9uYWxDbGFzc2VzW25hbWVdXG4gIH1cblxuICByZXR1cm4gY29tYmluZWRDbGFzc2VzXG59XG5cbmNvbnN0IG0xID0gbWVtb2l6ZTEobWVyZ2VDbGFzc2VzKVxuY29uc3QgbTIgPSBtZW1vaXplMihtZXJnZUNsYXNzZXMpXG4iLCJ0ZXN0cyI6W3sibmFtZSI6Im0xIiwiY29kZSI6Im0xKFsnYSddLFsnYiBjJywgJ2QnXSlcbm0xKFsnYycsJ2cnXSxbJ2YgYycsICdkJywgJ3InXSlcbm0xKFsnYSddLFsnYiBjJywgJ2QnXSlcbm0xKFsnYycsJ2cnXSxbJ2YgYycsICdkJywgJ3InXSkiLCJydW5zIjpbNjIwMDAsODAwMDAsMTA5MDAwLDMzMDAwLDk0MDAwLDEwMDAsMTk1MDAwLDIwMDAwLDc3MDAwLDEzMDAwLDE3NDAwMCwyMDAwMCwxMDAwLDI0NzAwMCwxMDAwLDQwMDAsNDEwMDAsNTYwMDAsMTc5MDAwLDEwMDAsMjQwMDAsMTg1MDAwLDU3MDAwLDE1NjAwMCw3MjAwMCwxODMwMDAsOTEwMDAsMTAwMCwxMzcwMDAsMTcwMDAwLDE2NTAwMCwxNjkwMDAsNDAwMCwyNzAwMCw0MDAwLDIxODAwMCwzMDAwMCwxNDAwMDAsNDAwMCw0MDAwLDEzNjAwMCw5MTAwMCwxMDAwLDIwMTAwMCwxNzAwMCw5MjAwMCwyMzgwMDAsMTI1MDAwLDE4MDAwMCwxNDUwMDAsMjI5MDAwLDcwMDAsODEwMDAsMTQwMDAsMTUwMDAsNTEwMDAsMTMyMDAwLDExNDAwMCwzNzAwMCwxNjkwMDAsMTg2MDAwLDEyMDAwLDQ1MDAwLDQwMDAsODAwMCw0MDAwLDE3NzAwMCw0MDAwLDExMDAwMCwxNjcwMDAsMjQ0MDAwLDUwMDAwLDEzODAwMCwyMzIwMDAsMTA1MDAwLDIyMjAwMCw0MDAwLDU3MDAwLDE5NDAwMCwxMDkwMDAsMTAwMCwyNTgwMDAsMjk1MDAwLDEwMDAsMjM2MDAwLDAsNzAwMCwyODAwMDAsMjg4MDAwLDEwNzAwMCwxMDAwLDIwNjAwMCw0MDAwLDc4MDAwLDQwMDAsMTk1MDAwLDEyOTAwMCwxNDQwMDAsMTcwMDAwLDEwMDBdLCJvcHMiOjk4MDEwfSx7Im5hbWUiOiJtMiIsImNvZGUiOiJtMihbJ2EnXSxbJ2IgYycsICdkJ10pXG5tMihbJ2MnLCdnJ10sWydmIGMnLCAnZCcsICdyJ10pXG5tMihbJ2EnXSxbJ2IgYycsICdkJ10pXG5tMihbJ2MnLCdnJ10sWydmIGMnLCAnZCcsICdyJ10pIiwicnVucyI6WzEyMDAwLDkyMDAwLDEwMDAsNzMwMDAsMjAwMCwxMzAwMCwzMjAwMCwxMDAwMDAsMTQ1MDAwLDEwNzAwMCwxMzAwMCwxMjMwMDAsMTAwMCw5MDAwLDYwMDAsNjcwMDAsMTI1MDAwLDEyMDAwMCwxMzAwMCw1NTAwMCw3OTAwMCwzNDAwMCwxMDgwMDAsOTAwMCw2MTAwMCwyNDAwMCwxMzAwMDAsNTMwMDAsMjAwMCwxMDAwLDIxMDAwLDgyMDAwLDgxMDAwLDgwMDAwLDc1MDAwLDE0MDAwLDEwNzAwMCwyMDAwLDg0MDAwLDYxMDAwLDIwMDAsMjAwMCw0MTAwMCw0OTAwMCwxMjQwMDAsMTAwMCw3NDAwMCwyMDAwLDM1MDAwLDExMjAwMCwxMzEwMDAsODEwMDAsNDQwMDAsNTUwMDAsOTYwMDAsMTM1MDAwLDExODAwMCwxMzAwMDAsMTYwMDAsMTMwMDAsMTA4MDAwLDU4MDAwLDYwMDAwLDcxMDAwLDIwMDAsODcwMDAsNjAwMCw2OTAwMCwyMDAwLDE0MTAwMCwzOTAwMCw4MTAwMCw3MDAwLDg2MDAwLDY1MDAwLDQwMDAwLDEwMDAwMCwxNDMwMDAsMzUwMDAsMjAwMCw2NzAwMCw3MjAwMCw5NjAwMCwxMDAwLDY0MDAwLDgyMDAwLDEzMDAwLDc3MDAwLDEwMDAwMCwxMDAwLDEwMDAsODAwMDAsNTEwMDAsMTI5MDAwLDIwMDAsMjcwMDAsNzMwMDAsMTEyMDAwLDM2MDAwLDIwMDBdLCJvcHMiOjU3NjYwfV0sInVwZGF0ZWQiOiIyMDIxLTA4LTIxVDIxOjA2OjQ5LjY5NFoifQ%3D%3D)

kof:
@behnammodi thanks for the benchmark, good we have the numbers now, there are 2 important things to understand about it:

1. Your cached version is twice faster
2. The numbers are on micro benchmark level: 70000 ops/s vs 130000 ops/s
3. The benchmark essentially measures the scenario where values are cached, because you didn't randomize the class names to compare the the non-cached scenario
4. I randomized the class names for your test [here](https://perf.link/#eyJpZCI6IjNwN3oyOGR0a3cyIiwidGl0bGUiOiJKU1MiLCJiZWZvcmUiOiJjb25zdCBtZW1vaXplMSA9IChmbikgPT4ge1xuICBjb25zdCBjYWNoZSA9IG5ldyBNYXAoKVxuXG4gIHJldHVybiAoLi4uYXJncykgPT4ge1xuICAgIGNvbnN0IGtleSA9IEpTT04uc3RyaW5naWZ5KGFyZ3MpXG4gICAgaWYgKGNhY2hlLmhhcyhrZXkpKSB7XG4gICAgICByZXR1cm4gY2FjaGUuZ2V0KGtleSlcbiAgICB9XG5cbiAgICBjb25zdCByZXN1bHQgPSBmbiguLi5hcmdzKVxuICAgIGNhY2hlLnNldChrZXksIHJlc3VsdClcblxuICAgIHJldHVybiByZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZW1vaXplMiA9IChmbikgPT4ge1xuICBsZXQgbGFzdEFyZ3NcbiAgbGV0IGxhc3RSZXN1bHRcblxuICByZXR1cm4gKC4uLmFyZ3MpID0%2BIHtcbiAgICBpZiAoQXJyYXkuaXNBcnJheShsYXN0QXJncykgJiYgYXJncy5sZW5ndGggPT09IGxhc3RBcmdzLmxlbmd0aCkge1xuICAgICAgbGV0IGlzU2FtZSA9IHRydWVcblxuICAgICAgZm9yIChsZXQgaSA9IDA7IGkgPCBhcmdzLmxlbmd0aDsgaSsrKSB7XG4gICAgICAgIGlmIChhcmdzW2ldICE9PSBsYXN0QXJnc1tpXSkge1xuICAgICAgICAgIGlzU2FtZSA9IGZhbHNlXG4gICAgICAgIH1cbiAgICAgIH1cblxuICAgICAgaWYgKGlzU2FtZSkge1xuICAgICAgICByZXR1cm4gbGFzdFJlc3VsdFxuICAgICAgfVxuICAgIH1cblxuICAgIGxhc3RBcmdzID0gYXJnc1xuICAgIGxhc3RSZXN1bHQgPSBmbiguLi5hcmdzKVxuXG4gICAgcmV0dXJuIGxhc3RSZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZXJnZUNsYXNzZXMgPSAoYmFzZUNsYXNzZXMsIGFkZGl0aW9uYWxDbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGNvbWJpbmVkQ2xhc3NlcyA9IHsuLi5iYXNlQ2xhc3Nlc31cblxuICBmb3IgKGNvbnN0IG5hbWUgaW4gYWRkaXRpb25hbENsYXNzZXMpIHtcbiAgICBjb21iaW5lZENsYXNzZXNbbmFtZV0gPVxuICAgICAgbmFtZSBpbiBjb21iaW5lZENsYXNzZXNcbiAgICAgICAgPyBgJHtjb21iaW5lZENsYXNzZXNbbmFtZV19ICR7YWRkaXRpb25hbENsYXNzZXNbbmFtZV19YFxuICAgICAgICA6IGFkZGl0aW9uYWxDbGFzc2VzW25hbWVdXG4gIH1cblxuICByZXR1cm4gY29tYmluZWRDbGFzc2VzXG59XG5cbmNvbnN0IG0xID0gbWVtb2l6ZTEobWVyZ2VDbGFzc2VzKVxuY29uc3QgbTIgPSBtZW1vaXplMihtZXJnZUNsYXNzZXMpXG4iLCJ0ZXN0cyI6W3sibmFtZSI6Im0xIHJhbmRvbWl6ZWQiLCJjb2RlIjoiY29uc3QgciA9IE1hdGgucmFuZG9tKClcbm0xKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxubTEoWydjJytyLCdnJytyXSxbJ2YgYycrciwgJ2QnK3IsICdyJytyXSlcbm0xKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxubTEoWydjJytyLCdnJytyXSxbJ2YgYycrciwgJ2QnK3IsICdyJytyXSlcbiIsInJ1bnMiOlsxMDAwLDIxMDAwLDEwMDAsNDAwMCw2MDAwMCwzNjAwMCwzMDAwLDE3MDAwLDcwMDAwLDMwMDAsNDYwMDAsMzAwMCwzMDAwLDExMDAwLDEwMDAwLDEyMDAwLDE1MDAwLDMwMDAsMzAwMCwxMDAwMCwxMDAwLDEyMDAwLDEwMDAwLDEzMDAwLDMwMDAsNDkwMDAsMzAwMCw0MDAwMCwzMDAwLDIxMDAwLDQwMDAsMTkwMDAsMzAwMCw2OTAwMCw2NTAwMCw0ODAwMCwxNDAwMCw1NjAwMCw2MDAwLDEwMDAsMTUwMDAsMjAwMCw1NTAwMCw0NjAwMCw4MDAwLDExMDAwLDY1MDAwLDMwMDAwLDEwMDAwLDUwMDAwLDIxMDAwLDQ4MDAwLDQ3MDAwLDQwMDAwLDIyMDAwLDc3MDAwLDIyMDAwLDMwMDAwLDE4MDAwLDY0MDAwLDQxMDAwLDIwMDAsMzAwMDAsNjQwMDAsMTAwMCwxOTAwMCwyMDAwMCwxOTAwMCwzNDAwMCw0MzAwMCwzMDAwLDU0MDAwLDE0MDAwLDU1MDAwLDY4MDAwLDYxMDAwLDEwMDAsNDAwMDAsMTAwMDAsNDEwMDAsMzcwMDAsNTYwMDAsNDUwMDAsODAwMCwxMjAwMCwxNjAwMCwxMDAwMCwzNzAwMCwyMjAwMCw5MDAwLDEyMDAwLDQzMDAwLDI1MDAwLDE4MDAwLDE2MDAwLDIwMDAsMTgwMDAsNzAwMCwzOTAwMCwxMTAwMF0sIm9wcyI6MjUxNjB9LHsibmFtZSI6Im0yIHJhbmRvbWl6ZWQiLCJjb2RlIjoiY29uc3QgciA9IE1hdGgucmFuZG9tKClcbm0yKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxubTIoWydjJytyLCdnJytyXSxbJ2YgYycrciwgJ2QnK3IsICdyJytyXSlcbm0yKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxubTIoWydjJytyLCdnJytyXSxbJ2YgYycrciwgJ2QnK3IsICdyJytyXSlcblxuIiwicnVucyI6WzE5MDAwLDk3MDAwLDcwMDAwLDkwMDAsNTYwMDAsNjYwMDAsMjcwMDAsMTAwMDAsNzcwMDAsNjYwMDAsMjAwMCw5NjAwMCwxMTQwMDAsMjAwMCw2NDAwMCwxMDAwLDIwMDAsMTIwMDAwLDY5MDAwLDg4MDAwLDYxMDAwLDEwMDAsMTI3MDAwLDEwMDAsNTkwMDAsODYwMDAsOTgwMDAsNzQwMDAsNzYwMDAsMTI2MDAwLDEwMzAwMCwxMDAwMCwxODAwMCwxMjMwMDAsMTAwMCw2MDAwLDEyMzAwMCw0OTAwMCwxODAwMCwyNzAwMCwxMDAwLDMwMDAsMTAwMCwzNjAwMCwyMDAwLDEyMDAwMCw0OTAwMCwyMDAwLDgwMDAwLDIxMDAwLDIwMDAwLDE4MDAwLDEwNTAwMCwxMDAwMCw5NTAwMCw4ODAwMCwxNDQwMDAsMTYwMDAsMTEyMDAwLDgyMDAwLDEzNTAwMCwxMDAwLDEzNzAwMCwxMDAwMCwyMDAwLDEyMjAwMCwxMDQwMDAsNzUwMDAsODMwMDAsMjAwMCwxNDAwMCwxMjAwMCw4MDAwMCwzMjAwMCwyNDAwMCwxMDAwMCwyMDAwLDIwMDAsOTUwMDAsMTAwMCwxMTAwMDAsMTUwMDAsNDUwMDAsOTkwMDAsMTQxMDAwLDEzNTAwMCwxMTMwMDAsMjAwMCwzMzAwMCwxMjQwMDAsOTUwMDAsMTM5MDAwLDkzMDAwLDE0MTAwMCwxNDEwMDAsNDYwMDAsMTAwMCw5MDAwMCwyMDAwLDEwMDBdLCJvcHMiOjU3NTYwfV0sInVwZGF0ZWQiOiIyMDIxLTA4LTIxVDIyOjA0OjExLjUxMFoifQ%3D%3D)

Randomized shows that you made non-cached version twice slower.


To summarize: 
1. I think non-cached version is going to be a more popular scenario than a cached one
2. Even if I am wrong about the first case for some scenario, the numbers we are talking about are on micro level and they are not worth micro optimizing. We have much much bigger performance issues in other areas I am sure.


I am going to politely decline this PR. If you have time and motivation I can point you into much more important problems JSS has, in particular bugs we urgently need to fix.

behnammodi:
> @behnammodi thanks for the benchmark, good we have the numbers now, there is a number important things to understand about it:
> 
> 1. Your cached version is twice faster
> 2. The numbers are on micro benchmark level: 70000 ops/s vs 130000 ops/s on my macbook
> 3. The benchmark essentially measures the scenario where values are cached, because you didn't randomize the class names to compare the the non-cached scenario
> 4. I randomized the class names for your test [here](https://perf.link/#eyJpZCI6Ijl6amxwN2tudjh1IiwidGl0bGUiOiJKU1MiLCJiZWZvcmUiOiJjb25zdCBtZW1vaXplMSA9IChmbikgPT4ge1xuICBjb25zdCBjYWNoZSA9IG5ldyBNYXAoKVxuXG4gIHJldHVybiAoLi4uYXJncykgPT4ge1xuICAgIGNvbnN0IGtleSA9IEpTT04uc3RyaW5naWZ5KGFyZ3MpXG4gICAgaWYgKGNhY2hlLmhhcyhrZXkpKSB7XG4gICAgICByZXR1cm4gY2FjaGUuZ2V0KGtleSlcbiAgICB9XG5cbiAgICBjb25zdCByZXN1bHQgPSBmbiguLi5hcmdzKVxuICAgIGNhY2hlLnNldChrZXksIHJlc3VsdClcblxuICAgIHJldHVybiByZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZW1vaXplMiA9IChmbikgPT4ge1xuICBsZXQgbGFzdEFyZ3NcbiAgbGV0IGxhc3RSZXN1bHRcblxuICByZXR1cm4gKC4uLmFyZ3MpID0%2BIHtcbiAgICBpZiAoQXJyYXkuaXNBcnJheShsYXN0QXJncykgJiYgYXJncy5sZW5ndGggPT09IGxhc3RBcmdzLmxlbmd0aCkge1xuICAgICAgbGV0IGlzU2FtZSA9IHRydWVcblxuICAgICAgZm9yIChsZXQgaSA9IDA7IGkgPCBhcmdzLmxlbmd0aDsgaSsrKSB7XG4gICAgICAgIGlmIChhcmdzW2ldICE9PSBsYXN0QXJnc1tpXSkge1xuICAgICAgICAgIGlzU2FtZSA9IGZhbHNlXG4gICAgICAgIH1cbiAgICAgIH1cblxuICAgICAgaWYgKGlzU2FtZSkge1xuICAgICAgICByZXR1cm4gbGFzdFJlc3VsdFxuICAgICAgfVxuICAgIH1cblxuICAgIGxhc3RBcmdzID0gYXJnc1xuICAgIGxhc3RSZXN1bHQgPSBmbiguLi5hcmdzKVxuXG4gICAgcmV0dXJuIGxhc3RSZXN1bHRcbiAgfVxufVxuXG5jb25zdCBtZXJnZUNsYXNzZXMgPSAoYmFzZUNsYXNzZXMsIGFkZGl0aW9uYWxDbGFzc2VzKSA9PiB7XG4gIGNvbnN0IGNvbWJpbmVkQ2xhc3NlcyA9IHsuLi5iYXNlQ2xhc3Nlc31cblxuICBmb3IgKGNvbnN0IG5hbWUgaW4gYWRkaXRpb25hbENsYXNzZXMpIHtcbiAgICBjb21iaW5lZENsYXNzZXNbbmFtZV0gPVxuICAgICAgbmFtZSBpbiBjb21iaW5lZENsYXNzZXNcbiAgICAgICAgPyBgJHtjb21iaW5lZENsYXNzZXNbbmFtZV19ICR7YWRkaXRpb25hbENsYXNzZXNbbmFtZV19YFxuICAgICAgICA6IGFkZGl0aW9uYWxDbGFzc2VzW25hbWVdXG4gIH1cblxuICByZXR1cm4gY29tYmluZWRDbGFzc2VzXG59XG5cbmNvbnN0IG0xID0gbWVtb2l6ZTEobWVyZ2VDbGFzc2VzKVxuY29uc3QgbTIgPSBtZW1vaXplMihtZXJnZUNsYXNzZXMpXG4iLCJ0ZXN0cyI6W3sibmFtZSI6Im0xIHJhbmRvbWl6ZWQiLCJjb2RlIjoibGV0IHJcbnIgPSBNYXRoLnJhbmRvbSgpXG5tMShbJ2EnK3JdLCBbJ2IgYycgKyByLCAnZCcgKyByXSlcbnIgPSBNYXRoLnJhbmRvbSgpXG5tMShbJ2MnK3IsJ2cnK3JdLFsnZiBjJytyLCAnZCcrciwgJ3InK3JdKVxuciA9IE1hdGgucmFuZG9tKClcbm0xKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxuciA9IE1hdGgucmFuZG9tKClcbm0xKFsnYycrciwnZycrcl0sWydmIGMnK3IsICdkJytyLCAncicrcl0pXG4iLCJydW5zIjpbMjEwMDAsMzEwMDAsOTAwMCwxMDAwLDMxMDAwLDIxMDAwLDQwMDAsMjgwMDAsMjIwMDAsNTUwMDAsMjgwMDAsMjIwMDAsNTkwMDAsMjQwMDAsNDcwMDAsNDgwMDAsMjcwMDAsNjAwMCwzNjAwMCw1MjAwMCw1NTAwMCwzMzAwMCw1OTAwMCwzMTAwMCwxMDAwLDIwMDAsMzAwMCw1NDAwMCwxNzAwMCw0OTAwMCwyMDAwLDQ5MDAwLDE5MDAwLDI4MDAwLDEyMDAwLDQ4MDAwLDQxMDAwLDYxMDAwLDQ2MDAwLDEwMDAwLDMwMDAwLDQ4MDAwLDM0MDAwLDIwMDAsNTYwMDAsMTAwMCwzOTAwMCwyMjAwMCwxMDAwLDI4MDAwLDU2MDAwLDMwMDAsNDcwMDAsNTEwMDAsNTAwMDAsNDgwMDAsNTAwMCw0ODAwMCw1MzAwMCwxMDAwLDM2MDAwLDUxMDAwLDE3MDAwLDYxMDAwLDEwMDAsMzIwMDAsMzgwMDAsMzkwMDAsNTIwMDAsMjAwMCwzMDAwMCw0MjAwMCwzMDAwMCwxMjAwMCwzNDAwMCw1MzAwMCw1ODAwMCwxMDAwLDUwMDAwLDIwMDAsNDIwMDAsNTYwMDAsNTIwMDAsNDIwMDAsNDUwMDAsMzAwMDAsMzEwMDAsNDkwMDAsNjQwMDAsNTkwMDAsNDcwMDAsNTYwMDAsMjAwMCw1MzAwMCwyMzAwMCwzMDAwLDE4MDAwLDQzMDAwLDE0MDAwLDQ3MDAwXSwib3BzIjozMjMyMH0seyJuYW1lIjoibTIgcmFuZG9taXplZCIsImNvZGUiOiJsZXQgclxuciA9IE1hdGgucmFuZG9tKClcbm0yKFsnYScrcl0sIFsnYiBjJyArIHIsICdkJyArIHJdKVxuciA9IE1hdGgucmFuZG9tKClcbm0yKFsnYycrciwnZycrcl0sWydmIGMnK3IsICdkJytyLCAncicrcl0pXG5yID0gTWF0aC5yYW5kb20oKVxubTIoWydhJytyXSwgWydiIGMnICsgciwgJ2QnICsgcl0pXG5yID0gTWF0aC5yYW5kb20oKVxubTIoWydjJytyLCdnJytyXSxbJ2YgYycrciwgJ2QnK3IsICdyJytyXSlcbiIsInJ1bnMiOlszNDAwMCw3MTAwMCw0MjAwMCwzMDAwMCw3ODAwMCw0MDAwLDIwMDAwLDIwMDAsOTIwMDAsNjAwMDAsMTAwMDAsMTA0MDAwLDU0MDAwLDIzMDAwLDQ5MDAwLDEwMDAwLDYwMDAsODAwMDAsMjAwMCw2NTAwMCwxMDAwMDAsMTIwMDAwLDg5MDAwLDExMDAwMCwxNzAwMCw2OTAwMCw3ODAwMCwzNTAwMCw0NDAwMCwxMjAwMCw1NTAwMCwyNjAwMCw0MDAwMCwxMDEwMDAsMTQwMDAsMTA3MDAwLDEwNjAwMCw1NjAwMCwyMDAwLDEwMDAsMTEzMDAwLDMzMDAwLDc0MDAwLDUxMDAwLDMzMDAwLDE4MDAwLDIwMDAsMTAwMDAsMzUwMDAsMTAwMCwyMDAwLDQ4MDAwLDI5MDAwLDUzMDAwLDEwMDAwLDc5MDAwLDg5MDAwLDkwMDAsOTUwMDAsMjIwMDAsMTAwMCwxMDAwLDc2MDAwLDUyMDAwLDgwMDAsMjAwMCwyNzAwMCwxMDAwMCw3MDAwLDYwMDAsNDMwMDAsMjAwMCw4ODAwMCwxMDkwMDAsMjAwMCw0MjAwMCw3MDAwLDMwMDAwLDE3MDAwLDY1MDAwLDIxMDAwLDExMDAwLDIwMDAsMTkwMDAsMTUwMDAsMjgwMDAsMTA1MDAwLDg5MDAwLDQ4MDAwLDQwMDAwLDcwMDAsMjAwMCw0OTAwMCwyOTAwMCw4MDAwLDI5MDAwLDgzMDAwLDgwMDAsMTQwMDAsMjEwMDBdLCJvcHMiOjQwNDcwfV0sInVwZGF0ZWQiOiIyMDIxLTA4LTIxVDIyOjA5OjU0LjM5M1oifQ%3D%3D)
> 
> Randomized shows that you made non-cached version twice slower.
> 
> To summarize:
> 
> 1. I think non-cached version is going to be a more popular scenario than a cached one
> 2. Even if I am wrong about the first case for some scenario, the numbers we are talking about are on micro level and they are not worth micro optimizing. We have much much bigger performance issues in other areas I am sure.
> 
> I am going to politely decline this PR. If you have time and motivation I can point you into much more important problems JSS has, in particular bugs we urgently need to fix.

Thanks, you're right, please lead me about currently performance bottleneck

kof:
I just labeled all issues I could find with "perf"

https://github.com/cssinjs/jss/issues?q=is%3Aopen+is%3Aissue+label%3Aperf

kof:
But even more than perf, I think we need to fix bugs, in particular some that seem to be impacting everyone.

kof:
For example https://github.com/cssinjs/jss/issues/1458

behnammodi:
Thanks @kof

PR #1549, opened 2021-08-31
feat(jss): Add format option to SheetsRegistry.toString()

briganti:

This PR adds a new option `format` to `SheetsRegistry.toString` so whitespaces are removed from stringified CSS.

```es6
jss.createRule('a', {float: 'left', width: '1px'})

// Default
rule.toString()
// outputs:
// .a-id {
//   float: left;
//   width: 1px;
// }

// With format:false
rule.toString({format: false}))
// outputs:
// .a-id{float:left;width:1px;}
```

## Todo

## Expectations on Changes

Indentation and line breaks are removed when `format: false` is passed to `SheetsRegistry.toString` or `rule.toString` methods.

## Changelog

- Add a new option `format` for removing whitespaces in `SheetsRegistry.toString(options)`

### DIFF
--- a/packages/jss/tests/integration/rules.js
+++ b/packages/jss/tests/integration/rules.js
@@ -388,6 +388,11 @@ describe('Integration: rules', () => {
         `)
       })
     })
+
+    it('should uglify CSS', () => {
+      const rule = jss.createRule('a', {float: 'left', width: '1px'})
+      expect(rule.toString({uglify: true})).to.be('.a-id {float: left;width: 1px;}')
+    })
   })
 
   describe('rule.toJSON()', () => {
--- a/packages/jss/tests/integration/rules.js
+++ b/packages/jss/tests/integration/rules.js
@@ -389,9 +389,9 @@ describe('Integration: rules', () => {
       })
     })
 
-    it('should uglify CSS', () => {
+    it('should remove whitespaces', () => {
       const rule = jss.createRule('a', {float: 'left', width: '1px'})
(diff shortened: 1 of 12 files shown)


packages/jss/tests/integration/rules.js
```diff
@@ -388,6 +388,11 @@ describe('Integration: rules', () => {
         `)
       })
     })
+
+    it('should uglify CSS', () => {
+      const rule = jss.createRule('a', {float: 'left', width: '1px'})
+      expect(rule.toString({uglify: true})).to.be('.a-id {float: left;width: 1px;}')
```
kof, line 394: it could be even float:left;width:1px;

kof: and the space between selector and curly brace is not needed either

kof:
I am not sure regarding "uglify" as an option name, because it may let people think its gonna do more than removing white spaces, but we won't add anything else than whitespace removal because real modification is costly at runtime.

kof:
How about "removeWhitespace" or "compact" or "format"

kof:
If "format" it would be `true` by default.

briganti:
Thanks for the fast review @kof. I've updated the PR to take into account your feedback.

packages/jss/tests/integration/rules.js
briganti: done in 2023d19

kof:
merged, thank you

kof:
released in https://github.com/cssinjs/jss/releases/tag/v10.8.0

PR #1554, opened 2021-09-13
Refine style types

kof:
Trying to make the types more accurate and ideally improve error messages in the end.

### DIFF
--- a/packages/jss/src/index.d.ts
+++ b/packages/jss/src/index.d.ts
@@ -6,44 +6,45 @@ import {Properties as CSSProperties, PropertiesHyphen as CSSPropertiesHyphen} fr
 //
 // TODO: refactor to only include Observable types if plugin is installed.
 export interface MinimalObservable<T> {
-  subscribe(
-    nextOrObserver: ((value: T) => void) | {next: (value: T) => void}
-  ): {unsubscribe: () => void}
+  subscribe(nextOrObserver: ((value: T) => void) | {next: (value: T) => void}): {
+    unsubscribe: () => void
+  }
 }
 
-type Func<P, T, R> = T extends undefined ? ((data: P) => R) : ((data: P & {theme: T}) => R)
+type Func<Data, Theme, Style> = Theme extends undefined
+  ? (data: Data) => Style | null | undefined
+  : (data: Data & {theme: Theme}) => Style | null | undefined
 
 type NormalCssProperties = CSSProperties<string | number> & CSSPropertiesHyphen<string | number>
 type NormalCssValues<K> = K extends keyof NormalCssProperties ? NormalCssProperties[K] : JssValue
 
-export type JssStyle<Props = any, Theme = undefined> =
-  | {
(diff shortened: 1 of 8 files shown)


kof:
This is a work in progress, before we can merge it I would like to turn all tests into ts files so that we have everything covered by the existing tests.

But I would love a pre-review already at this stage, because this is going to be a big PR.

packages/react-jss/tests/types/createUseStyles.ts
```diff
@@ -172,53 +171,14 @@ const noThemeArg4ClassesPass2 = noThemeArg4({...expectedCustomProps, theme: expe
...
       color: 'red',
-      // @ts-expect-error
-      nothingAllowed: ({theme: innerTheme, ...innerProps}) => ''
+      display: () => 'block'
     })
   }
 })
 
-// Nested declarations are banned (double nest test)
-const noThemeArg6 = createUseStyles<string, MyProps, MyTheme>({
```
kof, line 185: I don't clearly understand what this is testing, because nesting is not generally disallowed, its just only working with  nested syntax with &

packages/react-jss/tests/types/createUseStyles.ts
```diff
@@ -172,53 +171,14 @@ const noThemeArg4ClassesPass2 = noThemeArg4({...expectedCustomProps, theme: expe
...
-      secondNest: ({property, theme}) => ({
-        backgroundColor: 'blue',
-        // @ts-expect-error
-        nothingAllowed: ({theme: innerTheme, ...innerProps}) => ''
-      })
-    }
-  }
-})
-
-// Nested declarations are banned (triple nest test)
```
kof, line 201: why would one need double nest and triple nest test, is there code that is explicitly working in one of them and not in another?

packages/jss/src/index.d.ts
```diff
@@ -1,49 +1,50 @@
...
+    | null
+    | undefined
+} & // nesting
+{
+  [Key: `${string}&${string}`]: JssStyle<Data, Theme> | null | undefined
+}
 export type Styles<
   Name extends string | number | symbol = string,
-  Props = unknown,
+  Data = unknown,
```
kof, line 35: This was supposed to be Data all along, the fact it can be Props is up to user, it could be anything

packages/jss/src/index.d.ts
```diff
@@ -1,49 +1,50 @@
...
+> =
+  | Record<
+      Name,
+      | JssStyle<Data, Theme>
+      | Array<JssStyle<Data, Theme>>
+      | string
+      | Func<Data, Theme, JssStyle<undefined, undefined> | string | null | undefined>
+      | MinimalObservable<JssStyle | string | null | undefined>
+    >
+  | Record<`@keyframes ${string}`, Record<'from' | 'to' | `${number}%`, JssStyle<Data, Theme>>>
```
hosseinmd, line 46: Is need to add `@global` here?

PR #1556, opened 2021-09-18
don't crash js on failed injection

firstdorsal:
currently jss is crashing in an react material-ui csp env
this fixes the crash

## Todo

### DIFF
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -234,7 +234,7 @@ const insertRule = (container, rule, index) => {
 }
 
 const getValidRuleInsertionIndex = (container, index) => {
-  const maxIndex = container.cssRules.length
+  const maxIndex = container?.cssRules?.length
   // In case previous insertion fails, passed index might be wrong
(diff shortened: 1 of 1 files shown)


kof:
Thanks, can you add a test?

firstdorsal:
hi, thank you for your quick response.
i would love to but have no idea how.
i'm just using your project as a dependency for react material-ui and dont even know what it is actually about xD
currently your script is creating a js object access error that breaks further js/react execution on trying to inject css in a site with a strict content security policy. this breaking error is fixed by adding the null propagation operator.

kof:
Can you reproduce ig on codesandbox?

firstdorsal:
nope way to complicated to do it in sandbox.

i just tried to create a docker image for you but it is still not easy to recreate because of the moving parts.

this is the exact error that pops up:
image

kof:
First step towards this getting fixed is to reproduce it in a clean setup, without business logic. Without a test we can not merge this since we can't guarantee its not going to break some other way unless we have a use case encoded into a test.

PR #1563, opened 2021-10-10
fix memory leak in dynamic nested rule

seiyab:
## Corresponding Issue(s): 
#1360 

Fix memory leak in nested rule in function rule

## Todo

## Expectations on Changes
Resolve memory leak nested rule in function rule
```js
// e.g.
{
  a: ({ color }) => {
    '& button': {
      color
    }
  }
}
```

## Changelog
Fix memory leak with dynamic nested style.

## Note
- For nested-conditional style, I have not changed.
    -  Because change might be too big, and it might hold another problem.
    - I would tackle this on another PR.
- The test for memory leak is 24871d1 . Cherry-picking, you can verify that it works.
- I added some tests that are not irrelevant to memory leak to be sure that it doesn't break existing features.

### DIFF
--- a/packages/jss-plugin-global/src/index.test.js
+++ b/packages/jss-plugin-global/src/index.test.js
@@ -6,7 +6,7 @@ import nested from 'jss-plugin-nested'
 import global from './index'
 
 const settings = {
-  createGenerateId: () => (rule) => `${rule.key}-id`
+  createGenerateId: () => rule => `${rule.key}-id`
 }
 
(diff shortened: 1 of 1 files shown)


kof:
Wow this looks great!

I am going to investigate every change to see if everything is correct. The only thing that I don't like so far is the increase in bundle size. I have to think if we can achieve the same thing without adding that much logic.

kof:
released in https://github.com/cssinjs/jss/releases/tag/v10.8.1

Issue #1565, opened 2021-10-19
Version 10.8.1 release looks incompatible with 10.8.0.


mfaheemakhtar:
__Expected behavior:__
Things should work.

__Describe the bug:__
I am using MUI v5 and using a theme that is using jss in style provider. The theme uses @mui/styles which is depreciated for mui5 but was working fine. I am also using data-grid.

The recent release of 10.8.1 caused the app to break when a screen with data-grid is viewed. I spent multiple hours to find that there are different versions of jss and plugins in the yarn.lock, and 10.8.0 and 10.8.1 does not seem to go together because of replaceUrl or something. 

Did some tweaking but then there was a type issue. I don't know where the problem is and I could not find anything so I am posting it here. The problem is probably not with the package itself but a unique combination of different packages with different versions of jss and plugins.

__Reproduction:__ 

Unfortunately, I was not able to reproduce it outside of the project. May try on weekends if got time.

__Versions (please complete the following information):__
 - jss: 10.8.1
 - Browser [e.g. chrome, safari]: Brave
 - OS [e.g. Windows, macOS]: Windows 10 Pro, Ubuntu 20.x
 - @mui/styles: 5.0.0
 - @mui/x-data-grid-pro": "^5.0.0-beta.2"
Feel free to add any additional versions which you may think are relevant to the bug.

__Managing expectations:__

Maintainers will not be fixing the problem you have unless they have it too, if you want it to get fixed:

1. Submit a PR with a failing test
2. Discuss a solution
3. Implement it

You can also do the first step only and wait for someone else to work on a fix. Anything is much better than nothing.

I am short of time at the moment and can't really provide anything right now, maybe I can try to reproduce on weekends if I get time.

The yarn resolutions fixed the issue for me:
```
"resolutions": {
    "jss": "10.8.0",
    "jss-plugin-camel-case": "10.8.0",
    "jss-plugin-default-unit": "10.8.0",
    "jss-plugin-global": "10.8.0",
    "jss-plugin-nested": "10.8.0",
    "jss-plugin-props-sort": "10.8.0",
    "jss-plugin-rule-value-function": "10.8.0",
    "jss-plugin-vendor-prefixer": "10.8.0"
  }
```
If someone else face the similar issue, try this.

seiyab:
It seems to be better to revert 10.8.1 because #1565 is more serious than #1360 .

@mfaheemakhtar Thanks a lot. It will help.

seiyab:
[...]

PR #1574, opened 2021-10-29
fix memory leak again

seiyab:
## Corresponding Issue(s): 
#1360 #1565

Memory leak in nested function rule

## Todo

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,6 +3,5 @@
   "singleQuote": true,
   "trailingComma": "none",
   "bracketSpacing": false,
-  "printWidth": 100,
-  "arrowParens": "always"
+  "printWidth": 100
 }
--- a/.prettierrc
+++ b/.prettierrc
@@ -3,5 +3,6 @@
   "singleQuote": true,
   "trailingComma": "none",
   "bracketSpacing": false,
-  "printWidth": 100
+  "printWidth": 100,
+  "arrowParens": "always"
 }
--- a/packages/jss-plugin-global/package.json
+++ b/packages/jss-plugin-global/package.json
@@ -1,7 +1,7 @@
 {
   "name": "jss-plugin-global",
   "description": "Global styles for JSS",
-  "version": "10.8.2",
+  "version": "10.8.3",
   "license": "MIT",
   "homepage": "https://cssinjs.org/jss-global",
   "main": "dist/jss-plugin-global.cjs.js",
@@ -36,11 +36,11 @@
     "check-snapshot": "node ../../scripts/match-snapshot.js"
   },
   "devDependencies": {
-    "jss-plugin-nested": "10.8.2",
+    "jss-plugin-nested": "^10.8.3",
     "jss-v10_8_0": "npm:jss@10.8.0"
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
-    "jss": "10.8.2"
+    "jss": "^10.8.3"
   }
 }
--- a/packages/jss-plugin-global/package.json
+++ b/packages/jss-plugin-global/package.json
@@ -36,8 +36,7 @@
     "check-snapshot": "node ../../scripts/match-snapshot.js"
   },
   "devDependencies": {
-    "jss-plugin-nested": "^10.8.3",
-    "jss-v10_8_0": "npm:jss@10.8.0"
+    "jss-plugin-nested": "^10.8.3"
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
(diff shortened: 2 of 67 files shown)


packages/jss/tests/functional/sheet.js
```diff
@@ -636,4 +636,278 @@ describe('Functional: sheet', () => {
...
+      const newRule = sheet.replaceRule('b', {display: 'flex'})
+      sheet.attach()
+      expect(sheet.getRule('b')).to.be(newRule)
+    })
+
+    it('should link sheet in rules options', () => {
+      expect(sheet.getRule('a').options.sheet).to.be(sheet)
+    })
+
+    it('should not duplicate cssRules when adding replace to a detached sheet with link: true', () => {
```
kof, line 682:
```suggestion
    it('should not duplicate cssRules when replacing a rule in a detached sheet with link: true', () => {
```

packages/jss/tests/functional/sheet.js
```diff
@@ -636,4 +636,278 @@ describe('Functional: sheet', () => {
...
+    it('should not warn', () => {
+      expect(spy.callCount).to.be(0)
+    })
+  })
+
+  describe('.replaceRule() with @keyframes and attached sheet', () => {
+    let style
+    let sheet
+
+    // We skip this test as keyframes are not supported by browser.
```
kof, line 760: we should drop support the support of IE9, probably even IE10 too, so this won't be needed, but also not in the same pr, just letting you know

kof:
Good work @seiyab, which change fixes the concerns we had with the previous PR? I mean besides of splicing the index

seiyab:
@kof 
Current work just addresses splicing the index.
I'll work with `replaceRule is not a function` in other commits.
And I'm going to add more tests.

seiyab:
Current progress
- fixed
  - splicing cssRules index
  - `replaceRule is not a function`
- working
  - `react-css` might still cause memory leak. investigating.

kof:
just in case - react-css isn't documented, its an experimental interface that hopefully nobody is using in production

seiyab:
Oh I mistyped. I edited my comment. `react-css` → `react-jss`

seiyab:
The memory leak I encountered was generated by nested conditional rule, that is out of scope for this fix.
This PR might be ready for review. I will check changes I wrote before pushing "Ready for review" button.

.prettierrc
```diff
@@ -3,5 +3,6 @@
   "singleQuote": true,
   "trailingComma": "none",
   "bracketSpacing": false,
-  "printWidth": 100
+  "printWidth": 100,
+  "arrowParens": "always"
```
seiyab, line 7: I saw you added parentheses, that is removed by Prettier that runs on pre-commit.
So I changed this configuration to prevent removing parentheses from Prettier.
https://github.com/cssinjs/jss/commit/c60f3955646ffaa95c1121f85ce7942b1873126d

I found lint you committed is more likely written by Prettier v2 rather than v1 installed by devDependency.
When I accidentally run prettier installed globally in my machine and the result was similar to https://github.com/cssinjs/jss/commit/c60f3955646ffaa95c1121f85ce7942b1873126d .
`arrowParens` is `"always"` by default in Prettier v2.

kof: lets upgrade prettier

seiyab: upgraded 👍

packages/jss-plugin-global/package.json
```diff
@@ -36,7 +36,8 @@
     "check-snapshot": "node ../../scripts/match-snapshot.js"
   },
   "devDependencies": {
-    "jss-plugin-nested": "10.8.2"
+    "jss-plugin-nested": "10.8.2",
+    "jss-v10_8_0": "npm:jss@10.8.0"
```
kof, line 40: this looks strange

seiyab: What does strange mean?
- Doesn't work well?
- Not good idea?

Removing tests with older jss is OK with me.

kof:
Just realized I never sent the last review, it was in the pending state

packages/jss-plugin-global/src/index.test.js
```diff
@@ -420,4 +509,33 @@ describe('jss-plugin-global', () => {
...
   })
+
+  describe('backward compatibility', () => {
+    let oldJss
+
+    beforeEach(() => { 
+      oldJss = oldCreate(settings).use(global())
+    })
+
+    it('should replace rule on replaceRule even if jss core doesn\'t implement replaceRule', () => {
```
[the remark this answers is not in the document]
seiyab: That makes sense.
I removed tests that use incompatible version.

oleg008:
Merged it

kof:
merged

kof:
released in v10.9.0

PR #1575, opened 2021-10-29
fix: add missing index property

pahan35:
I'd like to add a missing property on which my IDE complains.
image

Index property itself is used at least here https://github.com/cssinjs/jss/blob/02cb79b345570f2495ce9650641593c8cec8b47f/packages/jss/src/Jss.js#L73.

## Todo

- [x] Add test(s) that verify the modified behavior. - Not relevant.
- [x] Add documentation if it changes public API. - Not relevant.

## Expectations on Changes
No warnings from IDE when the `index` property of sheets is used.

## Changelog

Just a single extra property.


kof:
thanks

Issue #1561, opened 2021-09-30
JSS plugin nested => incorrect nesting complex selector


heyymarco:
Hi, i found a bug:
if nesting with any pseudo class like `:is()` & `:where()` that contains two/more sub selector inside the `(...)`,
the strange result occured:

```
export default {
  'btn': {
    '&:is(.hello, .world)': {
      color: 'red'
    },
    '&:where(.hello, .world)': {
      color: 'red'
    },
    '&:not(:where(.hello, .world))': {
      color: 'red'
    },
  }
}
```

the expected result:
```
.btn-0-1-114:is(.hello, .world) {
  color: red;
}
.btn-0-1-114:where(.hello, .world) {
  color: red;
}
.btn-0-1-114:not(:where(.hello, .world)) {
  color: red;
}
```

but i got:
```
.btn-0-1-114:is(.hello, .btn-0-1-114 .world) {
  color: red;
}
.btn-0-1-114:where(.hello, .btn-0-1-114 .world) {
  color: red;
}
.btn-0-1-114:not(:where(.hello, .btn-0-1-114 .world)) {
  color: red;
}
```

I think the `jss-plugin-extend` just splitting the selector list with comma and then merging each of it with parent selector.
it doesn't aware of `parentheses balancing`

kof:
The bug is somewhere along [these lines](https://github.com/cssinjs/jss/blob/master/packages/jss-plugin-nested/src/index.js#L43), feel free to send a PR with a test and/or fix

heyymarco:
yes, i meant the `jss-plugin-nested`.
the bug is in [here](https://github.com/cssinjs/jss/blob/da36a295015745fc78018c0003cb24c68f5c8db2/packages/jss-plugin-nested/src/index.js#L31). The selector cannot just simply splited by regex. Regex cannot handle nested `parentheses balancing`. You need a `selector parser`.

kof:
I am not sure why we even try instead of simply replacing the ampersand. Need to read the change history and unit tests to figure this out.

seiyab:
jss-plugin-nested seems to intend to handle [selector list](https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list).
On the other hand, JSS itself might not support selector list.
```js
sheet = jss.createStyleSheet({
  'a, b': {
    float: 'left',
    '&.c': {color: 'red'}
  }
})
```
The code above yields sheet that passes following, that doesn't seem intended.
```js
[...]

PR #1591, opened 2021-12-04
fixed bug on jss-plugin-nested

heyymarco:
According a bug on [here](https://github.com/cssinjs/jss/issues/1561)

The bug is here:
```
  function replaceParentRefs(nestedProp, parentProp) {
    const parentSelectors = parentProp.split(separatorRegExp)
    const nestedSelectors = nestedProp.split(separatorRegExp)
```
it split the string by comma, but imagine a selector like this:
`&.hello, &.world:is(.awesome, .good, .bad)`

it will splitted to:
```js
['&.hello', '&.world:is(.awesome', '.good', '.bad)']
```
it's incorrect, and should be:
```
['&.hello', '&.world:is(.awesome, .good, .bad)']
```

and the another bug is with the parent selector `&`.  
Imagine if a selector like this:
`&.hello, &.world[data-message="you & me"]`
The `&` inside `"you & me"` is not a parent selector, it's just a string expression.

To solve the 2 problem above, i wrote a **selector parser**,  
you can see (and maybe make more perfection) here: [see my repo](https://github.com/cssfn/css-selector).
The parser is smart enough to distinguish between parent selector, universal selector, combinators, string expression, and so on.
The usage is simple: `parseSelector` => `flatMapSelectors` (replace `&` with something) => `selectorsToString`.

the package published with es9 format, but you can fork & re-transpile to es5.

### DIFF
--- a/packages/jss-plugin-nested/src/index.js
+++ b/packages/jss-plugin-nested/src/index.js
@@ -1,7 +1,12 @@
 import warning from 'tiny-warning'
 
-const separatorRegExp = /\s*,\s*/g
-const parentRegExp = /&/g
+// cssfn:
+import {
+  parseSelectors,
+  flatMapSelectors,
+  selectorsToString,
+}                           from '@cssfn/css-selector'
+
 const refRegExp = /\$([\w-]+)/g
 
 /**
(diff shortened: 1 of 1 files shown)


seiyab:
Great.
I think you should write tests.

See also:
https://github.com/cssinjs/jss/pull/1568#issuecomment-949967416
https://github.com/cssinjs/jss/pull/1523#issuecomment-869145245
https://github.com/cssinjs/jss/pull/1388#issuecomment-707008924

packages/jss-plugin-nested/src/index.js
```diff
@@ -1,7 +1,12 @@
 import warning from 'tiny-warning'
 
-const separatorRegExp = /\s*,\s*/g
-const parentRegExp = /&/g
+// cssfn:
+import {
+  parseSelectors,
+  flatMapSelectors,
+  selectorsToString,
+}                           from '@cssfn/css-selector'
```
kof, line 8: I am afraid we can't do this, its a huge bundle-size increase

kof: We need a solution that is super minimalistic

kof: Wow did you write the selector parser to fix this bug? If so huge props in any case!
But jss is runtime-executed, it gets bundled with a client-side app, the only way forward is to reduce the size, even potentially remove features to make it small.

haxxxton: @heyymarco wow dude! thats some serious work!

kof:
please reopen once you have time to continue

TchernyavskyDaniil:
This bug is from version 10.0.1 (more one year!!). Shouldn't it be fixed by the current library team? Part of my team is giving up on jss, as well as some people I know because of this problem. I honestly don't quite understand the point of preparing a move to 18 React if there are such critical bugs
Sorry if I offended or don't understand your processes, I'm speaking as a regular consumer of your library

kof:
Feel free to fix a bug with a proper test and I will merge it.

Oleg

PR #1592, opened 2021-12-10
Add optional chaining checking for sheet.

lenient:
This may cause a blank page while navigating.

## Todo


kof:
Hi
1. There must be a reason why you don't have a sheet in some case, so we need to figure out the reason and fix it, what you fixed is a consequence
2.  We need a test that reproduces the problem. Without a test I won't be able to merge

lenient:
Hi @kof ,
This happens when using micro-frontend, I don't have a standalone example to reproduce it. But this happens on my local env, and `?.` is the way to fix it.

kof:
You need to remove one dependency after another until its just jss

kof:
please reopen once you have time

Issue #1597, opened 2022-02-09
react-jss Dynamic values not work with react 18 new api


cqh963852:
__Expected behavior:__

"test" word show blue color.
![image](https://user-images.githubusercontent.com/17702287/153130093-e86f724e-066e-449d-bf1b-6a9ce4a27dcd.png)

__Describe the bug:__

style with dynamic-values seem not work.

"test" word didn't show blue color under strictMode with react 18 createRoot's render.
![image](https://user-images.githubusercontent.com/17702287/153130126-1f6d268a-46d6-4da3-be14-eb73fdd636ca.png)

__Reproduction:__ 

https://codesandbox.io/s/distracted-elgamal-2cewo

__Versions (please complete the following information):__
 - react-jss: 10.9.0
 - react: 18.0.0-rc.0
 - react: 18.0.0-rc.0

cqh963852:
![image](https://user-images.githubusercontent.com/17702287/153160714-7b676625-ba81-4442-9aa9-f49da3747d13.png)


In react 17 with strictMode
![image](https://user-images.githubusercontent.com/17702287/153161203-acfef7f2-da22-4c1a-82db-982f24a1baf5.png)

cleanup function not triggered

but in react 18 new api with strictMode the clean function been triggered

![image](https://user-images.githubusercontent.com/17702287/153161329-0e39564e-3ae7-459d-8dff-8afeba48a783.png)

cqh963852:
![image](https://user-images.githubusercontent.com/17702287/153166322-64dbf0b3-ccea-43ea-b5c9-a8e7d7c7d207.png)

The DynamicRules was add in useMemo.
It seems it won't execute again.

cqh963852:
It seems a react problem.
Due to react18's new api exec the effect clean function,But react17 not.

enisdenjo:
Because of this, `react-jss` is pretty much unusable with React 18!

OlegLustenko:
Should be related https://github.com/reactwg/react-18/discussions/110

PR #1598, opened 2022-02-10
fix: #1597 fix effect cleanup

cqh963852:
## Corresponding Issue(s): 

#1597

fix wrong effect cleanup.

## Todo


cqh963852:
Test case not passed.

PR #1604, opened 2022-03-17
fix: better sheet and dynamic rules management, leverage useInsertionEffect and support React 18

enisdenjo:
## Corresponding Issue(s): 
Closes #1597, starting point was #1598

## What Would You Like to Add/Fix?

[React 18 StrictMode runs each `useEffect` twice.](https://github.com/reactwg/react-18/discussions/19) This is done for a good reason where components experience this behaviour in production environments regularly.

Without this fix, the UI will break where components rerender fast. (See examples in #1597).

This fix makes sure the dynamic styles are loaded, and kept loaded, where necessary.

## Todo

- [ ] Add test(s) that verify the modified behavior
- [ ] Add documentation if it changes public API

PR #1608, opened 2022-04-15
fix(useStyles): manage sheet on create and then add dynamic styles

enisdenjo:
Fixes failing tests in #1604 by managing the sheet immediately after create, instead of doing so in a separate effect.


kof:
thank you, merged

kof:
released 10.9.1-alpha.0

PR #1610, opened 2022-04-21
feat(JssProvider): provide custom isSSR flag

enisdenjo:
Using `renderToString` in a browser environment is completely legal; however, in such cases, the `isInBrowser` check will be truthy - causing some parts to misbehave (like choosing the wrong effect, useLayoutEffect over useEffect).

This PR allows the user to supply a custom `isSSR` prop to the `JssProvider` overriding the default `isInBrowser` check.

Also necessary for fixing failing tests in #1609.

### DIFF
--- a/packages/react-jss/.size-snapshot.json
+++ b/packages/react-jss/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "react-jss.js": {
-    "bundled": 136053,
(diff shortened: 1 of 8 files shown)


kof:
This looks great!

kof:
DM'd you on twitter

PR #1611, opened 2022-04-23
remvoe is-in-browser dep

behnammodi:
remove `is-in-browser` module

## Todo

- All tests passed
- This is not a public API changed

## Expectations on Changes

knowing the environment is so easy and we don't need a module for that

## Changelog

remove `is-in-browser` module

### DIFF
--- a/packages/jss/.size-snapshot.json
+++ b/packages/jss/.size-snapshot.json
@@ -1,30 +1,30 @@
 {
   "jss.js": {
-    "bundled": 64937,
-    "minified": 23204,
-    "gzipped": 7190
+    "bundled": 64537,
+    "minified": 22895,
+    "gzipped": 7129
   },
   "jss.min.js": {
-    "bundled": 63541,
-    "minified": 22435,
-    "gzipped": 6848
+    "bundled": 63141,
+    "minified": 22127,
+    "gzipped": 6777
   },
   "jss.cjs.js": {
-    "bundled": 60615,
-    "minified": 26232,
-    "gzipped": 7308
+    "bundled": 60552,
+    "minified": 26185,
+    "gzipped": 7305
   },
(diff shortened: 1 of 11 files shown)


packages/react-jss/src/createUseStyles.js
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import isInBrowser from 'is-in-browser'
+import {isInBrowser} from 'jss'
```
kof, line 2: tbh I don't really want this to become a public API of jss

kof: I am fine with having it in both packages twice, unless its needed in more places, then it should be a separate package

behnammodi: I agree with you I will add it in both of the packages `jss` and `react-jss`

behnammodi: If we put it into a separate package, what name should be for this package? something like `jss-utils`?

kof: if we would do that, why not just use that package that we have now?

behnammodi: Actually, I wanted to save a few bytes, but I think this hasn't more worth

behnammodi: I will close this PR, are you agree?

kof: yep

kof: if you want to work on something, let me know and we will figure out a good issue
I recently added a discord if you want to chat https://discord.gg/e79XhqmaDH

behnammodi: Great, my pleasure

PR #1612, opened 2022-04-24
1207 - update stylesheet if !important is set

behnammodi:
Update stylesheet if `!important` flag is set before and after the update

## Todo

- All tests passed
- It wasn't a public API

## Changelog

- A issue is about when you want to update a CSS rule value with a `!important` flag

### DIFF
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -41,15 +41,15 @@ const setProperty = (cssRule, prop, value) => {
       cssValue = toCssValue(value)
     }
 
-    const indexOfImportantFlag = cssValue ? cssValue.indexOf('!important') : -1
-
-    const cssValueWithoutImportantFlag =
-      indexOfImportantFlag > -1 ? cssValue.substr(0, indexOfImportantFlag - 1) : cssValue
-
     // Support CSSTOM.
     if (cssRule.attributeStyleMap) {
       cssRule.attributeStyleMap.set(prop, cssValue)
     } else {
+      const indexOfImportantFlag = cssValue ? cssValue.indexOf('!important') : -1
+
+      const cssValueWithoutImportantFlag =
+        indexOfImportantFlag > -1 ? cssValue.substr(0, indexOfImportantFlag - 1) : cssValue
+
       cssRule.style.setProperty(
         prop,
         cssValueWithoutImportantFlag,
--- a/packages/jss/src/DomRenderer.js
+++ b/packages/jss/src/DomRenderer.js
@@ -38,19 +38,23 @@ const setProperty = (cssRule, prop, value) => {
     let cssValue = value
 
     if (Array.isArray(value)) {
-      cssValue = toCssValue(value, true)
-
-      if (value[value.length - 1] === '!important') {
-        cssRule.style.setProperty(prop, cssValue, 'important')
-        return true
-      }
+      cssValue = toCssValue(value)
     }
+
+    const indexOfImportantFlag = cssValue ? cssValue.indexOf('!important') : -1
+
+    const cssValueWithoutImportantFlag =
+      indexOfImportantFlag > -1 ? cssValue.substr(0, indexOfImportantFlag - 1) : cssValue
 
     // Support CSSTOM.
     if (cssRule.attributeStyleMap) {
       cssRule.attributeStyleMap.set(prop, cssValue)
     } else {
-      cssRule.style.setProperty(prop, cssValue)
+      cssRule.style.setProperty(
+        prop,
+        cssValueWithoutImportantFlag,
+        indexOfImportantFlag > -1 ? 'important' : ''
+      )
     }
   } catch (err) {
     // IE may throw if property is unknown.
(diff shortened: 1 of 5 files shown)


packages/jss/src/DomRenderer.js
```diff
@@ -36,12 +36,13 @@ const getPropertyValue = (cssRule, prop) => {
 const setProperty = (cssRule, prop, value) => {
   try {
     let cssValue = value
+    const important = 'important'
```
kof, line 39: I would not worry about reusing that string, js VM is doing optimizations that make it not important from perf perspective

kof:
Can you add a test please?

packages/jss/src/DomRenderer.js
behnammodi: Done

behnammodi:
Sure, please tell me what kind of test do you mean?

kof:
since this bug is in dom renderer, it will need a functional test that reads the applied css from the dom, some examples https://github.com/cssinjs/jss/blob/master/packages/jss/tests/functional/sheet.js#L548

behnammodi:
> since this bug is in dom renderer, it will need a functional test that reads the applied css from the dom, some examples https://github.com/cssinjs/jss/blob/master/packages/jss/tests/functional/sheet.js#L548

Great, I'm on it

behnammodi:
@kof Done

packages/jss/tests/functional/sheet.js
```diff
@@ -575,6 +575,31 @@ describe('Functional: sheet', () => {
     })
   })
 
+  describe('!important flag', () => {
```
kof, line 578: nit:
```suggestion
  describe('sheet.update() with !important', () => {
```

packages/jss/src/DomRenderer.js
```diff
@@ -38,19 +38,23 @@ const setProperty = (cssRule, prop, value) => {
...
-        cssRule.style.setProperty(prop, cssValue, 'important')
-        return true
-      }
+      cssValue = toCssValue(value)
     }
 
+    const indexOfImportantFlag = cssValue ? cssValue.indexOf('!important') : -1
+
+    const cssValueWithoutImportantFlag =
+      indexOfImportantFlag > -1 ? cssValue.substr(0, indexOfImportantFlag - 1) : cssValue
```
kof, line 47: 2 things for potential improvement:

1. need to compare if there is a serious advantage in performance: .indexOf + .substr vs .replace, because you could be replacing instead of searching and then .substr, but I am not sure there is a perf difference, we need an isolated bench test to see if there is a big difference, I am usually using https://esbench.com/
2. In any case we only need this on line 55, inside the else, so we are not running it in modern browsers which support attributeStyleMap, so we need to move it down

kof (changes requested):
found another thing that may potentially improve performance of this quite a bit

packages/jss/src/DomRenderer.js
kof: The reason its important to move it down into else is that this is being called very frequently and if we can avoid doing that string manipulation on every value - we should

behnammodi: What about this code?

```js
let includedImportantFlag = false
const cssValueWithoutImportantFlag = cssValue
  ? cssValue.replace('!important', () => {
      includedImportantFlag = true
      return ''
    })
  : cssValue

if (cssRule.attributeStyleMap) {
  cssRule.attributeStyleMap.set(prop, cssValue)
} else {
  cssRule.style.setProperty(
    prop,
    cssValueWithoutImportantFlag,
    includedImportantFlag ? 'important' : ''
  )
}
```

behnammodi:
> found another thing that may potentially improve performance of this quite a bit

I'm e

packages/jss/src/DomRenderer.js
behnammodi: or this:

```js
let priority = ''
const cssValueWithoutImportantFlag = cssValue
  ? cssValue.replace('!important', () => {
      priority = 'important'
      return ''
    })
  : cssValue

if (cssRule.attributeStyleMap) {
  cssRule.attributeStyleMap.set(prop, cssValue)
} else {
  cssRule.style.setProperty(prop, cssValueWithoutImportantFlag, priority)
}
``

Comment (kof) [reply packages/jss/src/DomRenderer.js:47]: lets compare the perf on esbench between this and .indexOf version, we might be unaware of high difference and this is a hot path

Comment (kof) [reply packages/jss/src/DomRenderer.js:47]: btw. on esbench I would only compare the string replacement in isolation, nothing else

Comment (behnammodi) [reply packages/jss/src/DomRenderer.js:47]: I tested by this sample:

# indexOf + substr
```js
function run(value){
  const index = value.indexOf("!important")

  const result = value.substr(0, index);
}

run("background-color:red !important;");
run("color:rbg(0, 0, 0) !important;");
run("background-color:rbg(0, 0, 0);");
```

# replace
```js
function run(value){
  const result = value.replace("!important", "");
}

run("background-color:red !important;");
run("color:rbg(0, 0, 0) !important;");
run("background-color:rbg(0, 0, 0);");
```

And the result is:
Screenshot 2022-04-24 at 23 27 54

kof:
merged 🎉

kof:
released https://github.com/cssinjs/jss/releases/tag/v10.9.1-alpha.2

PR #1623, opened 2022-06-24
Update react-jss-ts.md

arpithindukuri:
- Fix typo in line 111 (Thme -> Theme)
- Added a section on how to use the custom theme type defined in global.d.ts

- Fixing a type in react-jss typescript instructions.
- Adding instructions on how to use the defined theme. This is relevant because the custom theme interface is exported by react-jss as `DefaultTheme`, but it is declared by the user as `Theme`, leading to confusion on how to use the custom interface.

## Todo

- [ ] Verify that importing `DefaultTheme` works as shown in the proposed docs. I have tested on my end, and this works well (using next.js) but someone can verify perhaps?

## Expectations on Changes

Accept the pull request? The things I changed /addedwere cause of confusion for me, so hopefully this helps someone else :)

### DIFF
--- a/docs/react-jss-ts.md
+++ b/docs/react-jss-ts.md
@@ -108,7 +108,7 @@ If `CustomTheme` is the same across your entire application, you can define a de
 
 declare global {
   namespace Jss {
-    export interface Thme {
+    export interface Theme {
       background: string
     }
   }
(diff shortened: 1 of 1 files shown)


kof:
thank you!

Issue #1618, opened 2022-05-26
export 'useInsertionEffect' (imported as 'useInsertionEffect') was not found in 'react'


pgarciaegido:
__Expected behavior:__
I should be able to use the newer versions of `react-jss` (`10.9.1-alpha.2`) both with React 17 and React 18

__Describe the bug:__
It looks like when running the newer alpha versions with React 17, we encounter the error `export 'useInsertionEffect' (imported as 'useInsertionEffect') was not found in 'react'`.

<img width="1167" alt="Screenshot 2022-05-26 at 11 14 57" src="https://user-images.githubusercontent.com/20989060/170457809-d42bf737-eb67-45c6-8e94-0ab8f1c8a6b7.png">

That makes sense, since that function does not exist in React 17.

__Reproduction:__ 

I have created the following repo: https://github.com/pgarciaegido/bug-react17-react-jss You should be able to clone it, install and see the error. It uses version (`10.9.1-alpha.2`).

__Versions (please complete the following information):__
 - jss: `react-jss@10.9.1-alpha.2`
 - Browser: Chrome
 - OS: MacOS

---

It looks like the changes included in [this commit](https://github.com/cssinjs/jss/commit/3a7ddc3cb4d8fcddd6e83f3cfb35c5d48b51bc3b) and later deployed in `v10.9.1-alpha.1` might cause this.

Do you guys expect this version to be retrocompatible? If not, what would you recommend for someone who uses `react-jss` in a package that needs to work with both React 17 and React 18?

I hope I have described correctly the bug. Please let me know otherwise and Ill try to add more information.

Many thanks!! 🙏

rlocatelli9:
I have the same problem. The react version in my project is 17.0.2

behnammodi:
Thanks for report this issue will be resolved by this [PR](https://github.com/cssinjs/jss/pull/1627), I've tested on your repo https://github.com/pgarciaegido/bug-react17-react-jss and it works by new change.

kof:
released 10.9.2

PR #1627, opened 2022-07-31
fix import issue about useInsertionEffect on react 17

behnammodi:
## Corresponding Issue(s): 
https://github.com/cssinjs/jss/issues/1618

By this PR JSS consumers can use React 17 and 16

## Changelog
Fix `useInsertionEffect` import issue on react 16 and 17


kof (approved):
seems like react has inconsistent bundling for new features if that works

kof:
released 10.9.2

PR #1633, opened 2022-10-21
Line 23 grammar mistake

bubliman:
## Old:
1. Does JSS has overhead? 
## New:
1. Does JSS have overhead? 
## Why?
When the subject of a sentence is a third person (he, she, it) we need to follow the rules of Subject-Verb Agreement. Basically add an s to the end of the verb, but when the verb "do" is used it has priority over the other verb so it always matches the subject. In this case the verb "have" has also been matched by mistake.

## Todo

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -20,7 +20,7 @@
 
    Yes.
 
-1. Does JSS has overhead?
+1. Does JSS have overhead?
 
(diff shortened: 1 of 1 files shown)


kof:
thanks

kof:
it depends, if you want to be sure people understand why the change is happening, then yes - provide an explanation. if people don't understand the fix they won't merge it or will close it potentially

Issue #1635, opened 2022-12-13
Re-rendering root node breaks dynamic styles


WalterWeidner:
__Expected behavior:__
The root component of a React that is using `react-jss` for styles should be able to re-render and not have its styles broken.

__Describe the bug:__
When a root node re-renders and a component contains dynamic styles those styles will break. From my research it looks like this is caused by the `indexOf` check in RuleList's `[remove](https://github.com/cssinjs/jss/blob/master/packages/jss/src/RuleList.js#L117)` function. The `indexOf` uses strict equality checks which fails to find a match after the re-render. This means that you are splicing with a -1 index and effectively removing the wrong items from the list.

```
['a', 'b', 'c'].splice(-1, 1)
// ['c']
```

__Reproduction:__ 
You can see an example of this failure on this [codepen](https://codesandbox.io/s/sad-wave-7zrwwn?file=/src/app.tsx). In that example one instance of the component is unmounting while another is mounting. The cleanup function for the first instances's dynamic styles runs after the second instance has created its sheet and it removes the new sheet. 

__Versions (please complete the following information):__
 - jss: 10.9.2
 - Browser [e.g. chrome, safari]: Chrome
 - OS [e.g. Windows, macOS]: macOS

fjpedrosa:
I am interested on this. Will you open a PR?

Bundas:
Hey guys, we have encountered the same problem in our project. Basically, each time the route changed, some styles didn't load properly. For example, if the table got rerendered, the last row would have broken styles. This indeed fixes the problem. I have created a PR

ernestostifano:
@Bundas @kof 

Hi again guys, I can confirm that our issue is the same. Applied the modifications of the PR into a patch and everything is now working as expected! :tada:

Bundas:
@ernestostifano I am glad it's working for you now. Unfortunately, the fix proposed by @WalterWeidner is not addressing the root cause. But it does the trick as a hotfix to get going. 

btw for anybody wondering how to apply the patch to their project - we use https://www.npmjs.com/package/patch-package

ernestostifano:
@Bundas oh, I understand. Is someone already working on it? Is it a priority for you? If needed, I could try to help. For us it turns to be an important issue (all the static styles get lost when it happens).

Regarding the patch, for people using modern versions of yarn, especially with PnP, we used [yarn patch](https://yarnpkg.com/cli/patch).

ernestostifano:
[...]

PR #1642, opened 2023-04-05
Fix for missing dynamic rules

Bundas:
https://github.com/cssinjs/jss/issues/1635
## Todo

### DIFF
--- a/packages/jss/src/RuleList.js
+++ b/packages/jss/src/RuleList.js
@@ -114,7 +114,10 @@ export default class RuleList {
   remove(rule) {
     this.unregister(rule)
     delete this.raw[rule.key]
-    this.index.splice(this.index.indexOf(rule), 1)
+    const index = this.index.findIndex(v => v.key === rule.key);
+    if (index > -1) {
+        this.index.splice(index, 1)
+    }
   }
 
   /**


packages/jss/src/RuleList.js
```diff
@@ -114,7 +114,10 @@ export default class RuleList {
   remove(rule) {
     this.unregister(rule)
     delete this.raw[rule.key]
-    this.index.splice(this.index.indexOf(rule), 1)
```
kof, line 117: why is by reference not working?

ernestostifano: In fact, it should work with `indexOf`, while only adding the `-1` check, right? If help is needed, I can try myself. What else is needed to merge the PR? Some tests?

Bundas: I lack the greater context to effectively debug the reason why is 'find by reference' not working. 

> In fact, it should work with indexOf, while only adding the -1 check, right? 

If we checked only for -1 it wouldn't be removing old styles

ernestostifano:
Hi @Bundas, just wanted to know if there are any updates about this. Can I help in some way? Cheers

packages/jss/src/RuleList.js
festimprebreza98: > why is by reference not working?

I am not quite sure, but I think it has to do with the following:

{7233EC94-9A76-4DF6-8DDE-CA49C6CD6B0B}

When some dynamic rules are updated because the data might change, those new updated rules are not stored in the [useStyles](https://github.com/cssinjs/jss/blob/master/packages/react-jss/src/createUseStyles.js) hook. So, finally when `removeDynamicRules` happens, we remove the old rule references, instead of the new ones. I think this fix handles that correctly, but some new test cases should be added. Would be very nice if a new version would be released soon, since JSS is still widely used.

